PR #2, opened 2018-08-01
Add automated EnMasse installation playbooks

aidenkeating:
This adds two ansible playbooks. One for retrieving a released
version of the EnMasse playbook from GitHub and adding a playbook
into it that uses the 'local' host instead of using the 'enmasse'
host. This is currently run as a prerequisite of the other Ansible
playbook.

The second Ansible playbook will run the EnMasse playbook from the
location it was downloaded to.

The reason for two playbooks is due to the way calling one playbook
from another works in Ansible. It uses import_playbook. This task
is static though, meaning it's pre-processed and therefore checks
whether the playbook exists before running the play.

Since we need to download the EnMasse playbook before using
import_playbook it will always fail as the pre-processing can't
find the playbook on the machine.


aidenkeating:
@pmccarthy Would you mind taking a look? This was the approach I could get working for the `download and then run another playbook` pattern. If we're happy to include an `enmasse` group in the inventory file then we can remove https://github.com/integr8ly/installation/pull/2/files#diff-f6a6d00689d5b017759d9138b555ab63R14

odra:
Do you think it makes sense to add the enmasse ansible role in the repo and declare it as a dependency in the role's meta/main.yml file?

evals/inventories/hosts
```diff
@@ -2,4 +2,12 @@
 127.0.0.1
 
 [local:vars]
-ansible_connection=local
\ No newline at end of file
+ansible_connection=local
+# These are EnMasse-specific variables. When we run the EnMasse playbook it
+# still uses our inventory file.
+namespace=enmasse
```
pmccarthy: @aidenkeating The naming convention I went with for RH-SSO vars was rhsso_*. To avoid confusion, it might be best to rename these variables to something like enmasse_namespace etc. 

We could also move these into a vars file in the role instead otherwise it could get quite complex when dealing with multiple product installs.

aidenkeating: @pmccarthy Ah yep, didn't realise we can pass variables down to an imported playbook. Will update

aidenkeating:
@odra Which EnMasse role is that? I couldn't find one in Ansible Galaxy?

I didn't want to include the playbook, roles etc (https://github.com/EnMasseProject/enmasse/tree/master/templates/install/ansible) explicitly in the repo due to having to replace them on each release of EnMasse .

odra:
I mean the release form github, unpack, add to a folder and make it part of this repo (the same way we use the "vendor" folder in golang)

aidenkeating:
@odra Ah I get you, sounds good

david-martin:
@aidenkeating This approach seems reasonable to me.
The main concern (which it doesn't look like we have here) is that we don't depend on any of the internals for the enmasse playbook

evals/inventories/hosts
aidenkeating: @pmccarthy Just following up. Looks like `import_playbook` can't be executed from inside a role, or a running play.(https://docs.ansible.com/ansible/2.5/modules/import_playbook_module.html#id4). So we can't use a normal role vars file.

So for moving the vars somewhere, I'm not overly familiar with this part of Ansible but from what I've read we can use a `group_vars` directory  instead of placing vars in the inventory file directly. This is done in https://github.com/integr8ly/installation/pull/2/commits/c89ca88d6cf97f4762d149767d28090c59df1e31

Also I've namespaced the vars to use enmasse_*.

aidenkeating:
Verified by @ maleck13

PR #3, opened 2018-08-01
ipaas (ignite) installation

odra:
## Changes

* Installs ignite in an openshift namespace by downloading and running its installation bash script
* adds user edit role in the project
* creates the project if it does not exist

Running/testing:
```sh
ansible-playbook -i roles/ipaas/tests/inventory roles/ipaas/tests/test.yml #test playbook role
ansible-playbook -i inventories/hosts roles/ipaas/tests/ipaas.yml #ipaas playbook
```

ping @maleck13 @david-martin @aidenkeating @pmccarthy


aidenkeating:
:eyes:

evals/roles/ipaas/meta/main.yml
```diff
@@ -0,0 +1,57 @@
+galaxy_info:
```
aidenkeating: I think we can remove this, I don't think we're planning on publishing to galaxy.

evals/roles/ipaas/README.md
```diff
@@ -0,0 +1,38 @@
+Role Name
```
aidenkeating: We can probably remove this file unless there's any specific info we need.

maleck13:
Looks ok to me agree with @aidenkeating comments

evals/roles/ipaas/defaults/main.yml
```diff
@@ -0,0 +1,6 @@
+---
+# defaults file for ipaas
+namespace: eval
```
aidenkeating: I think these vars should be prefixed with `ipaas_`. See RHSSO as an example https://github.com/integr8ly/installation/blob/master/evals/roles/rhsso/vars/main.yml

evals/roles/ipaas/tasks/main.yml
```diff
@@ -0,0 +1,36 @@
...
+
+-
+  name: Select openshift project
+  shell: oc project {{ ipaas_namespace }}
+  register: project_cmd
+  failed_when: false
+
+-
+  name: Create openshift project
+  shell: oc new-project {{ nipaas_amespace }}
```
aidenkeating: Typo `nipaas_amespace -> ipaas_namespace`

evals/roles/ipaas/defaults/main.yml
```diff
@@ -0,0 +1,6 @@
+---
+# defaults file for ipaas
+ipaas_namespace: eval
```
maleck13: do we want to extract this out into a more generic ```eval_namespace``` perhaps as it may be needed by other components

maleck13:
Installed fine, but I had to run it from the evals directory. Which is fine as long as it gets documented.

odra:
There is a trello card about documentation so I think we can add all the required doc info in another pr

PR #4, opened 2018-08-02
Add role for configuring EnMasse in a users namespace

aidenkeating:
* Creates the namespace if it doesn't already exist
* Gives edit permissions to a specified user
* Retrieve the svcat binary and place it in a temp directory
* Provisions EnMasse (standard) in the namespace
* Remove the svcat binary

The play should be idempotent

### DIFF
--- a/evals/roles/enmasse/defaults/main.yml
+++ b/evals/roles/enmasse/defaults/main.yml
@@ -1,0 +1,8 @@
+---
+enmasse_serviceclass_name: enmasse-standard
+enmasse_serviceinstance_name: "{{ enmasse_serviceclass_name }}-instance"
+enmasse_service_plan: unlimited-standard
+enmasse_service_namespace: "{{ eval_namespace | default('eval') }}"
+enmasse_service_name: "{{ enmasse_service_namespace }}"
+enmasse_username: evals@example.com
+enmasse_svcat_download_path: /tmp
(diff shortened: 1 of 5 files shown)


aidenkeating:
@pmccarthy @odra @david-martin @maleck13 Mind taking a look? We can bundle the svcat binary in with the role if we want.

odra:
:eye:

odra:
@aidenkeating Just ran the playbook and it worked as expected.

One question: Is there a way `when` criteria in the enmasse infrastructure playbook? Something like `when: "{{ enmasse_deploy_infra | bool | default(true) }}"` so we can deploy a mq instance using an existing enmasse infra project.

We can leave as it is if it requires more changes than just adding a bool var.

evals/roles/enmasse/defaults/main.yml
```diff
@@ -0,0 +1,8 @@
+---
+enmasse_serviceclass_name: enmasse-standard
+enmasse_serviceinstance_name: "{{ enmasse_serviceclass_name }}-instance"
+enmasse_service_plan: unlimited-standard
+enmasse_service_namespace: "{{ eval_namespace | default('eval') }}"
+enmasse_service_name: "{{ enmasse_service_namespace }}"
+enmasse_username: evals@example.com
```
odra: I think it is a good idea to use the namespace var scheme here: `"{{ eval_username| default('evals@example.com') }}"`

aidenkeating:
@odra Like this you mean? https://github.com/integr8ly/installation/pull/4/commits/2af7418cb2da51b7bedc7cdd0c4d68a770a39b4e

odra:
@aidenkeating worked as expected :+1:

PR #12, opened 2018-08-09
Provision launcher

aidenkeating:


aidenkeating:
@david-martin Would you mind taking a look? Instructions for running are included in https://github.com/integr8ly/installation/pull/12/commits/32d4f0ad32e7c85ba57f6771e2c0e5e19e4dc03a

odra:
Getting this error: 
```
TASK [launcher : Generate Launcher SSO auth token for admin user] ************************************************************************************
fatal: [127.0.0.1]: FAILED! => {"msg": "The conditional check 'launcher_sso_auth_response.status == 200' failed. The error was: error while evaluating conditional (launcher_sso_auth_response.status == 200): 'launcher_sso_auth_response' is undefined"}

``'

File: evals/roles/launcher/defaults/main.yml [hunk 1]
```diff
@@ -0,0 +1,34 @@
+---
+launcher_namespace: "{{ eval_namespace | default('eval') }}"
```
david-martin: @aidenkeating A default of `launcher` would make sense here as the developer doesn't need to see the launcher resources, just the console url

odra:
@aidenkeating @david-martin  It worked but it deployed a rhsso instance for the launcher, is there a reason for that or it be using a shared sso instance?

david-martin:
> It worked but it deployed a rhsso instance for the launcher, is there a reason for that or it be using a shared sso instance?

@odra at this time it is expected that launcher will install and use it's own sso instance.
There is a goal to use a single shared sso isntance, but this PR is OK as the first step

evals/roles/launcher/templates/client-openshift-confidential.json.j2
```diff
@@ -0,0 +1,132 @@
+{
+    "clientId": "launcher-openshift-users",
+    "name": "Launcher Cloud Service Cluster Users",
+    "rootUrl": "https://{{ launcher_sso_route }}",
+    "adminUrl": "https://{{ launcher_sso_route }}",
+    "surrogateAuthRequired": false,
+    "enabled": true,
+    "clientAuthenticatorType": "client-secret",
+    "redirectUris": [
+        "https://{{ launcher_sso_route }}/*"
```
odra: I think we need both https://{{ launcher_sso_route }}" and "https://{{ launcher_sso_route }}/*"

evals/roles/launcher/templates/client-openshift-confidential.json.j2
```diff
@@ -0,0 +1,132 @@
...
+    ],
+    "webOrigins": [
+        "https://{{ launcher_sso_route }}/*"
```
odra: Same as above

evals/roles/launcher/tasks/provision-launcher.yml
```diff
@@ -0,0 +1,245 @@
+---
+- set_fact:
+    launcher_sso_openshift_idp_client_secret: "{{ 99999 | random | to_uuid }}"
+
+- name: Get Launcher SSO secure route
```
odra: do we need this var: `launcher_openshift_sso_route`?

We only need this secure route var considering that the role is deploying an instance of rhsso in the launcher's project.

aidenkeating: The var is the route of the OpenShift cluster SSO (the SSO that the RHSSO role playbook creates). That `Get Launcher SSO secure route` task is getting the route of the SSO that Launcher creates. The role needs to do stuff in both SSO services.

README.md
```diff
@@ -88,3 +88,50 @@ in the Service Catalog. This can be provisioned into your namespace to use EnMas
...
+
+The Launcher playbook also requires information about the existing SSO that was
+provisioned previously. It needs to know the route of the SSO. This can be
+retrieved using:
+
+```shell
+oc get route secure-sso -o jsonpath='{.spec.host}' -n rhsso
+```
+
+It also needs to know the realm to interact with. By default this would be
```
david-martin: @aidenkeating The sso user/pass should be an administrator of keycloak who can create a client.
The evals user shouldn't have this kind of permission, so maybe not a good default

evals/roles/launcher/tasks/provision-launcher.yml
odra: That's what I did but it was failing because it was creating access tokens on two different rhsso instances, I had to set the `launcher_openshift_sso_route` var to the secure sso route value to make it work.

aidenkeating:
@david-martin Comments addressed

evals/roles/launcher/tasks/provision-launcher.yml
```diff
@@ -0,0 +1,245 @@
...
+- name: Create client in OpenShift SSO
+  uri:
+    url: "https://{{ launcher_openshift_sso_route }}/auth/admin/realms/{{ launcher_openshift_sso_realm }}/clients"
+    method: POST
+    body: "{{ lookup('file', '/tmp/client-openshift-confidential.json') }}"
+    validate_certs: "{{ launcher_sso_validate_certs }}"
+    body_format: json
+    headers:
+      Authorization: "Bearer {{ launcher_openshift_sso_auth_response.json.access_token }}"
+    status_code: [200, 409]
```
odra: needs to be `[201, 409]`

aidenkeating: @odra Done

david-martin:
I'm OK with this merge. Apologies for the slow response

PR #13, opened 2018-08-09
Remove old default values in Launcher role

aidenkeating:
Currently there are default values from a now deprovisioned cluster
and removed OAuth application. These were used to test the Launcher
playbook and were included accidentally, they can be removed.

### DIFF
--- a/evals/roles/launcher/defaults/main.yml
+++ b/evals/roles/launcher/defaults/main.yml
@@ -19,13 +19,13 @@ launcher_sso_image_stream: redhat-sso72-openshift
 launcher_sso_image_stream_tag: "1.1"
 launcher_sso_force_image_streams_update: true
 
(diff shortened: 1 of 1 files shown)


aidenkeating:
@maleck13 Mind taking a look?

Also @maleck13 @david-martin Is there supposed to be some URL somewhere in the users namespace pointing to Launcher (which is in a namespace not visible to the user)? At the moment there isn't.

david-martin:
> Also @maleck13 @david-martin Is there supposed to be some URL somewhere in the users namespace pointing to Launcher (which is in a namespace not visible to the user)? At the moment there isn't.

Currently, no.
The user has to just know what the url is.

This card would change that, albeit a hacky way of making the url available
https://trello.com/c/NVn2XsVc/82-create-provisioned-services-that-point-at-the-existing-che-and-launcher-via-template-broker-ie-create-just-a-config-map-and-retu

PR #14, opened 2018-08-10
Deploy all

odra:
Adds the required changes to run a single playbook to deploy all services.

Verification:

```sh
ansible-playbook -e "eval_app_host=your-cluster-host"  playbooks/all.yml
``` 

It can also be used with tags:

```
#specify which tags to run
ansible-playbook -e "eval_app_host=your-cluster-host" playbooks/all.yml --tags 'bootstrap,rhsso,che'
#skip some tags
ansible-playbook -e "eval_app_host=your-cluster-host"  playbooks/all.yml --skip-tags 'rhsso'
```

Specify the action (install or uninstall) - only supported in che's role for now:

```sh
ansible-playbook -e "eval_app_host=your-cluster-host" -e 'eval_action=uninstall' -e 'che_delete_namespace=true' playbooks/all.yml --tags 'bootstrap,che'
```

PR #17, opened 2018-08-14
Remove edit permissions in Che namespace

aidenkeating:


README.md
```diff
@@ -89,7 +89,32 @@ cd evals/
...
+Before running the playbook, create a new OAuth Application on GitHub. This can
+be done at https://github.com/settings/developers. Note the `Client ID` and
+`Client Secret` fields of the OAuth Application, these are required by the Che
+playbook.
+
+Set the following variables:
+
+* `che_github_client_id` - The `Client ID` of the created GitHub OAuth Application.
+* `che_github_client_secret` - The `Client Secret` of the created GitHub OAuth Application.
+* `che_route_suffix` - The router suffix of the OpenShift cluster.
```
pmccarthy: @aidenkeating Not directly related to this PR but if the ```che_route_suffix``` is similar to the ```evals_app_host``` var that was added as part of https://github.com/integr8ly/installation/pull/14 then we should be able to dynamically set the router suffix? See https://github.com/integr8ly/installation/blob/7f4a87e63e194aada112662d557c247350fd330e/evals/playbooks/install-all.yml#L47

The less vars that need to be set the better 😃

pmccarthy:
@aidenkeating Since we're updating the README here, would you mind adding Che to the list of products in the overview section please? https://github.com/integr8ly/installation/blob/master/README.md#overview

README.md
```diff
@@ -89,7 +89,32 @@ cd evals/
...
+* `che_keycloak_user` - Username to authenticate as, this would be the admin user by default.
+* `che_keycloak_password` - Password of the user.
+* `che_namespace` - The namesapce to provision che into.
+* `che_infra_namespace` - This can usually be the same as `che_namespace`.
+
+Run the playbook.
+
+```shell
+cd evals/
+ansible-playbook -i inventories/hosts playbooks/che.yml
```
pmccarthy: The playbook is named ```che-install.yml``` rather than ```che.yml```

README.md
aidenkeating: @pmccarthy Yep that makes sense! I'll update the defaults to use the variable if it's available. That can be used by the `install-all.yml` playbook in #14. This documentation is for running the Che playbook independently so I'll keep the documentation here for now.

pmccarthy:
@aidenkeating The removal of the user edit permissions for the che namespace and the update to Che docs made it back to master as part of https://github.com/integr8ly/installation/pull/14

I also noticed that the merged README mentioned Che and Launcher but didn't have them listed so I went ahead and issued a separate PR for that: https://github.com/integr8ly/installation/pull/20

So for this PR, the only thing needed at this point is setting the ```che_route_suffix ``` var in defaults

PR #19, opened 2018-08-15
[WIP] DO NOT MERGE - Keycloak operator 

witmicko:
Installs Keycloak operator based on its tag at the moment https://github.com/aerogear/keycloak-operator/tree/v0.0.1
Verification notes:
Follow installation notes and set up your hosts, run:
`ansible-playbook -i inventories/hosts playbooks/rhsso-operator.yml`
at the end of the play you ll see sumarry with users credentials for evaluation


evals/inventories/hosts
```diff
@@ -8,7 +8,7 @@ ansible_connection=local
 master
 
 [OSEv3:vars]
-ansible_user=<ssh-username>
+ansible_user=ec2-user
```
david-martin: @pmccarthy Is this OK to change as the default ansible user, or should we leave it as a placeholder?

evals/playbooks/rhsso-operator.yml
```diff
@@ -0,0 +1,35 @@
+---
```
david-martin: @witmicko 
Is the rhsso-operator playbook & role a copy/paste of the existing rhsso ones, but with some modifications?

Is this intended to replace the rhsso playbook & role?

evals/roles/rhsso-operator/tasks/install-operator.yml
```diff
@@ -0,0 +1,28 @@
...
+  changed_when: output.rc == 0
+
+- name: Clone Keycloak-operator
+  git:
+    repo: '{{ keycloak_operator_repo_remote }}'
+    dest: /tmp/keycloak-operator
+    version: '{{ keycloak_operator_repo_tag }}'
+
+- name: install
+  command: make install chdir=/tmp/keycloak-operator/
```
david-martin: I'd be against doing any sort of compile/make at installation time.
The install should use published resources e.g. an image in quay.io, a template in a particular branch/tag of a repo.

evals/roles/rhsso-operator/tasks/install-operator.yml
```diff
@@ -0,0 +1,28 @@
...
+
+- name: deploy operator
+  command: "kubectl create -f deploy/operator.yaml -n '{{keycloak_operator_namespace}}' chdir=/tmp/keycloak-operator/"
```
david-martin: You an reference the template from a raw github url instead of having to clone the repo

evals/inventories/hosts
pmccarthy: @david-martin `ec2-user` will work with PDS but for other clusters it may vary. For example the other Integreatly clusters require this to be set to `root`. If we feel the majority of install runs will be against PDS then it should be ok to change. Otherwise it may lead to confusion.

If we do change this we'll also need to update the README

witmicko: I will revert that to the example and update README with known usernames/cluster pairs

witmicko:
@david-martin addressed all comments, please disregard rhsso-operator role, for now, keeping it in for reference until rhsso is verified to work in same fashion

pmccarthy:
@david-martin @witmicko Is this PR still valid?

david-martin:
@pmccarthy Changes are on hold.
Likely still valid, but out of date.
Could probably close and new pr opened when needed

cc @maleck13

witmicko:
Closing in favour of https://github.com/integr8ly/installation/pull/123/

PR #21, opened 2018-08-17
Making launcher catalog configurable through template install

pmccarthy:
**Summary**
Make booster catalog URL and GIT Ref configurable by exposing them as variables in the launcher ansible role

**Validation**
* Install launcher and check that a new integreatly messaging booster is available
* Ensure this new booster provisions successfully

screen shot 2018-08-17 at 10 13 11


odra:
@pmccarthy worked as expected.

One question: it tried to pull the following image: nodejs-messaging-work-queue-frontend:latest but it failed which didn't stop the deployment itself - is this expected?

pmccarthy:
@odra Hmm ok, it should just work. Is it possible for me to get access to the cluster you tested against? So that I can check configs?

odra:
@pmccarthy looks to be the default launcher behavior (it starts a second deployment with the correct image)  so it should be ok.

PR #24, opened 2018-08-20
Che launcher sso

odra:
## Changes

* Using launcher's sso info for che
* Uses multi user vars when deploying che
* Fixes launcher deployment when sso client id was already created

### DIFF
--- a/evals/roles/launcher/tasks/provision-launcher.yml
+++ b/evals/roles/launcher/tasks/provision-launcher.yml
@@ -70,7 +70,7 @@
   register: openshift_client_list_response
 
 -
-  name: Get client
+  name: Get existent client
   uri:
     url: "https://{{ launcher_openshift_sso_route }}/auth/admin/realms/{{ launcher_openshift_sso_realm }}/clients/{{ openshift_client_list_response.json[0].id }}"
     method: GET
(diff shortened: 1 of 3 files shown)


evals/roles/launcher/tasks/provision-launcher.yml
```diff
@@ -55,6 +55,44 @@
...
+    validate_certs: "{{ launcher_sso_validate_certs }}"
+    body_format: json
+    headers:
+      Authorization: "Bearer {{ launcher_openshift_sso_auth_response.json.access_token }}"
+    status_code: [200]
+  when: openshift_client_list_response is not skipped
+  register: openshift_client_read_response
+
+-
+  debug: var=openshift_client_read_response
```
pmccarthy: Do we need this debug line?

pmccarthy:
@odra Does this mean the Github oauth application should be the same for both?

odra:
No, still one github auth for each role

pmccarthy:
@odra Hmm ok. I thought the point of having Che use the Launcher SSO was so that we would only need a single oauth app? The Launcher keycloak has a github identity provider configured meaning both Launcher and Che could authenticate that provider?

odra:
Ah, you right since it uses the sso url as the callback url so we can use the same auth app for both roles now

pmccarthy:
Validated using the below command:
```
```
Was able to authorize Github account from launcher initially and then go to the Che dashboard and link to Github seemlessly.

PR #25, opened 2018-08-22
Set sso redirect uri

pmccarthy:
**Summary**
Changing logic for retrieving the openshift master url by reading in the masterPublicURL value from the master config file

**Validation**
Install rhsso independently:
```
ansible-playbook -i inventories/hosts playbooks/rhsso.yml
```
Install rhsso using install-all.yml playbook
```
```
In both cases, ensure that you can login to the cluster as the ```evals@example.com``` user

### DIFF
--- a/evals/playbooks/rhsso.yml
+++ b/evals/playbooks/rhsso.yml
@@ -1,8 +1,23 @@
 ---
+- hosts: master
+  gather_facts: no
+  tasks:
+    - name: Retrieve master public URL
+      slurp:
+        src: "{{ rhsso_openshift_master_config_path }}"
+      register: rhsso_master_config
+      become: yes
+
+    - add_host:
+        name: RHSSO_VARS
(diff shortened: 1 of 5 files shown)


evals/playbooks/rhsso.yml
```diff
@@ -1,8 +1,23 @@
...
+  tasks:
+    - name: Retrieve master public URL
+      slurp:
+        src: "{{ rhsso_openshift_master_config_path }}"
+      register: rhsso_master_config
+      become: yes
+
+    - add_host:
+        name: RHSSO_VARS
+        openshift_master_url: "{{ (rhsso_master_config['content'] | b64decode | from_yaml)['masterPublicURL'] }}"
```
odra: Can you add a "remote" tag for those two tasks so it stays the same as the one we use in the "install-all" playbook?

odra:
@pmccarthy just one ow two comments (there is nothing blocking the pr) - I gave it a try and it worked as expected.

PR #26, opened 2018-08-23
replace ignore_errors with failed_when

odra:
replacing `ignore_errors` with `failed_when` checks.

### DIFF
--- a/evals/roles/che/tasks/uninstall.yml
+++ b/evals/roles/che/tasks/uninstall.yml
@@ -19,13 +19,13 @@
   name: Retrieve service account secrets
   shell: "oc get sa -o jsonpath='{.items[*].secrets[*].name}' -l app={{ che_app_label }} -n {{ che_namespace }}"
   register: che_sa_cmd
-  ignore_errors: true
+  failed_when: false
 
 -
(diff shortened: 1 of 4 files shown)


evals/roles/che/tasks/uninstall.yml
```diff
@@ -19,13 +19,13 @@
   name: Retrieve service account secrets
   shell: "oc get sa -o jsonpath='{.items[*].secrets[*].name}' -l app={{ che_app_label }} -n {{ che_namespace }}"
   register: che_sa_cmd
-  ignore_errors: true
+  failed_when: false
```
david-martin: If the serviceaccount doesn't exist, I think its OK to continue (bad install or user manually removed the sa) as the delete step only runs if this command exits 0.

david-martin:
@odra what is the difference with this change?
Are there different error behaviours that do cause the playbook to fail?

odra:
It doesn't show as much red text in the screen as it was being displayed, since we didn't want to give the impression that the playbook failed for some reason.

david-martin:
@aidenkeating i'm good with these changes if you are

PR #27, opened 2018-08-23
Sso cluster admin user

pmccarthy:
**Summary**
Adding cluster admin user to keycloak
Adding support for removing htpasswd identity provider from master config

**Validation**
Install rhsso either through the individual playbook:
```
ansible-playbook -i inventories/hosts playbooks/rhsso.yml
```
Or through the install-all.yaml playbook:
```
```
Once completed, you should be able to login to openshift as cluster-admin user:
```
admin@example.com / Password1
```
The htpasswd identity provider should not be given as an option when logging into the console. SSO should be the sole identity provider.


odra:
@pmccarthy I was able to login as the admin user but it is now failing to login as the "evals@example.com" user with the following error: Could not create user.

pmccarthy:
@odra By any chance did you already have an evals@example.com user on your cluster before running the install? I've hit this issue previously. Need to remove all users from both keycloak and openshift and run the install again. It shouldn't happen on a vanilla install

odra:
I already had a running sso instance but I deleted it before running the installer

pmccarthy:
@odra After removing the SSO namespace, did you also run?:
```
oc delete user evals@example.com admin@example.com
```

odra:
no, let me give it a try :)

odra:
@pmccarthy that did the trick :+1:  feel free to merge the pr

pmccarthy:
@odra Great, thanks. Unfortunately the users need to be cleaned up if a previous SSO had been configured. Hopefully not something users will encounter as they shouldnt need to re-deploy SSO for demo purposes

PR #29, opened 2018-08-27
Add prerequisite checks for GitHub OAuth application credentials

aidenkeating:
* Change `launcher_github_client_*` to `eval_github_client_*`.
* Ensure the two values are set before starting the installation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,14 +65,14 @@ There are currently two options for installing:
 
 All products can be installed using the ```install-all.yml``` playbook located in the ```evals/playbooks/``` directory.
 
-Before running the playbook, create new OAuth Applications on GitHub for both Che and Launcher. This can
+Before running the playbook, create a new OAuth Applications on GitHub. This can
 be done at https://github.com/settings/developers. Please note the `Client ID` and
-`Client Secret` fields of the OAuth Applications and pass them into the install command as follows:
+`Client Secret` fields of the OAuth Application and pass them into the install command as follows:
 
 ```shell
 oc login https://<openshift-master-url>
 cd evals/
-ansible-playbook -i inventories/hosts playbooks/install-all.yml -e launcher_github_client_id=<launcher-client-id> -e launcher_github_client_secret=<launcher-client-secret> -e che_github_client_id=<che-client-id> -e che_github_client_secret=<che-client-secret>
+ansible-playbook -i inventories/hosts playbooks/install-all.yml -e eval_github_client_id=<github-client-id> -e eval_github_client_secret=<github-client-secret>
 ```
 
 #### Install each product individually
--- a/README.md
+++ b/README.md
@@ -65,14 +65,14 @@ There are currently two options for installing:
 
 All products can be installed using the ```install-all.yml``` playbook located in the ```evals/playbooks/``` directory.
 
-Before running the playbook, create a new OAuth Applications on GitHub. This can
+Before running the playbook, create a new OAuth Application on GitHub. This can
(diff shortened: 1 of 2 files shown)


evals/playbooks/install-all.yml
```diff
@@ -65,6 +77,8 @@
         launcher_openshift_sso_username: "{{ eval_rhsso_admin_username }}"
         launcher_openshift_sso_password: "{{ eval_rhsso_admin_password }}"
         launcher_openshift_sso_route: "{{ eval_rhsso_host }}"
+        launcher_github_client_id: "{{ eval_github_client_id }}"
```
pmccarthy: Does this mean that when running the install-all playbook we should specify eval_github* rather than launcher_github* ?

If so, we should update the README to reflect this change. For example: https://github.com/integr8ly/installation#install-all-products-from-a-single-playbook

aidenkeating: Exactly, I think it looks nicer than having the launcher prefix. Because it's used by both Launcher and Che, and the end-user doesn't need to know that.

Good point, will update

README.md
```diff
@@ -72,7 +72,7 @@ be done at https://github.com/settings/developers. Please note the `Client ID` a
 ```shell
 oc login https://<openshift-master-url>
 cd evals/
-ansible-playbook -i inventories/hosts playbooks/install-all.yml -e launcher_github_client_id=<launcher-client-id> -e launcher_github_client_secret=<launcher-client-secret> -e che_github_client_id=<che-client-id> -e che_github_client_secret=<che-client-secret>
+ansible-playbook -i inventories/hosts playbooks/install-all.yml -e eval_github_client_id=<launcher-client-id> -e eval_github_client_secret=<launcher-client-secret>
```
pmccarthy: Should probably change the example values to <client-id> and <client-secret>

It would also be good if we could update the paragraph [1] just above this example to state that we only need to create a single OAuth application, rather than two separate ones for Che and Launcher if that makes sense?

[1] https://github.com/integr8ly/installation/blob/81ca5600c2e7ef879386edf33ab9c61d793d8428/README.md#install-all-products-from-a-single-playbook

pmccarthy:
Ran install-all playbook without any params which failed the pre-req check as expected:
```
TASK [Prerequisite check] ************************************************************************************************************************************
failed: [127.0.0.1] (item={u'name': u'eval_github_client_id', u'value': u''}) => {"changed": false, "failed": true, "item": {"name": "eval_github_client_id", "value": ""}, "msg": "Variable eval_github_client_id is undefined or an empty string"}
failed: [127.0.0.1] (item={u'name': u'eval_github_client_secret', u'value': u''}) => {"changed": false, "failed": true, "item": {"name": "eval_github_client_secret", "value": ""}, "msg": "Variable eval_github_client_secret is undefined or an empty string"}
```

Re-ran the install specifying the new `eval_github` params and all worked as expected:
```
ansible-playbook -i inventories/hosts playbooks/install-all.yml -e eval_github_client_id=XXXXXX -e eval_github_client_secret=XXXXXX
```

README.md
```diff
@@ -65,14 +65,14 @@ There are currently two options for installing:
 
 All products can be installed using the ```install-all.yml``` playbook located in the ```evals/playbooks/``` directory.
 
-Before running the playbook, create new OAuth Applications on GitHub for both Che and Launcher. This can
+Before running the playbook, create a new OAuth Applications on GitHub. This can
```
pmccarthy: Slight typo, applications instead of application. Just to avoid confusion

PR #30, opened 2018-08-28
Make work on OpenShift 3.10

aidenkeating:
The installer currently does not work with OpenShift 3.10.

These changes make the installer work on 3.10, but it'll no longer work with 3.9.

* Add a `ca` value to the OpenID Identity Provider in the OpenShift master config. This does not seem to be needed when the cluster uses valid certs. So there has been a flag added which defaults to using invalid certs.
* Use the new approach of restarting the API and Controllers service.

### DIFF
--- a/evals/roles/rhsso/tasks/identityprovider.yml
+++ b/evals/roles/rhsso/tasks/identityprovider.yml
@@ -18,6 +18,10 @@
   when: rhsso_disable_htpasswd_identity_provider
   become: yes
 
+- set_fact:
+    rhsso_identity_provider_ca_cert_path: ""
+  when: not (eval_self_signed_certs | bool)
+
 - name: Add RH-SSO identity provider to master config
   blockinfile:
     path: "{{ rhsso_openshift_master_config_path }}"
@@ -33,6 +37,7 @@
             kind: OpenIDIdentityProvider
             clientID: {{ rhsso_client_id }}
             clientSecret: {{ client_config.json.value }}
+            ca: {{ rhsso_identity_provider_ca_cert_path }}
             urls:
               authorize: https://{{ rhsso_route }}/auth/realms/{{ rhsso_realm }}/protocol/openid-connect/auth
               token: https://{{ rhsso_route }}/auth/realms/{{ rhsso_realm }}/protocol/openid-connect/token
@@ -49,19 +54,13 @@
   register: master_config_update
   become: yes
 
-- name: restart openshift master api service
-  service:
-    name: atomic-openshift-master-api
-    state: restarted
+- name: restart openshift master services
+  shell: "{{ rhsso_master_restart_shim_path }} {{ item }}"
   become: yes
   when: master_config_update.changed
-
-- name: restart openshift master controller service
-  service:
-    name: atomic-openshift-master-controllers
-    state: restarted
-  become: yes
-  when: master_config_update.changed
+  with_items:
+  - api
+  - controllers
 
 - name: Delete local client config file
   file: path=/tmp/client-config.json state=absent
--- a/README.md
+++ b/README.md
@@ -66,13 +66,21 @@ There are currently two options for installing:
 All products can be installed using the ```install-all.yml``` playbook located in the ```evals/playbooks/``` directory.
 
 Before running the playbook, create a new OAuth Application on GitHub. This can
-be done at https://github.com/settings/developers. Please note the `Client ID` and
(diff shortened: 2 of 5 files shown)


aidenkeating:
@david-martin @pmccarthy Would you mind taking a look?

evals/roles/rhsso/tasks/identityprovider.yml
```diff
@@ -49,19 +54,13 @@
...
-    name: atomic-openshift-master-api
-    state: restarted
-  become: yes
-  when: master_config_update.changed
-
-- name: restart openshift master controller service
-  service:
-    name: atomic-openshift-master-controllers
-    state: restarted
+- name: restart openshift master services
```
pmccarthy: Does this change mean we are removing support for Openshift v3.9? I wonder if theres a way to support both versions

aidenkeating: Yep it does, from what I understand until there's a proper release of the installer we don't need to worry about supporting any versions

aidenkeating:
@wei-lee @jasonmadigan This is the PR that makes 3.10 updates, if you're trying stuff out

pmccarthy:
Seems to be failing for me on:
```
TASK [launcher : Fail if variable {{ item }} is undefined] ***************************************************************************************************
failed: [127.0.0.1] (item=) => {"changed": false, "item": "", "msg": "Variable  is undefined"}
failed: [127.0.0.1] (item=) => {"changed": false, "item": "", "msg": "Variable  is undefined"}
```

jasonmadigan:
thanks @aidenkeating - I'll give it a spin on my cluster

aidenkeating:
@pmccarthy That's probably because I haven't rebased yet, `launcher_github_client_*` still needs to be specified instead of `eval_github_client_*`. I'll rebase (and fix that templating in the task name) now

aidenkeating:
@pmccarthy Updated now

pmccarthy:
@aidenkeating Looks like that's fixed the initial problem but I'm hitting another issue now:
```
TASK [rhsso : set_fact] **************************************************************************************************************************************
fatal: [ip-10-0-19-108.us-west-2.compute.internal]: FAILED! => {"msg": "The conditional check 'not (eval_self_signed_certs | bool)' failed. The error was: error while evaluating conditional (not (eval_self_signed_certs | bool)): 'eval_self_signed_certs' is undefined\n\nThe error appears to have been in '/tmp/installation/evals/roles/rhsso/tasks/identityprovider.yml': line 21, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- set_fact:\n  ^ here\n"}
fatal: [ip-10-0-40-143.us-west-2.compute.internal]: FAILED! => {"msg": "The conditional check 'not (eval_self_signed_certs | bool)' failed. The error was: error while evaluating conditional (not (eval_self_signed_certs | bool)): 'eval_self_signed_certs' is undefined\n\nThe error appears to have been in '/tmp/installation/evals/roles/rhsso/tasks/identityprovider.yml': line 21, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- set_fact:\n  ^ here\n"}
fatal: [ip-10-0-95-39.us-west-2.compute.internal]: FAILED! => {"msg": "The conditional check 'not (eval_self_signed_certs | bool)' failed. The error was: error while evaluating conditional (not (eval_self_signed_certs | bool)): 'eval_self_signed_certs' is undefined\n\nThe error appears to have been in '/tmp/installation/evals/roles/rhsso/tasks/identityprovider.yml': line 21, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- set_fact:\n  ^ here\n"}
```

pmccarthy:
@aidenkeating Seems to look for a route in the Che namespace but there are no routes?

```
TASK [che : Get che route] ***********************************************************************************************************************************
fatal: [127.0.0.1]: FAILED! => {"changed": true, "cmd": "oc get route/che -o jsonpath='{.spec.host}' -n che", "delta": "0:00:00.147259", "end": "2018-08-29 14:36:57.146217", "msg": "non-zero return code", "rc": 1, "start": "2018-08-29 14:36:56.998958", "stderr": "Error from server (NotFound): routes.route.openshift.io \"che\" not found", "stderr_lines": ["Error from server (NotFound): routes.route.openshift.io \"che\" not found"], "stdout": "", "stdout_lines": []}
```

PR #31, opened 2018-08-30
Adding logout support for SSO

pmccarthy:
**Summary**
Fixing logout issues when SSO set as sole identity provider in an Openshift cluster

**Validation**

- Install RHSSO using ansible playbooks
- Login as admin and/or evals user
- Verify that you can logout successfully


aidenkeating:
:eyes:

PR #32, opened 2018-08-30
add playbook and role for managed service broker installation

JameelB:
## Description
Add playbook and role for managed service broker installation and also include this in `install-all.yml`.

## Progress
- [x] Create msbroker role
- [x] Create msbroker playbook
- [x] Add msbroker installation in `install-all.yml`

## Verification Steps
Test both playbooks:
- Use the `install-all.yml` playbook to install all products
- Run `ansible-playbook -i inventories/hosts playbooks/msbroker.yml` to install the managed service broker by itself.

**Expected Results:**
- `managed-services-broker` namespace created
- `managed-services` role and `default-account-managed-services` rolebinding created in this namespace
- `managed-services` clusterrole and `default-cluster-account-managed-services` clusterrolebinding created.
- `msb` deployment config created in the `managed-services-broker` namespace.
- `msb` should have an environment variable for `ROUTE_SUFFIX` with the correct value.
- `managed-services-broker` clusterservicebroker should be created.

### DIFF
--- a/evals/roles/msbroker/tasks/main.yml
+++ b/evals/roles/msbroker/tasks/main.yml
@@ -10,12 +10,26 @@
   when: msbroker_check_namespace_exists.rc != 0
 
 # Create RBAC
+- name: Check if rbac roles exists
+  shell: oc get -f {{ msbroker_rbac_url }} -n {{ msbroker_namespace }}
+  register: check_rbac_exist
+  failed_when: false
+
+- set_fact:
+    total_roles: 4 # Total number of roles that is in rbac.yml
+    total_roles_found: "{{ (check_rbac_exist.stdout_lines | length) / 2 }}" # Total number of roles that already exists
+
+- set_fact:
+    num_roles_to_create: "{{ total_roles - (total_roles_found | int) }}"
+
 - name: Create RBAC for the managed services broker
   shell: oc create -f {{ msbroker_rbac_url }} -n {{ msbroker_namespace }}
+  register: create_rbac_result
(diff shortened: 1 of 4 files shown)


JameelB:
ping @sedroche

evals/roles/msbroker/tasks/main.yml
```diff
@@ -0,0 +1,24 @@
...
+  shell: oc get namespace {{ msbroker_namespace }}
+  failed_when: false
+  register: msbroker_check_namespace_exists
+
+- name: Create managed services broker namespace
+  shell: oc new-project {{ msbroker_namespace }}
+  when: msbroker_check_namespace_exists.rc != 0
+
+# Create RBAC
+- name: Create RBAC for the managed services broker
```
sedroche: Does this fail if the role already exists? Can we add a check to see if this exists already?

sedroche:
Do we need to add these steps for TLS? https://github.com/aerogear/managed-services-broker#how-the-broker-uses-tlsssl

JameelB:
@sedroche The annotation and CA bundle described there are already in the broker template. https://github.com/aerogear/managed-services-broker/blob/master/templates/broker.template.yaml#L14, https://github.com/aerogear/managed-services-broker/blob/master/templates/broker.template.yaml#L120

sedroche:
@JameelB :+1:

evals/roles/msbroker/tasks/main.yml
JameelB: Yes, the playbook fails if the role already exists. Problem here is that both the role and clusterrole are in the same file which makes it hard to do a check for each. We can probably divide this rbac file in the broker into two which will help us add a check here. wdyt @david-martin @maleck13?

sedroche:
Verified. Installs broker and is repeatable

PR #34, opened 2018-09-04
add launcher and che dashboard url to the broker template as params

JameelB:
## Description
Get the Launcher and Che dashboard URL and pass them to the broker template as parameters so that the broker can pass them back when a user provisions an instance of Launcher and Che.

## Progress
- [x] Retrieve Launcher/Che dashboard URL
- [x] Pass the retrieved values as parameters to the broker template

## Verification Steps

- Change the targeted [image org](https://github.com/integr8ly/installation/blob/master/evals/roles/msbroker/defaults/main.yml#L3) to the correct docker org with the [managed-services-broker changes](https://github.com/aerogear/managed-services-broker/pull/12)
- Change the targeted broker template to this [link](https://raw.githubusercontent.com/JameelB/managed-services-broker/add-launcher-che-deployer/templates/broker.template.yaml) with the latest managed-services-broker changes.
- Ensure that installation is successful using either the `install-all` playbook or the `msbroker` playbook.
- Ensure that the Launcher/Che dashboard url are stored as environment variable on the managed services broker (`msb`) deployment.

### DIFF
--- a/evals/playbooks/install-all.yml
+++ b/evals/playbooks/install-all.yml
@@ -70,13 +70,6 @@
         eval_app_host: "{{ hostvars['EVAL_VARS']['eval_app_host'] }}"
       when: eval_app_host == ''
     - 
-      name: Install managed services broker
-      include_role:
-        name: msbroker
-      vars:
-        route_suffix: "{{ eval_app_host }}"
-      tags: ['msbroker']
-    - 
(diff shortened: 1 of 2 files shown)


JameelB:
ping @sedroche @maleck13 @david-martin @philbrookes

PR #38, opened 2018-09-07
Add the walkthrough webapp install role

david-martin:
To test this with a local `oc cluster up`,

```
ansible-playbook -K -i roles/webapp/tests/inventory roles/webapp/tests/test-install.yml
```

If testing in a different cluster, you will need to set the `openshift_master_config_path`
var back to `/etc/origin/master/master-config.yaml` or use a different inventory file.

After install, there should be a new namespace `webapp` with the tutorial-web-app deployed
and accessible at a route.

At this time the version of the webapp deployed will be using a mock list of services with incorrect urls.
This is expected and will change at a later point.


aidenkeating:
👀

evals/roles/webapp/defaults/main.yml
```diff
@@ -0,0 +1,6 @@
+webapp_template: https://raw.githubusercontent.com/integr8ly/tutorial-web-app/development/deployment/openshift-template.yml
+webapp_namespace: webapp
+webapp_client_id: tutorial-web-app
+webapp_route: tutorial-web-app
+
+openshift_master_config_path: /etc/origin/master/master-config.yaml
```
aidenkeating: We have a group variable to define this same value, called `eval_openshift_master_config_path`, that'd be used as an override in the install-all playbook. We can probably use that value if it's available.

So we could do something like:

```
webapp_openshift_master_config_path: {{ eval_openshift_master_config_path | default('/etc/origin/master/master-config.yaml') 
```

evals/roles/webapp/tasks/provision-webapp.yml
```diff
@@ -0,0 +1,54 @@
...
+  shell: oc get route/{{ webapp_route }} -o template --template \{\{.spec.host\}\} -n {{ webapp_namespace }}
+  register: webapp_secure_route
+  retries: 60
+  delay: 5
+  until: webapp_secure_route.rc == 0
+
+- set_fact:
+    webapp_secure_route: "{{ webapp_secure_route.stdout }}"
+
+- name: Add webapp route to corsAllowedOrigins in master config
```
aidenkeating: This command will need to be run on the master host group, if we're running the playbook on a machine that isn't a master the file won't be found.

We have a similar thing in https://github.com/integr8ly/installation/blob/1f55c13e953b95c590aa2080f0c790e64b9f9ac9/evals/playbooks/install-all.yml#L192

david-martin: :+1: Pushed and update https://github.com/integr8ly/installation/pull/38/commits/688f0146fc813797c23e36d7a7fe8ca86e90f509

evals/roles/webapp/tasks/cors.yml
```diff
@@ -0,0 +1,10 @@
+---
+- name: Add webapp route to corsAllowedOrigins in master config
+  blockinfile:
+    path: "{{ webapp_openshift_master_config_path }}"
+    insertafter: "corsAllowedOrigins"
+    backup: yes
+    block: |2
+      - https://{{ webapp_secure_route }}
```
aidenkeating: This appears to be undefined when running the install-all playbook. We may need to set `webapp_secure_route` as a hostvar? Something like https://github.com/integr8ly/installation/blob/6cd746f89b3fddc7da5c35785954d119b0d23cd2/evals/playbooks/rhsso.yml#L11

david-martin:
@aidenkeating I gave this 1 more spin locally and it looks good.

PR #47, opened 2018-09-14
Support 3Scale SSO integration

pmccarthy:
**Summary**
Add 3Scale/SSO integration support

**Validation**
Run installer:
```
ansible-playbook -i inventories/hosts playbooks/install-all.yml -e eval_github_client_id=XXXXXX -e eval_github_client_secret=XXXXXX
```
* Go to the 3scale admin console: `https://3scale-admin.<cluster-route-suffix>`
* Select SSO as login option and verify you can login as `admin@example.com` with standard SSO credentials
* Once logged in, verify admin access by clicking the Cog in the top right hand corner of the console -> Account -> Users and checking that the permissions group for the admin user is set to `Unlimited Access`

NOTE: After speaking with the 3Scale guys, it is not currently possible to configure the "Enforce SSO for all users" option via the API. This means that we will be presented with two login options on the admin console. However, I believe that 3Scale may be looking at the longer term solution of bypassing the login screen completely when backed with SSO

### DIFF
--- a/evals/inventories/group_vars/all/common.yml
+++ b/evals/inventories/group_vars/all/common.yml
@@ -5,8 +5,8 @@ eval_username: evals@example.com
 eval_app_host: ''
 eval_openshift_master_config_path: /etc/origin/master/master-config.yaml
 eval_self_signed_certs: true
+eval_sso_validate_certs: false
 eval_che_namespace: che
 eval_rhsso_namespace: sso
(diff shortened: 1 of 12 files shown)


david-martin:
:eyes:

david-martin:
Upstream issue for bypassing the auth provider choice screen https://issues.jboss.org/browse/THREESCALE-1292

PR #53, opened 2018-09-20
install walkthrough 1 templates

witmicko:
add following app's templates to the catalog:
- https://github.com/integr8ly/nodejs-messaging-work-queue/tree/deploy-frontend-without-stock
- https://github.com/benoitf/spboot-example

## Verification:
- be on pds cluster
- `git clone https://github.com/witmicko/installation.git`
- `cd installation/evals`
- `git checkout walkthrough-1`
- `ansible-playbook -i inventories/hosts playbooks/walkthrough_1.yml`
- `oc get clusterservicebroker template-service-broker -o yaml > broker.yml`
- `oc delete clusterservicebroker template-service-broker`
- wait for it to be deleted, check with `get`
- oc create -f broker.yml
- create new project, wait for catalog to load up
- provision "nodejs-messaging-work-queue-frontend" using default parameters
- create new project
- provision "Spring Boot - REST HTTP and CRUD" using default parameters

## Verification of the builds card
- ensure that there is no build phase during provisioning of the templated apps

assert that both apps build and run.

### DIFF
--- a/evals/roles/reboot_template_broker/tasks/main.yml
+++ b/evals/roles/reboot_template_broker/tasks/main.yml
@@ -1,0 +1,12 @@
+---
+# tasks file for reboot_template_broker
+- name: Dump current template broker
+  command: oc get clusterservicebroker/template-service-broker -o yaml > /tmp/template-service-broker.yml
+
+- name: Create template-service-broker
+  command: oc create -f /tmp/template-service-broker.yml
+  register: result
+  until: result.stdout.find("created") == 1
+  retries: 200
+  delay: 5
+  failed_when: result.stderr != "" and result.stderr.find("AlreadyExists") == -1
--- a/evals/roles/reboot_template_broker/tasks/main.yml
+++ b/evals/roles/reboot_template_broker/tasks/main.yml
@@ -1,12 +1,16 @@
 ---
 # tasks file for reboot_template_broker
 - name: Dump current template broker
-  command: oc get clusterservicebroker/template-service-broker -o yaml > /tmp/template-service-broker.yml
+  shell: oc get clusterservicebroker template-service-broker -o yaml > /tmp/template-service-broker.yml
+
+
+- name: Delete current template broker
+  shell: oc delete clusterservicebroker template-service-broker
 
 - name: Create template-service-broker
   command: oc create -f /tmp/template-service-broker.yml
   register: result
-  until: result.stdout.find("created") == 1
+  until: '"created" in result.stdout'
   retries: 200
   delay: 5
-  failed_when: result.stderr != "" and result.stderr.find("AlreadyExists") == -1
+  failed_when: result.stderr != '' and 'AlreadyExists' not in result.stderr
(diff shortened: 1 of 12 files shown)


evals/roles/walkthroughs/templates/crud_spboot_example.yml
```diff
@@ -0,0 +1,1092 @@
...
+      - name: SOURCE_REPOSITORY
+        value: https://github.com/witmicko/spboot-example.git
```
aidenkeating: Perhaps we should be creating/using a fork from the integreatly repo for this. In case we need to make changes.

evals/roles/reboot_template_broker/tasks/main.yml
```diff
@@ -0,0 +1,12 @@
+---
+# tasks file for reboot_template_broker
+- name: Dump current template broker
+  shell: oc get clusterservicebroker/template-service-broker -o yaml > /tmp/template-service-broker.yml
+
+- name: Create template-service-broker
```
aidenkeating: Is the aim of this to force refresh the template service broker? If so, shouldn't we be removing it before trying to create it? I'm seeing this check repeat indefinitely because it's getting an "already exists" error and the `until` check is never passing.

witmicko: that's correct, fixing it now

evals/roles/walkthroughs/templates/messaging_work_queue_nodejs.yml
```diff
@@ -0,0 +1,174 @@
...
+    required: true
+    from: '[a-zA-Z0-9]{40}'
+    generate: expression
+objects:
+- apiVersion: v1
+  kind: ConfigMap
+  metadata:
+    name: {{ msg_work_queue_node_service_name }}
+  data:
+    MESSAGING_SERVICE_HOST: "{{ msg_work_queue_node_service_host }}"
```
aidenkeating: Could we have these default to template parameters if they're available? That way, when the webapp provisions these from the Service Catalog it doesn't need to find the secret, just provide extra parameters

evals/roles/walkthroughs/templates/messaging_work_queue_nodejs.yml
```diff
@@ -147,8 +92,8 @@ objects:
           envFrom:
           - configMapRef:
               name: messaging-service
-          image: nodejs-messaging-work-queue-frontend:latest
-          imagePullPolicy: IfNotPresent
+          image: quay.io/integreatly/nodejs-messaging-work-queue:latest
+          imagePullPolicy: Always
```
aidenkeating: Do we want this to be `Always`?

evals/roles/reboot_template_broker/tasks/main.yml
```diff
@@ -0,0 +1,16 @@
+---
+# tasks file for reboot_template_broker
+- name: Dump current template broker
```
aidenkeating: Saw an issue around this failing if the template-service-broker doesn't exist yet. Maybe we should add a loop before this to check whether it exists yet or not

PR #55, opened 2018-09-21
Create an eval user ahead of time in 3Scale, with admin privileges.

david-martin:
This is required so that the eval user can create & modify APIs in the 3Scale dashboard.

The role change will list 3scale users first and see if there's already an evals user.
If there is, make sure it has admin role.
If there isn't, create it, then make sure it has admin role.
This should ensure the user is created on install, and has the correct role, but also be safe to re-run if the user already exists.

To verify after provisioning:

* Go to the 3Scale admin dashboard i.e. `https://3scale-admin.<router-suffix>`
* Login as evals@example.com
* Verify you have the admin role (Cog in top right > Account > Users)

### DIFF
--- a/evals/roles/3scale/defaults/main.yml
+++ b/evals/roles/3scale/defaults/main.yml
@@ -26,7 +26,5 @@ threescale_evals_username: "{{ threescale_evals_email }}"
 threescale_evals_password: Password1
 
 threescale_cluster_admin_email: admin@example.com
-threescale_cluster_admin_username: "{{ threescale_cluster_admin_email }}"
-threescale_cluster_admin_password: Password1
-
-
+threescale_cluster_admin_old_username: admin
+threescale_cluster_admin_new_username: "{{ threescale_cluster_admin_email }}"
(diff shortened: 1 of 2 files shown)


evals/roles/3scale/defaults/main.yml
```diff
@@ -27,6 +27,3 @@ threescale_evals_password: Password1
 
 threescale_cluster_admin_email: admin@example.com
 threescale_cluster_admin_username: "{{ threescale_cluster_admin_email }}"
-threescale_cluster_admin_password: Password1
```
david-martin: this isn't used, so cleaning up

pmccarthy: @david-martin Good spot. Originally we were using valid auth credentials but then we switched to modifying the existing admin user which meant we didn't need to set a password

pmccarthy:
👀

pmccarthy:
@david-martin Successfully provisioned against PDS and verified that both `admin@example.com` and `evals@example.com` could login and had admin privileges.

Once the commits have been rebased this should be good to merge 👍

david-martin:
@pmccarthy rebased

PR #56, opened 2018-09-26
Fix component playbooks

pmccarthy:
**Summary**
Fixing RHSSO and Che install playbooks so that they can run individually.

_Note:_ Che has a requirement on Launcher at the moment so you'll need to provision that first

**Validation**
Run each of the individual component install playbooks as per the README in this PR and ensure that all is working as expected and that the steps listed in the README make sense.


evals/playbooks/install.yml
```diff
@@ -121,8 +121,6 @@
         che_ws_protocol: wss
         che_keycloak_realm: "{{ eval_launcher_sso_realm }}"
         che_keycloak_client_id: che-client
-        che_github_client_id: "{{ eval_github_client_id }}"
-        che_github_client_secret: "{{ eval_github_client_secret }}"
```
pmccarthy: `che_github_client_id` and `che_github_client_secret` are not used anywhere so removing them.

PR #58, opened 2018-09-27
fix some issues with uninstall playbook and delete namespaces

philbrookes:


evals/playbooks/uninstall.yml
```diff
@@ -1,6 +1,33 @@
 ---
 - hosts: localhost
   tasks:
+    - 
+      when: keep_namespaces is undefined or not keep_namespaces | bool
```
JameelB: Might be good to add this to the docs

PR #65, opened 2018-10-01
Updating webapp and launcher role/playbook master url config

pmccarthy:
**Summary**
Updating the way the launcher and webapp roles currently retrieve the openshift master public URL by reading in the master-config rather than using oc cli.


david-martin:
@pmccarthy Just noticed a subtle difference in how the url was set for launcher vs webapp

```
#webapp
oc version | grep Server | cut -d ' ' -f2 | cut -d '/' -f3
tc-in-OpenS-URXYKCDGG1J0-f1be7c4fb230829d.elb.us-west-2.amazonaws.com:443

#launcher
oc version | grep Server | cut -d ' ' -f2
https://tc-in-OpenS-URXYKCDGG1J0-f1be7c4fb230829d.elb.us-west-2.amazonaws.com:443
```

For webapp, it doesn't require the protocol.
We can fix this in a follow up PR.
the cmd `cut -d '/' -f3` could be used to strip the protocol off for webapp to use

PR #66, opened 2018-10-02
update msb image tag version

JameelB:
## Description
Update tag version used by the managed service broker.

This version includes changes from:
- https://github.com/integr8ly/managed-service-broker/pull/2


JameelB:
ping @maleck13. Can be tested in the same cluster as before

PR #68, opened 2018-10-03
Install acme controller and manage TLS routes

mikenairn:
**Description:**
* Add acme playbook and role to install openshift acme controller (https://github.com/tnozicka/openshift-acme)
* Add acme playbook to install and uninstall playbooks
* Default self signed certs to false
* Search for all TLS routes in component namespaces and update annotations.

**Notes**

* Any routes with hosts longer than 64 chars are ignored as letsencrypt will not issue certs for these
* Enmaase routes are all currently set to "passthrough" and further work needs to be done here to get valid certs working for these routes.
* Only dealing with routes created at cluster creation time i.e routes creating when running the install playbook

### DIFF
--- a/evals/playbooks/acme.yml
+++ b/evals/playbooks/acme.yml
@@ -2,3 +2,10 @@
   hosts: local
   roles:
   - acme
+  tasks:
+    - import_role:
+        name: acme
+        tasks_from: test
+      vars:
+        acme_test_namespace: acme-test
+      tags: [ 'never', 'test' ]
--- a/evals/playbooks/acme.yml
+++ b/evals/playbooks/acme.yml
@@ -1,0 +1,4 @@
+- name: Install OpenShift ACME
+  hosts: local
+  roles:
+  - acme
--- a/evals/playbooks/install.yml
+++ b/evals/playbooks/install.yml
@@ -15,9 +15,13 @@
           eval_app_host: "{{ (eval_master_config['content'] | b64decode | from_yaml)['routingConfig']['subdomain'] }}"
           openshift_master_url: "{{ (eval_master_config['content'] | b64decode | from_yaml)['masterPublicURL'] }}"
       tags: ['bootstrap', 'remote']
-        
+    
 - hosts: localhost
   tasks:
+    - name: Install ACME
+      include_role:
+        name: acme
+      tags: ['acme']
     - name: Install Nexus
       include_role:
         name: nexus
@@ -74,7 +78,7 @@
       tags: ['launcher']
     -
       name: Retrieve launcher sso env vars
-      shell: "oc get dc/launcher-sso \
+      shell: "oc get dc/sso \
         -o jsonpath='{.spec.template.spec.containers[0].env[?(@.name==\"{{ item }}\")].value}' \
         -n {{ eval_launcher_namespace }}"
       with_items:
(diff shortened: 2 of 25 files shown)


evals/playbooks/install.yml
```diff
@@ -69,7 +71,7 @@
       tags: ['launcher']
     -
       name: Retrieve launcher sso env vars
-      shell: "oc get dc/launcher-sso \
+      shell: "oc get dc/sso \
```
mikenairn: Having launcher unnecessarily prepended to this always brought the host to over 64 chars on PDS.

evals/roles/acme/tasks/routes.yml
```diff
@@ -0,0 +1,13 @@
...
+  register: get_tls_routes_cmd
+  failed_when: false
+
+- set_fact:
+    routes: "{{ get_tls_routes_cmd.stdout | from_json }}"
+
+- name: Add ACME route annotation
+  shell: oc annotate routes {{ item.metadata.name }} kubernetes.io/tls-acme=true -n {{ namespace }} --overwrite
+  with_items: "{{ routes.get('items') }}"
+  when: item.spec.tls is defined and item.spec.tls.termination != 'passthrough' and item.spec.host|length <= 64
```
mikenairn: This will apply the annotation to any route in the given namespace that is either "edge or "reencrypt"" and the host is within the limits for letsencrypt (64 bytes)

evals/roles/rhsso/templates/original-sso-x509-template.json
```diff
@@ -109,7 +109,8 @@
             "kind": "Route",
             "metadata": {
                 "annotations": {
-                    "description": "Route for application's https service."
+                    "description": "Route for application's https service.",
+                    "kubernetes.io/tls-acme": "true"
```
mikenairn: Don't believe this template is actually used, but added the annotation here anyway. We can remove all these form here if needs be since we always check the routes have the annotation during the run of the install playbook anyway.

evals/roles/webapp/tasks/provision-webapp.yml
```diff
@@ -8,6 +8,9 @@
   shell: oc process -n {{ webapp_namespace }} -f {{ webapp_template }} --param=OPENSHIFT_HOST={{ openshift_master_url }} --param=WEBAPP_IMAGE_TAG={{ webapp_image_tag }}  | oc create -n {{ webapp_namespace }} -f -
   when: webapp_exists_cmd.rc != 0
 
+- name: Add ACME route annotation
+  shell: oc annotate routes {{ webapp_route }} kubernetes.io/tls-acme=true -n {{ webapp_namespace }} --overwrite
```
mikenairn: @mikenairn remove this

evals/playbooks/acme.yml
```diff
@@ -0,0 +1,11 @@
...
+  hosts: local
+  roles:
+  - acme
+  tasks:
+    - import_role:
+        name: acme
+        tasks_from: test
+      vars:
+        acme_test_namespace: acme-test
+      tags: [ 'never', 'test' ]
```
mikenairn: I was under the impression that this meant it would "never" be included unless you explicitly add the "test" tag, but that appears not to be the case on PDS bastian for some reason, worked for me locally.

mikenairn: Will probably just remove all the test stuff

philbrookes:
Verifying locally showed an issue with ratelimits in letsencrypt.

PR #73, opened 2018-10-03
Only create launcher oauthclient the first time to ensure the secret doesn't change

david-martin:
This fixes an issue where multiple runs of the launcher role results in the
launcher oauthclient getting deleted and recreated.
This results in a different oauth client secret every time.
However, the secret is not udpated in the openshift identity
provider in the launcher sso/keycloak.

It doesn't make sense to update the secret just because the role was 
re-run.


pmccarthy:
👀

PR #77, opened 2018-10-05
Adding temp workaround for 3scale RWX PVC Systemstorageclass name

pmccarthy:
**Summary**
Temporary workaround for 3scale install until their AMP template gets updated to add support for storageclassnames

See https://issues.jboss.org/browse/THREESCALE-1323

**Validation**
Run the installer on one of the Tony clusters and ensure 3scale comes up as expected


david-martin:
Verified in lfa3xlh cluster by deleting the 3scale namespace, then running install.
All seems good after

PR #80, opened 2018-10-09
Updating rwx storageclass logic

pmccarthy:
**Summary**
Fixing issue with running installer that does not have an EFS storageclass configured

**Validation**
Run installer on a cluster that is not EFS backed e.g. PDS

```
ansible-playbook -i inventories/hosts playbooks/install.yml
```

Tested this myself on a cluster _without_ the EFS storage class and a cluster _with_ the EFS storage class and both worked as expected without any param changes


mikenairn:
Changes look fine, just verifying on a PDS cluster now.

PR #82, opened 2018-10-09
Seed eval users

mikenairn:
* Add evalsxx sso and 3scale seed user creation, creates a given amount of evaluation users e.g. evals01@example.com, default is 50


evals/roles/3scale/defaults/main.yml
```diff
@@ -25,6 +25,10 @@ threescale_evals_email: evals@example.com
 threescale_evals_username: "{{ threescale_evals_email }}"
 threescale_evals_password: Password1
 
+rhsso_seed_users_email_format: evals%02d@example.com
+rhsso_seed_users_count: 50
+rhsso_seed_users_password: Password1
+
```
mikenairn: Went through various ways of setting these so that we could have a different variable name here, whilst trying to avoid the need for yet another set of eval_* variables. None worked, so just made the vars the same name here as in the rhsso role.

evals/roles/3scale/defaults/main.yml
```diff
@@ -25,6 +25,10 @@ threescale_evals_email: evals@example.com
 threescale_evals_username: "{{ threescale_evals_email }}"
 threescale_evals_password: Password1
 
+rhsso_seed_users_email_format: evals%02d@example.com
```
mikenairn: Can default these to whatever, what makes most sense eval01@example.com, user1@example.com?

evals/roles/3scale/defaults/main.yml
```diff
@@ -25,6 +25,10 @@ threescale_evals_email: evals@example.com
...
+rhsso_seed_users_count: 50
+rhsso_seed_users_password: Password1
```
mikenairn: Currently this is setting a permanent password for all these users, do we want to force a change on first login or are we happy to keep it as is for these users?

evals/roles/3scale/tasks/_user.yml
```diff
@@ -0,0 +1,35 @@
...
+    eval_user_id: "{{ created_eval_user.json.user.id }}"
+  when: created_eval_user is defined and created_eval_user.json is defined
+
+- name: "Promote user to admin: {{ username }}"
+  uri:
+    url: "https://{{ threescale_admin_route }}/admin/api/users/{{ eval_user_id }}/admin.xml"
+    method: PUT
+    body: "access_token={{ threescale_admin_token }}"
+    validate_certs: "{{ threescale_sso_validate_certs }}"
+    status_code: [200]
```
mikenairn: Using an _ in the name here (_user) to indicate that this is a kind of internal/partial task that is only here to be included in another task. Not sure what the convention in Ansible is for this sort of thing tbh, happy to change it to whatever.

evals/roles/3scale/tasks/install.yml
```diff
@@ -16,21 +16,6 @@
...
-
-- name: Generate system-storage pvc template
-  template:
-    src: "system-storage-pvc.yml.j2"
-    dest: /tmp/system-storage-pvc.yml
-
-- name: "Create system-storage PVC with storageClassName: {{ threescale_pvc_system_storageclassname }}"
-  shell: oc create -f /tmp/system-storage-pvc.yml
-  when: threescale_resources_exist.stderr == "No resources found."
-
```
mikenairn: Revert before merge or after https://github.com/integr8ly/installation/pull/80 is merged

evals/roles/3scale/tasks/sso.yml
```diff
@@ -130,4 +96,4 @@
 - name: Delete sso client config file
   file:
     path: /tmp/3scale-sso-client.json
-    state: absent
\ No newline at end of file
+    state: absent
```
mikenairn: 3scale admin user is still being created here, but should probably be moved to users,. if we decide this approach is OK.

evals/roles/3scale/tasks/users.yml
```diff
@@ -0,0 +1,20 @@
+---
+
+- import_tasks: admin_token.yml
```
mikenairn: Moved this to a separate task as it was getting called in several places, maybe a better way to reuse things in Ansible? No idea tbh.

evals/roles/3scale/tasks/users.yml
```diff
@@ -0,0 +1,20 @@
...
+    body: "access_token={{ threescale_admin_token }}&role=admin"
+    validate_certs: "{{ threescale_sso_validate_certs }}"
+    status_code: [200]
+
+- name: Create evaluation user
+  include: _user.yml email={{ threescale_evals_email }} username={{ threescale_evals_username }} password={{ threescale_evals_password }}
+
+- name: Seed evaluation users
+  include: _user.yml email={{ rhsso_seed_users_email_format|format(item|int) }} username={{ email }} password={{ rhsso_seed_users_password }}
+  with_sequence: count={{ rhsso_seed_users_count }}
```
mikenairn: Iterate the _users.yml task for the number of seed users we wanted. Ansible does not make looping multiple tasks easy and you cant loop on a block. This appears to be the best option? Maybe?

evals/roles/rhsso/defaults/main.yml
```diff
@@ -20,6 +20,10 @@ rhsso_evals_email: evals@example.com
 rhsso_evals_username: "{{ rhsso_evals_email }}"
 rhsso_evals_password: Password1
 
+rhsso_seed_users_email_format: evals%02d@example.com
+rhsso_seed_users_count: 50
+rhsso_seed_users_password: Password1
```
mikenairn: vars duplicated here unfortunately, and using the same name in both roles for the reasons described above.

evals/roles/rhsso/tasks/_user.yml
```diff
@@ -0,0 +1,40 @@
+---
+- import_tasks: admin_token.yml
```
mikenairn: The admin token we get back from keycloak is very short lived, 60 seconds i believe. This was previously being called in the client task, and was still valid by the time it got to adding users, but we cant rely on there being less than 60 seconds between that and finishing this, so we have to ask for a new one every time.

evals/roles/rhsso/tasks/_user.yml
```diff
@@ -0,0 +1,40 @@
...
+
+- name: "Create user: {{ username }}"
+  uri:
+    url: "https://{{ rhsso_route }}/auth/admin/realms/{{ rhsso_realm }}/users"
+    method: POST
+    body: "{{ lookup('template', './user.json.j2') }}"
```
mikenairn: Replaced the individual templates and creating local files by just adding this here. The vars the template requires are passed into this (username, password and email) so will be different on each execution depending on what was passed into it.

evals/roles/3scale/tasks/_user.yml
mikenairn: Is there a reason the eval user was being made into an admin on 3scale? Going by the walkthrough steps it appears we have to, but seems a bid odd to do this, especially since any user could login to 3scale and delete another user. @jasonmadigan @david-martin ?

evals/roles/3scale/defaults/main.yml
mikenairn: @jasonmadigan @david-martin Any preference on what these seeded users emails/usernames should look like?

evals/roles/3scale/defaults/main.yml
mikenairn: @jasonmadigan @david-martin ?

evals/roles/rhsso/defaults/main.yml
pmccarthy: My hope here is that we can avoid this type of problem by adding role dependencies in the component playbooks. For example - in this case 3scale expects RHSSO to exist, therefore the 3scale role should depend on rhsso allowing it to read in these vars without duplication.

Ideally we will be able to eliminate the need for eval_ vars altogether.

evals/roles/3scale/tasks/_user.yml
david-martin: They are being made an admin to allow them to create the necessary things in 3scale.
There is a subset of permissions that could be applied, but admin is easier at this time.
Sure, users could login and delete others, but this isn't a concern in the types of controlled environments this will be deployed to.

In future, each user will likely get their own 3scale tenant (and be an admin of that tenant), so this is short lived behaviour.

evals/roles/3scale/defaults/main.yml
jasonmadigan: @mikenairn no hard preference. `eval01@example.com` -> onwards sounds fine

evals/roles/3scale/defaults/main.yml
jasonmadigan: this is configurable in KC afaik, right now we just set

evals/roles/3scale/tasks/_user.yml
```diff
@@ -0,0 +1,35 @@
+---
+
+- name: "Get user ID: {{ username }}"
+  xml:
+    path: /tmp/3scale_users.xml
+    xpath: /users/user[email[text()='{{ email }}']]/id
+    content: text
+  register: eval_user_xml
+  ignore_errors: True
```
pmccarthy: Not directly related to this PR but it would be nice if we could handle these errors rather than displaying them in the install output. Wasn't a big issue with just a couple of users but now that its 50, theres a lot of red showing in the installer which might lead someone to believe that there is a problem

mikenairn: tbh it's off putting to see red output at any point 1 line or 50, i'd rather we removed it if possible.

pmccarthy:
@mikenairn Yeah apologies, I confused myself I think. Not sure where I got evals1 from!

PR #85, opened 2018-10-10
add check before editting master config

witmicko:
Card: https://trello.com/c/DfF9Menr/268-uninstall-script-seems-to-break-etc-origin-master-master-configyaml-configuration


tremes:
Looks good to me.

PR #86, opened 2018-10-10
Add 3scale wt2 route creation

mikenairn:
* Adds pre seeded 3scale routes for each evals user required for walkthrough 2.


pmccarthy:
👀

PR #89, opened 2018-10-12
Add evals admin user

mikenairn:
* Add evals admin user (evals-admin@example.com), has view access on a cluster.

### DIFF
--- a/evals/roles/3scale/defaults/main.yml
+++ b/evals/roles/3scale/defaults/main.yml
@@ -25,6 +25,10 @@ threescale_evals_email: evals@example.com
 threescale_evals_username: "{{ threescale_evals_email }}"
 threescale_evals_password: Password1
 
+rhsso_evals_admin_email: evals-admin@example.com
+rhsso_evals_admin_username: "{{ rhsso_evals_admin_email }}"
+rhsso_evals_admin_password: Password1
+
 rhsso_seed_users_email_format: evals%02d@example.com
 rhsso_seed_users_count: 50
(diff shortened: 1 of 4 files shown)


evals/roles/rhsso/defaults/main.yml
```diff
@@ -20,6 +20,10 @@ rhsso_evals_email: evals@example.com
 rhsso_evals_username: "{{ rhsso_evals_email }}"
 rhsso_evals_password: Password1
 
+rhsso_evals_admin_email: evals-admin@example.com
+rhsso_evals_admin_username: "{{ rhsso_evals_admin_email }}"
+rhsso_evals_admin_password: Password1
```
mikenairn: If we keep adding additional users we might want to consider a different way of specifying them, and there roles if any, rather than constantly adding more variables.

evals/roles/3scale/defaults/main.yml
```diff
@@ -25,6 +25,10 @@ threescale_evals_email: evals@example.com
 threescale_evals_username: "{{ threescale_evals_email }}"
 threescale_evals_password: Password1
 
+rhsso_evals_admin_email: evals-admin@example.com
+rhsso_evals_admin_username: "{{ rhsso_evals_admin_email }}"
+rhsso_evals_admin_password: Password1
```
mikenairn: Duplicated yet again :cry: , Need to look into this as part of any refactoring/improvement work planned for this repo.

PR #90, opened 2018-10-15
Sso logout route expose

witmicko:
- added cluster's subdomain as valid redirectUri to the openshift client in sso
e.g. `*.apps.mogrodni.openshiftworkshop.com*`

- added `http://localhost:3006` as valid sso redirectUri to allow local dev logout
- added `SSO_ROUTE` param to the webapp, this is needed as logout redirectUri 

TODO: update template raw uri after this pr is merged - https://github.com/integr8ly/tutorial-web-app/pull/132

@aidenkeating @pmccarthy @JameelB


evals/playbooks/webapp.yml
```diff
@@ -4,7 +4,7 @@
   tasks:
     - name: Retrieve master public URL
       slurp:
-        src: "{{ rhsso_openshift_master_config_path }}"
+        src: "{{ rhsso_openshift_master_config_path | default(eval_openshift_master_config_path) }}"
```
witmicko: this fixes undefined var when running it standalone

pmccarthy: This is fine for now as we're doing the same in other roles/playbooks. Once we look to refactor this repo ideally we will remove the `eval_` vars altogether.

pmccarthy:
See the same issue as @aidenkeating on a fresh install also

pmccarthy:
@witmicko @aidenkeating I believe I have resolved this issue. As we're looking to get this closed out to create a release tag I went ahead and pushed the fix up to this branch. Just testing everything now. Will ask @aidenkeating to test again once I've finished

witmicko:
@pmccarthy @aidenkeating manually double wildcard worked for me, but it doesnt now.
I ve added a fix where we build actuall expected webapp URL to be used as a redirectUri in the SSO

evals/roles/rhsso/defaults/main.yml
```diff
@@ -38,3 +38,5 @@ rhsso_master_restart_shim_path: /usr/local/bin/master-restart
 rhsso_openshift_webconsole_namespace: openshift-web-console
 rhsso_openshift_webconsole_configmap: webconsole-config
 rhsso_openshift_update_webconsole_configmap: false
+
+eval_webapp_url_prefix: https://tutorial-web-app-webapp
```
pmccarthy: We shouldn't store eval variables in individual roles. If set, they should only be set in the group_vars common file

evals/roles/rhsso/templates/client.json.j2
```diff
@@ -5,7 +5,9 @@
     "enabled": true,
     "clientAuthenticatorType": "client-secret",
     "redirectUris": [
-        "{{ rhsso_redirect_uri }}/*"
+        "{{ eval_webapp_url_prefix }}.{{ apps_redirect_uri }}",
```
pmccarthy: @witmicko Rather than setting a new evals variable, could we not just get the webapp route and set that as the value?

PR #92, opened 2018-10-16
Rerunning msbroker should not fail

sedroche:
Re running the msbroker playbook was failing. Added some fixes.

### DIFF
--- a/evals/roles/msbroker/tasks/main.yml
+++ b/evals/roles/msbroker/tasks/main.yml
@@ -27,17 +27,34 @@
   shell: oc new-project {{ msbroker_namespace }}
   when: msbroker_check_namespace_exists.rc != 0
 
-- name: Check namespace for existing resources
-  shell: oc get all -n {{ msbroker_namespace }}
-  register: msbroker_resources_exist
-
-- name: Create Required CRDs
-  shell: "oc create -f {{ item }}"
+- name: Ensure CRDs do not exist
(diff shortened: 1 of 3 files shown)


evals/roles/msbroker/tasks/main.yml
```diff
@@ -27,17 +27,34 @@
   shell: oc new-project {{ msbroker_namespace }}
   when: msbroker_check_namespace_exists.rc != 0
 
-- name: Check namespace for existing resources
-  shell: oc get all -n {{ msbroker_namespace }}
-  register: msbroker_resources_exist
+- name: Ensure CRDs do not exist
+  shell: "oc delete -f {{ item }}"
```
pmccarthy: Do we always want to remove these? In its current state every re-run of the installer will remove and re-create the CRDs

pmccarthy: Same applies to the clusterrole, clusterrolebinding and the clusterservicebroker

maleck13: I think these are ok to remove and update each time as they would be removed pretty much straight away (IE there is  not running pods etc to wait on)

sedroche: Yep, I only delete/create on resources that are quickly deleted/created. The deployment is still checked to see if it exists.

PR #96, opened 2018-10-19
webapp operator vars

odra:
To be used together with https://github.com/integr8ly/tutorial-web-app-operator/pull/1

Needs the full integreatly collection of services to be running (install.yml playbook).

Usage:

```
ansible-playbook \
-i inventories/hosts \
 playbooks/webapp.yml 
```

### DIFF
--- a/evals/roles/webapp/templates/cr.yaml.j2
+++ b/evals/roles/webapp/templates/cr.yaml.j2
@@ -1,0 +1,16 @@
+apiVersion: "integreatly.org/v1alpha1"
+kind: "WebApp"
+metadata:
+  name: "tutorial-web-app-operator"
+  labels:
+    app: "{{ webapp_app_label }}"
+spec:
+  app_label: "{{ webapp_app_label }}"
+  template:
+    path: "{{ webapp_operator_template_path }}"
+    parameters:
+      IMAGE: "{{ webapp_image_url }}"
+      IMAGE_TAG: "{{ webapp_image_tag }}"
+      OAUTH_CLIENT_ID: "{{ webapp_client_id }}"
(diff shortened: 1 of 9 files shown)


evals/roles/webapp/templates/cr.yaml.j2
```diff
@@ -0,0 +1,16 @@
...
+    app: "{{ webapp_app_label }}"
+spec:
+  app_label: "{{ webapp_app_label }}"
+  template:
+    path: "{{ webapp_operator_template_path }}"
+    parameters:
+      IMAGE: "{{ webapp_image_url }}"
+      IMAGE_TAG: "{{ webapp_image_tag }}"
+      OAUTH_CLIENT_ID: "{{ webapp_client_id }}"
+      OPENSHIFT_HOST: "{{ openshift_host.stdout }}"
```
maleck13: more for the crd and operator side but wondering could this be discovered from the cluster via an API

odra: I has the same thought but I didn't want to remove the param from the template definition in the crd but I was thinking in implementing some kind of special variable in the cr so the operator can identify those vars and retrieve it using the kubernetes/openshift api (some kind of cr  value templating).

maleck13:
@odra seeing an error here 
```

msg="error syncing key (webapp-test/tutorial-web-app-operator): open : no such file or directory"
--
  | time="2018-10-19T12:07:31Z" level=error msg="Error while processing the template: open : no such file or directory"


```

odra:
Where did you see the error? While running ansible?

odra:
@maleck13 I did a considerable amount of changes in the operator yesterday, you may need to build and push its image again.

PR #99, opened 2018-10-22
add retry to nodejs messaging queue

pb82:
Add a retry to the `Add messaging work queue nodejs to the catalog` task to mitigate sporadic timeouts. Ansible evaluates the `until` before the `failed_when`, so the previous condition still applies (as it should).

ping @witmicko

### DIFF
--- a/evals/roles/walkthroughs/tasks/messaging_work_queue_nodejs.yml
+++ b/evals/roles/walkthroughs/tasks/messaging_work_queue_nodejs.yml
@@ -10,3 +10,6 @@
   command: oc create -f /tmp/messaging_work_queue_nodejs.yml -n openshift
   register: out
   failed_when: out.stderr != "" and out.stderr.find("AlreadyExists") == -1
+  until: out.stderr != ""
+  retries: 5
+  delay: 10


pmccarthy:
This is a timing issue where the master services are restarted seconds before we attempt to run this command. I'm happy to have the retries here but ideally we would just move the walkthrough setup steps above the block that restarts the master services in our install.yml playbook.

See: https://github.com/integr8ly/installation/blob/master/evals/playbooks/install.yml#L211-L232

I believe @philbrookes may have a PR issued soon that applies this fix

witmicko:
@pb82 hey, didnt see this and partially repeated it here https://github.com/integr8ly/installation/pull/104/files
spring boot app also needs that repeat.
will your change not cause timeout when template already exists?

pb82:
@pmccarthy Avoiding the retries by moving the task seems preferrable. I can close this PR.

@witmicko If the template already exists it will exhaust the repeats but the final `failed_when` should give it a pass.

witmicko:
@pb82 changed play order and added wait times in my pr here https://github.com/integr8ly/installation/pull/104
I think its better to have both so walkthrough play can be run safely on its own too.
cc @pmccarthy

pb82:
@witmicko Ok, i'm happy to merge it then. A good verification would be to run it twice (or more).

PR #101, opened 2018-10-22
Add user for admin work in Che/Launcher SSO

aidenkeating:
Che has an env var (CHE_SYSTEM_ADMIN_NAME) that can be set to give
a user with a matching username in Keycloak admin permissions in
Che. This user will be needed if we want to set permissions on
objects in Che like stacks (to allow all users to see and able to
use stacks they didn't create).

This change creates a user named admin and sets the env var to use
that name.

@pmccarthy Mind taking a look? I can make this `admin@example.com` either, it just creates an extra step for the `admin@example.com` user when they login to Che. They get a "user already exists" prompt and have to link the accounts by entering the credentials again.

### DIFF
--- a/evals/roles/che/defaults/main.yml
+++ b/evals/roles/che/defaults/main.yml
@@ -11,6 +11,7 @@ che_template_folder: /tmp/che-templates
 che_app_label: che
 che_delete_namespace: false
 che_validate_certs: "{{ eval_sso_validate_certs | default('true') }}"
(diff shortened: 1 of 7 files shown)


aidenkeating:
@maleck13 Mind taking a look?

aidenkeating:
@maleck13 Mind taking one more look, verified on the demo2 cluster

PR #102, opened 2018-10-22
create secret for 3scale master access token

JameelB:
Create a secret which stores the 3Scale master access token during the 3Scale installation process so that it can be accessed when needed.

image


JameelB:
@maleck13 @pmccarthy This has been verified now, is it possible for any of you to merge this? @witmicko  's verification doesn't enable merging for me.

PR #103, opened 2018-10-23
allow username to be overriden

maleck13:
allow the username to be set as a variable to different workshops can align this with their docs


maleck13:
@mikenairn trying this out currently but does it look reasonable

maleck13:
@pmccarthy If this works would like to get it into the workshop clusters.

mikenairn:
@maleck13 Changes look fine, i remember wondering about his when i added these users, but think i was just trying to limit the amount of variables we had all over the place.

maleck13:
verified on PDS cluster able to login with user1 and password also able to access enmasse and fuse

PR #104, opened 2018-10-23
added wait before ading templates

witmicko:
@pmccarthy

### DIFF
--- a/evals/playbooks/install.yml
+++ b/evals/playbooks/install.yml
@@ -207,6 +207,15 @@
       include_role:
         name: amq
       tags: ['amq']
+- hosts: localhost
+  tasks:
+    - name: Install walkthrough 1
+      include_role:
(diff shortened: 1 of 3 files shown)


philbrookes:
Have you tested this against an RHPDS cluster? When I moved that `installation.yml` block in this manner, it caused a weird exception in my ansible-playbook execution

evals/playbooks/install.yml
```diff
@@ -208,6 +208,15 @@
         name: amq
       tags: ['amq']
 
+- hosts: localhost
```
philbrookes: you are already in a hosts: localhost tasks block here, so you should be able to remove lines 211-213

philbrookes: that's very strange isn't it! Can you squash this PR down to one commit, looks good otherwise.

PR #106, opened 2018-10-23
Patching enmasse playbooks to support TLS

pmccarthy:
**Summary**
Temporary patch so that we can support TLS in Enmasse. Once we upgrade to the next release this should no longer be needed.


pmccarthy:
@david-martin Mind approving again please? Had to rebase

PR #108, opened 2018-10-23
Only remove MSB CSB when the MSB deployment isn't found

aidenkeating:
@pmccarthy @maleck13 Mind taking a look? This seems to be the cause of the MSB CSB vanishing. I'm assuming all the deleting is done on purpose? So this change will just run the template to recreate any non-existing components.

Related to https://github.com/RedHatWorkshops/dayinthelife-integration/issues/80


maleck13:
@aidenkeating before merging take a look at https://github.com/integr8ly/installation/pull/109/files I think it does the same thing plus some additional pieces

aidenkeating:
@maleck13 Yep, looks like it does, happy to close this in place of #109

PR #109, opened 2018-10-23
always run msbroker template

philbrookes:
also:

- remove check for msbroker deployment
- change msbroker tag to 1.0.0
- remove enmasse clusterservicebroker


aidenkeating:
:eyes:

PR #110, opened 2018-10-25
Identity provider change challenge to true

tremes:
User can do `oc login <cluster> -u <user> -p <password>` (when the cluster is behind SSO) with this option set to true otherwise it always asks to provide a token (which can be obtained at `<clusterUrl>/oauth/token/request` url which is again behind SSO - so no easy way to login without browser).

I am not a SSO expert and hope this won't break anything. The option is  described at https://docs.openshift.com/container-platform/3.10/install_config/configuring_authentication.html

### DIFF
--- a/evals/roles/rhsso/tasks/identityprovider.yml
+++ b/evals/roles/rhsso/tasks/identityprovider.yml
@@ -36,7 +36,7 @@
     backup: yes
     block: |2
         - name: rh_sso
-          challenge: false
+          challenge: true
           login: true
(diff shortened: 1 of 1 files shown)


tremes:
Forgot to mention that this change will make testing easier.

tremes:
@david-martin @maleck13  Can someone take a look please? I would like to know this decision.

tremes:
@david-martin  OK thanks.

PR #111, opened 2018-10-26
Revise reboot template service broker

JameelB:
These changes includes:
- Modifying the way that the template service broker was restarted. Instead of deleting and creating the template service broker, we can increment the `relistRequest` value within the template service broker resource [1].
- Removing the reboot template service broker role and moving it as a task within the walkthrough role as part of the walkthrough setup.
- Removing the uninstall che and any reference of the eval_action from the install playbook.

The walkthrough setup needs to be separated into a new tasks list due to a limitation in the number of `include_roles` which can be included in a task. This causes a `maximum recursion depth exceeded` error when added with the rest of the tasks list. This is a known issue in the ansible playbook [2].

[1] https://github.com/openshift/ansible-service-broker/blob/master/docs/debugging.md#forcing-a-relist
[2] https://github.com/ansible/ansible/issues/38357


JameelB:
ping @pmccarthy @philbrookes for visibility

philbrookes:
Could you add some steps to verify this, please?

JameelB:
@philbrookes added :+1:

camilamacedo86:
@JameelB could you check if it still required and if yes, could you rebase it with the master for we are able to test it?

camilamacedo86:
@JameelB it has conflicting files. Could you rebase with the master for we are able to check it?

JameelB:
@camilamacedo86 Rebased, mind taking another look?

odra:
@JameelB it needs another rebase (installation is failing due to keycloak operator changes)

JameelB:
@philbrookes @odra rebased against master

evals/roles/walkthroughs/tasks/reboot_template_broker.yml
```diff
@@ -0,0 +1,18 @@
+---
```
david-martin: @JameelB Could all this be simplified to a 1 liner `svcat relist <broker>` command instead?
That cli should be installed out of the box.
https://docs.openshift.com/container-platform/3.11/architecture/service_catalog/service_catalog_cli.html

david-martin: OK. 2 options:

* Use what you already have
* Install svcat. 

The latter might prove troublesome if the host isn't subscribed via rhsm (using `yum install atomic-enterprise-service-catalog-svcat`), but the binary could also be retreived from https://github.com/kubernetes-incubator/service-catalog/releases.

I don't mind if you want to keep what you have, given the extra complexity to install the cli.

JameelB: I think we can just keep the changes done here since it pretty much does the same on what svcat does. Would also allow us to not add extra steps during the uninstall for removing the cli

PR #114, opened 2018-10-27
implement wildcard subdomain support for 3scale

evanshortiss:
@VinayBhalerao @hguerrero can you test this in the domain linked in Slack?

### DIFF
--- a/evals/playbooks/install.yml
+++ b/evals/playbooks/install.yml
@@ -134,7 +134,7 @@
       include_role:
         name: 3scale
       vars:
-        threescale_route_suffix: "{{ eval_app_host }}"
+        threescale_route_suffix: "amp.{{ eval_app_host }}"
       tags: ['3scale']
     - 
       name: Configure 3Scale SSO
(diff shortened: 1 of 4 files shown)


evals/playbooks/install.yml
```diff
@@ -134,15 +134,15 @@
       include_role:
         name: 3scale
       vars:
-        threescale_route_suffix: "{{ eval_app_host }}"
+        threescale_route_suffix: "amp.{{ eval_app_host }}"
```
evanshortiss: `amp` is a prefix used in much of the existing 3scale technical demos and content. It was requested to be added by the team working on the "Day In The Life" content.

evanshortiss:
You can see the new wildcard route in this screenshot.

screen shot 2018-10-26 at 17 21 45

evals/playbooks/install.yml
evanshortiss: Alternatively it seems we could use a modified yaml that adds the `amp` [on this line](https://github.com/3scale/3scale-amp-openshift-templates/blob/2.2.0.GA/amp/amp.yml#L1219)

PR #117, opened 2018-10-30
Fail msb creation on actual errors

JameelB:
Added a `failed_when` to the creation of managed service broker resources to ensure that it catches actual errors other than when resources already exists.

### DIFF
--- a/evals/roles/msbroker/tasks/main.yml
+++ b/evals/roles/msbroker/tasks/main.yml
@@ -70,4 +70,5 @@
   --param=APICURIO_DASHBOARD_URL={{ apicurio_dashboard_url }} \
   --param=IMAGE_ORG={{ msbroker_image_org }} \
   --param=IMAGE_TAG={{ msbroker_image_tag }} | oc create -n {{ msbroker_namespace }} -f -"
-  failed_when: false
+  register: create_msb_resources_cmd
+  failed_when: create_msb_resources_cmd.stderr != '' and 'AlreadyExists' not in create_msb_resources_cmd.stderr


evals/roles/msbroker/tasks/main.yml
```diff
@@ -39,4 +39,5 @@
 
 - name: Create managed service broker template in {{ msbroker_namespace }}
   shell: oc process -n {{ msbroker_namespace }} -f {{ msbroker_template_url }} --param=NAMESPACE={{ msbroker_namespace }} --param=ROUTE_SUFFIX={{ route_suffix }} --param=LAUNCHER_DASHBOARD_URL={{ launcher_dashboard_url }} --param=CHE_DASHBOARD_URL={{ che_dashboard_url }} --param=THREESCALE_DASHBOARD_URL={{ threescale_dashboard_url }} --param=IMAGE_ORG={{ msbroker_image_org }} --param=IMAGE_TAG={{ msbroker_image_tag }} | oc create -n {{ msbroker_namespace }} -f -
-  failed_when: false
+  register: create_msb_resources_cmd
+  failed_when: create_msb_resources_cmd.stderr != '' and 'AlreadyExists' not in create_msb_resources_cmd.stderr
```
camilamacedo86: Should it not better not execute the task when it already exists? Could we not add a validation to check if already exist?

camilamacedo86:
Hi @JameelB,

Is it still a need? If yes, could you rebase with the master for we are able to check/test it? Also, could you add the steps to reproduce this scenario?

trepel:
Hi @JameelB , the code looks good to me. Could you please rebase? I would like to run this against my PDS before approving.

odra:
:eyes:

trepel:
@odra @JameelB I ran the installer twice (with and without managed-service-broker namespace), all good. I think it is ok to merge.

PR #118, opened 2018-10-31
Revert "fix: generate wildcard route using conventional route"

aidenkeating:
This reverts commit 365298d6ae6acbdc1aa52b42ae18bb6d079d45a1.

These changes will instead be placed in the DIL installer.


mikenairn:
@pmccarthy No.

PR #119, opened 2018-10-31
increase cpu limits

witmicko:


odra:
Looks to be a bit old but I would prefer to properly test che limits before merging this + I think those values should be used as roles variables instead of being hardcoded into the template.

camilamacedo86:
Hi @odra, I will try to do this so. :-) Thank you for your collab.

PR #120, opened 2018-10-31
Gitea Setup

JameelB:
Add a Gitea role with an install and uninstall tasks for setting and cleaning up Gitea installation. The following changes are based on the installation steps defined in the [Gitea operator repository](https://github.com/integr8ly/gitea-operator).

### DIFF
--- a/evals/roles/gitea/defaults/main.yml
+++ b/evals/roles/gitea/defaults/main.yml
@@ -1,0 +1,8 @@
+gitea_namespace: gitea
+
+gitea_resources:
+  - https://raw.githubusercontent.com/integr8ly/gitea-operator/master/deploy/service_account.yaml # Will need to point at tagged version
+  - https://raw.githubusercontent.com/integr8ly/gitea-operator/master/deploy/role.yaml # Will need to point at tagged version
+  - https://raw.githubusercontent.com/integr8ly/gitea-operator/master/deploy/role_binding.yaml # Will need to point at tagged version
+  - https://raw.githubusercontent.com/integr8ly/gitea-operator/master/deploy/crds/integreatly_v1alpha1_gitea_crd.yaml # Will need to point at tagged version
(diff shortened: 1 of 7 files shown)


evals/roles/gitea/defaults/main.yml
```diff
@@ -0,0 +1,10 @@
+gitea_namespace: gitea
+
+gitea_resources:
+  - https://raw.githubusercontent.com/integr8ly/gitea-operator/master/deploy/service_account.yaml # Will need to point at tagged version
+  - https://raw.githubusercontent.com/integr8ly/gitea-operator/master/deploy/role.yaml # Will need to point at tagged version
+  - https://raw.githubusercontent.com/integr8ly/gitea-operator/master/deploy/role_binding.yaml # Will need to point at tagged version
+  - https://raw.githubusercontent.com/integr8ly/gitea-operator/master/deploy/crds/integreatly_v1alpha1_gitea_crd.yaml # Will need to point at tagged version
+  - https://raw.githubusercontent.com/integr8ly/gitea-operator/master/deploy/operator.yaml # Will need to point at tagged version
+  - https://raw.githubusercontent.com/integr8ly/gitea-operator/master/deploy/crds/integreatly_v1alpha1_gitea_cr.yaml # Will need to point at tagged version
```
pb82: The cr has currently a hardcoded value for `hostname`: https://github.com/integr8ly/gitea-operator/blob/master/deploy/crds/integreatly_v1alpha1_gitea_cr.yaml#L6

Maybe it would be better to keep a template of this in the installer repo and fill in the hostname dynamically?

JameelB: Good catch, will move that resource to a template

odra:
:eyes:

odra:
Gitea depoyment worked and could access it as well in a rhpds cluster (running install.yml playbook)

PR #121, opened 2018-10-31
3scale operator

mikenairn:
Install 3scale via the 3scale operator https://github.com/integr8ly/3scale-operator

**Verification**

* Run the install playbook and verify everything comes up as expected
* Verify all expected users are created in 3scale
* Verify expected identity provider is created in 3scale
* Verify you can login to 3scale via SSO

**Related PRs:**

* https://github.com/integr8ly/3scale-operator/pull/1

PR #122, opened 2018-11-02
Add back the creation of wt2 routes.

mikenairn:
* Adds back the creation of the wt2 routes for each user, was removed as part of https://github.com/integr8ly/installation/pull/121

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,7 +178,7 @@ Note: 3Scale requires access to ReadWriteMany PVs. As such, it will only work on
 ```shell
 oc login https://<openshift-master-url>
 cd evals/
-ansible-playbook -i inventories/hosts playbooks/3scale.yml -e threescale_route_suffix=amp.<openshift-router-suffix>
+ansible-playbook -i inventories/hosts playbooks/3scale.yml -e threescale_route_suffix=<openshift-router-suffix>
 ```
 
 ##### Run Webapp install playbook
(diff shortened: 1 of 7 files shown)


mikenairn:
@david-martin @pmccarthy Do we also need to revert the amp prefix on the threescale_route_suffix? I'm assuming we do, and remove  wildcardPolicy: "Subdomain" form the CR. This would bring it back to working as it was before https://github.com/integr8ly/installation/pull/114

pmccarthy:
@aidenkeating FYI ^^

pmccarthy:
@mikenairn Looks like we also remove that amp prefix as part of https://github.com/integr8ly/installation/pull/118/files

david-martin:
> Do we also need to revert the amp prefix on the threescale_route_suffix

Yes, as the webapp shows the incorrect route url.

`https://wt2-evals-example-com-3scale.amp.apps.mnairn-65f8.openshiftworkshop.com` instead of
`https://wt2-evals-example-com-3scale.apps.mnairn-65f8.openshiftworkshop.com`

This is based on a placeholder of `https://wt2-{user-sanitized-username}-3scale.{openshift-app-host}` in the webapp

mikenairn:
@david-martin @pmccarthy Yep, reverted that all now also. Re-creating that entire integreatly install from scratch now to verify.

david-martin:
@mikenairn Webapp is showing the correct url now

> Enter https://wt2-evals-example-com-3scale.apps.mnairn-65f8.openshiftworkshop.com in the Staging Public Base URL. Note that this route should point to the shared staging APIcast in the 3scale .....

PR #124, opened 2018-11-05
3scale updates

mikenairn:
* Fix issue with 3scale on clusters using efs, modified version of the check we used to have.
* Change 3scale operator api group integreatly.org -> 3scale.net
* Use variables for 3scale operator image and tag

### DIFF
--- a/evals/roles/3scale/defaults/main.yml
+++ b/evals/roles/3scale/defaults/main.yml
@@ -40,7 +40,7 @@ threescale_cluster_admin_username: "{{ threescale_cluster_admin_email }}"
 
 #Template Params
 threescale_pvc_size_mysql: 100Gi
-threescale_pvc_rwx_storageclassname: ""
+threescale_pvc_rwx_storageclassname: efs
 
 #Operator
 threescale_operator_image_url: 'quay.io/integreatly/3scale-operator'
--- a/evals/roles/3scale/defaults/main.yml
+++ b/evals/roles/3scale/defaults/main.yml
@@ -41,3 +41,7 @@ threescale_cluster_admin_username: "{{ threescale_cluster_admin_email }}"
 #Template Params
 threescale_pvc_size_mysql: 100Gi
 threescale_pvc_rwx_storageclassname: ""
+
+#Operator
+threescale_operator_image_url: 'quay.io/integreatly/3scale-operator'
+threescale_operator_image_tag: '0.0.2'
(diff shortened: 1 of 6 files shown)


evals/roles/3scale/tasks/install.yml
```diff
@@ -9,6 +9,15 @@
...
 
+- name: "Check for storage class: {{ threescale_pvc_rwx_storageclassname }}"
+  shell: oc get storageclass {{ threescale_pvc_rwx_storageclassname }}
+  register: storageclass_exists
+  failed_when: storageclass_exists.stderr != '' and 'NotFound' not in storageclass_exists.stderr
+
+- set_fact:
+    threescale_pvc_rwx_storageclassname: ""
+  when: not storageclass_exists.rc == 0
+
```
mikenairn: Slightly modified check from what we had previously, just resets the value to "" if the current value does not have a corresponding storage class that exists "efs".

evals/roles/3scale/templates/cr.yaml.j2
```diff
@@ -1,4 +1,4 @@
-apiVersion: integreatly.org/v1alpha1
+apiVersion: 3scale.net/v1alpha1
```
mikenairn: Was renamed in the init PR after a review https://github.com/integr8ly/3scale-operator/pull/1

evals/roles/3scale/defaults/main.yml
```diff
@@ -40,4 +40,8 @@ threescale_cluster_admin_username: "{{ threescale_cluster_admin_email }}"
 
 #Template Params
 threescale_pvc_size_mysql: 100Gi
-threescale_pvc_rwx_storageclassname: ""
+threescale_pvc_rwx_storageclassname: efs
```
mikenairn: Would prefer if we didn't default to this, and instead was more an explicit decision to modify it, but seems the easier option to remove the value if its not supported. It's also what we had done previously.

evals/roles/3scale/defaults/main.yml
```diff
@@ -40,4 +40,8 @@ threescale_cluster_admin_username: "{{ threescale_cluster_admin_email }}"
...
+
+#Operator
+threescale_operator_image_url: 'quay.io/integreatly/3scale-operator'
+threescale_operator_image_tag: '0.0.2'
```
mikenairn: Bumped version to avoid breaking master due to the api group change

mikenairn:
Verified on both clusters with efs storage and without. All worked as expected.

PR #125, opened 2018-11-05
update gitea resource links

JameelB:
The Gitea resource links are outdated and is causing the installation to fail. Update the Gitea resources to the link changes according to https://github.com/integr8ly/gitea-operator/pull/4


tremes:
Installation is successful with this update. Thanks @JameelB .

PR #126, opened 2018-11-06
generate gitea token for gitea cr and set webapp gitea token env var

JameelB:
The following changes includes:
- Generating a token to be used for the Gitea `INTERNAL_TOKEN`
- Set value for the `giteaInternalToken` parameter in the Gitea custom resource
- Set webapp Gitea token env var

Verification
1. Run the Gitea/install all playbook.
2. Verify that a `GITEA_TOKEN` env var is set in the web app deployment config.
3. Verify that it sets the gitea token in the `gitea-config` configmap and webapp env var only when a custom resource is created.


evals/roles/gitea/tasks/main.yml
```diff
@@ -5,6 +5,10 @@
   register: gitea_create_namespace_cmd
   failed_when: gitea_create_namespace_cmd.stderr != '' and 'AlreadyExists' not in gitea_create_namespace_cmd.stderr
 
+- name: Generate Gitea token
+  set_fact:
+    gitea_token: "{{ 99999999999999 | random | hash('sha256') }}"
+
 - name: Generate Gitea custom resource template
```
david-martin: It it possible that the token will change with each run of this role?

JameelB: @david-martin yes it will regenerate each time. We can probably set it so that the token is only set when the custom resource is created so that it won't set the env var each time the installer is ran. There should only ever be 1 custom resource and it should fail when it tries to recreate it. Therefore, even if it generates a new token, it should only set it in the custom resource and the env var once.

JameelB:
@david-martin updated according to review

evals/playbooks/install.yml
```diff
@@ -195,14 +195,13 @@
...
       tags: ['rhsso']
-    -
-      name: Setup AMQ Templates
-      include_role:
-        name: amq
-      tags: ['amq']
 
 - hosts: localhost
   tasks:
+    - name: Setup AMQ Templates
```
david-martin: @JameelB was this meant to change in this PR?

david-martin:
I'll give this a spin shortly

evals/playbooks/install.yml
JameelB: @david-martin yes. There's an ansible issue where there's a limit to how many `include_roles` you can have within a task, giving a maximum depth recursion error. I'm not sure if you've come across this but I've come across it when adding tasks to our installer multiple times. https://github.com/ansible/ansible/issues/38357

PR #134, opened 2018-11-08
skip gitea token env var deletion if webapp is unavailable

JameelB:
Skip the Gitea token env var removal from the webapp if the webapp is unavailable.

### DIFF
--- a/evals/roles/gitea/tasks/uninstall.yml
+++ b/evals/roles/gitea/tasks/uninstall.yml
@@ -5,5 +5,11 @@
   failed_when: output.stderr != '' and 'not found' not in output.stderr and 'The system is ensuring all content is removed from this namespace.' not in output.stderr
   changed_when: output.rc == 0
 
+- name: "Check if webapp is available"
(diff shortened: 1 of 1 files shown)


evals/roles/gitea/tasks/uninstall.yml
```diff
@@ -5,5 +5,11 @@
   failed_when: output.stderr != '' and 'not found' not in output.stderr and 'The system is ensuring all content is removed from this namespace.' not in output.stderr
   changed_when: output.rc == 0
 
+- name: "Check if webapp is available"
+  shell: oc get dc/tutorial-web-app -n {{ webapp_namespace }}
+  register: check_webapp_available_cmd
+  failed_when: false
```
pb82: Do we need the `failed_when` here?

JameelB: yep, this is to ensure that this doesn't fail if this returns an error stating the webapp can't be found

PR #135, opened 2018-11-09
Customisation

maleck13:
## Motivation
https://trello.com/c/Olk3Fnq7/234-8-installer-customisation

## What
a new inventory added to a secret in the cluster to allow customisation

## Why
to allow external parties to add customisations they need for things like workshops

## Verification Steps
clone repo
run installer
check the secret is created with the values expected
``` oc get secret manifest -n webapp --template '{{index .data "generated_inventory"}}'  | base64 -D ```

## Progress
- [x] Complete custom inventory file
- [x] Finish doc Todo

### DIFF
--- a/docs/customisation.md
+++ b/docs/customisation.md
@@ -1,0 +1,33 @@
+# Customising an Integreatly Cluster
+
+# Prerequisites 
+
+- Familiar with Ansible
+- Have credentials to login as an admin in the Integreatly cluster
+- Ensure logged into your cluster via oc
+
+
+## Pull down your cluster's inventory file
+
+Each Integreatly cluster has a secret in the webapp namespace with an inventory file that
+can be used to help you add customisations using ansible playbooks.
+
+``` 
+oc get secret manifest -n webapp --template '{{index .data "generated_inventory"}}'  | base64 -D > inventory
+
+```
+
+Each of the components has a set of variables exposed via this inventory file. Each variable has a comment explaining what the value is. If there are things that you need which are missing, please create a issue on the installer repo.
+
+## Limitations 
+
+You do not have ssh access to the cluster so all customisations are limited to what can be done by the user you are logged in as via the OpenShift API and also via the various product APIs.
+
+
+## Example Customisation
+
+See the following repo for some examples of how to do customisations using the in cluster inventory file https://github.com/integr8ly/example-customisations
+
+
+
+
--- a/evals/roles/customisation/tasks/main.yaml
+++ b/evals/roles/customisation/tasks/main.yaml
@@ -93,3 +93,6 @@
 
 - name: Create secret with custom inventory
   shell: oc create secret generic manifest --from-file=/tmp/generated_inventory --type=ansible/inventory -n {{webapp_namespace}}
+
+- name: remove generated template
+    file: path='/tmp/generated_inventory' state=absent  
--- a/evals/roles/customisation/tasks/main.yaml
+++ b/evals/roles/customisation/tasks/main.yaml
@@ -1,0 +1,95 @@
+- name: Retrieve 3Scale admin auth token
+  shell: oc describe dc/system-app -n 3scale | grep ADMIN_ACCESS_TOKEN | tail -1 | awk '{print $2}'
+  register: admin_auth_config
+- set_fact:
+    threescale_admin_token: "{{ admin_auth_config.stdout }}"  
+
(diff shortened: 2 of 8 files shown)

### CODE REVIEW

File: docs/customisation.md [hunk 1]
```diff
@@ -0,0 +1,28 @@
+# Customising an Integreatly Cluster
+
```
maleck13: Ignore this doc it is still in progress

maleck13:
@pmccarthy  ok to take a look @aidenkeating

maleck13:
@pmccarthy If we feel we get enough value from this as is without the rest of the card, I say we go ahead and merge once verified and progress the manifest versions and optional service installation separately

evals/roles/customisation/tasks/main.yaml
```diff
@@ -0,0 +1,95 @@
...
+- name: Retrieve launcher sso hostvars
+  shell: "oc get route/{{launcher_sso_prefix}} -o jsonpath='{.spec.host}' -n {{ launcher_namespace }}"
+  register: launcher_sso_host_cmd
+
+- set_fact: 
+    launcher_sso_route: "{{launcher_sso_host_cmd.stdout}}"          
+
+
+
+- name: Generate custom inventory
```
pmccarthy: Might be worth adding a task to clean up the generated inventory file once the secret has been created

evals/roles/customisation/templates/inventory.j2
```diff
@@ -0,0 +1,109 @@
+[local:vars]
+ansible_connection=local
+
+# The openshift apps suffix
+ocp_apps_domain="{{eval_app_host}}"
```
pmccarthy: I think we should rename `ocp_apps_domain` to something like `ocp_routing_subdomain` as this is how it's referenced in the Openshift docs [1] Apps domain may cause confusion.

[1] https://docs.openshift.com/container-platform/3.11/install_config/router/default_haproxy_router.html#customizing-the-default-routing-subdomain

evals/roles/customisation/templates/inventory.j2
```diff
@@ -0,0 +1,109 @@
...
+seed_user_count="{{eval_seed_users_count}}"
```
pmccarthy: Is `seed_user_count` required since we already have `rhsso_seed_users_count` listed?

pmccarthy:
Verifying changes against PDS (v3.10 cluster)

pmccarthy:
Ran installer against PDS and attempted to get the secret. The command documented to get the secret failed as the `-D` param for base64 is not recognised. Changed to lower case `-d` and it worked.

The PDS clusters run RHEL v7.5

maleck13:
@pmccarthy yeah mac has ```-D``` and linux ```-d``` will update the doc

evals/roles/customisation/templates/inventory.j2
maleck13: wasn't sure on that one. But we seem to have different vars for each?

evals/roles/customisation/templates/inventory.j2
maleck13: sure sounds good

maleck13:
@pmccarthy updated based on review. Let me know if you are happy to approve

evals/roles/customisation/templates/inventory.j2
pmccarthy: Had a second look at the code and it looks like the `rhsso_seed_users_count` gets set to whatever the value of `eval_seed_users_count` is. I think to be safe we should probably leave both for now.

camilamacedo86:
@maleck13 this PR has sintax errors. 

```shell
$ ansible-playbook -i inventories/hosts playbooks/install.yml -e github_client_id=fd300e7e29fc2ed7bb49 -e github_client_secret=19efea317c67b416b536e26f7627b89eecf3fc2e 
ERROR! Syntax Error while loading YAML.
  mapping values are not allowed in this context

The error appears to have been in '/Users/camilamacedo/integr8ly/installation/evals/roles/customisation/tasks/main.yaml': line 98, column 9, but may
be elsewhere in the file depending on the exact syntax problem.

The offending line appears to be:

- name: remove generated template
    file: path='/tmp/generated_inventory' state=absent
        ^ here

```

maleck13:
@camilamacedo86 thanks for spotting this, I will add a fix

PR #145, opened 2018-11-14
Add a route for walkthrough 3 for every user.

david-martin:
This change follows the same pattern as walkthrough 2
with each route exposing the apicast-staging service.

### DIFF
--- a/evals/roles/3scale/tasks/_wt3_route.yml
+++ b/evals/roles/3scale/tasks/_wt3_route.yml
@@ -1,0 +1,31 @@
+---
+
+- set_fact:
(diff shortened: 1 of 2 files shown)


mikenairn:
@david-martin This looks fine to me. It could be refactored further to remove the duplication (pass the route vars into the partial, route name etc..), but ....

Did we ever figure out if we could use the wildcard router instead of creating all these routes?

david-martin:
> Did we ever figure out if we could use the wildcard router instead of creating all these routes?

After chatting with @pmccarthy , it sounds like automating the setup of the wildcard apicast router is non trivial.

PR #146, opened 2018-11-14
Shorten crud app route

aidenkeating:
## Motivation
https://github.com/integr8ly/tutorial-web-app/issues/251

@david-martin Would you mind taking a look? I'm testing this on my cluster now to see how it works with the webapp.


aidenkeating:
@david-martin There isn't, the webapp works on the `spec.to` field. So because the Service name hasn't changed, it'll still work.

david-martin:
thanks @aidenkeating.

closes https://github.com/integr8ly/tutorial-web-app/issues/251

PR #150, opened 2018-11-15
Add location-soap template to the cluster for use by walkthrough 3

david-martin:
## Motivation
https://trello.com/c/tD9Qj6sL/202-13-walkthrough-3-provide-a-walkthrough-which-demonstrates-data-transformation-capabilities-as-a-solution-pattern


evals/roles/walkthroughs/tasks/walkthrough_1.yml
```diff
@@ -3,4 +3,5 @@
   include_tasks: '{{ item  }}.yml'
   with_items:
     - "messaging_work_queue_nodejs"
-    - "crud_spboot_example"
\ No newline at end of file
+    - "crud_spboot_example"
+    - "location_soap"
```
pb82: I suppose it's ok, but the filename is `walkthrough_1.yml` (instead of walkthrough_3).

david-martin: Good spot, I'll add a new file

david-martin: Actually went with a rename of that file to 'walkthroughs.yml'

david-martin:
Did another run on a fresh cluster after rebasing.

Looks good still
```oc get -n openshift templates | grep -i soap
location-soap                                   Location Contacts -  SOAP.                                                         18 (3 blank)      8
```

pb82:
Locatin Soap App is in the catalog after running the `walkthrough` role:

bildschirmfoto 2018-11-16 um 18 32 19

PR #151, opened 2018-11-16
create gitea admin and walkthrough users

pb82:
## Motivation

Create Gitea Admin User, Token and Evals users.

## What

The gitea role was updated to create the users and tokens. Also one property that was also needed in the rhsso and 3scale playbooks was moved to the global vars inventory section (`eval_seed_users_count`)

## Why
After installing gitea we also want to create an admin user, admin token (for use in the webapp) and the correct number of evals users.

## Verification Steps

Delete the gitea workspace on your cluster and run the gitea role in this branch using:

```sh
$ ansible-playbook -i inventories/hosts playbooks/gitea.yml
```

Open the gitea dashboard and use those credentials to log in as `giteaadmin` (default password is Password1). Check that the evals users have been created. Then check (under Settings -> Applications) that the admin user has a token created under his account.

Look for the admin token in the ansible output:

```
TASK [gitea : Print out the gitea admin token] ********************************************************************************************
ok: [127.0.0.1] => {
    "msg": "Gitea admin token is <token sha1 value>"
}
```

Now, in Openshift, go to the Webapp deployment and make sure that the env vars with the name `GITEA_TOKEN` and `GITEA_HOST` are present and has the token value is that of the admin token.

The role should be fully idempotent. You can try to delete some users or the admin token and rerun the role. It should succeed and recreate the deleted resources. Note that the admin password and token will be updated in every rerun.

## Progress

- [x] Finished task

### DIFF
--- a/evals/inventories/group_vars/all/common.yml
+++ b/evals/inventories/group_vars/all/common.yml
@@ -11,3 +11,4 @@ eval_threescale_namespace: 3scale
 eval_launcher_namespace: launcher
 eval_launcher_sso_realm: launcher_realm
 eval_webapp_url_prefix: tutorial-web-app-webapp
+eval_seed_users_count: 50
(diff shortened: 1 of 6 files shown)


JameelB:
:eyes:

david-martin:
> Trying a re-run to check idempotency

Re-ran after removing a user. User was re-created.
:+1:

PR #152, opened 2018-11-19
Improve the Issue Template and PR template

camilamacedo86:
## Motivation
- Get the required data in both cases
- Follow up the previous standards defined for other projects (E.g Aerogear project)

## What
- Add a request to inform OCP version, SO version and Project version/tag in the ISSUE template
- Use the same files defined in the standards project of Aerogear

## Why
- Keep a standard in the projects that we are working with
- Obtained the required info in these situations. ( For example to solve an issue is high important know what version of OCP and installer was used )


camilamacedo86:
Hi @pb82,

It is another simple one. Please, could you check it? Really tks.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,20 +1,35 @@
...
 ## Verification Steps
-Add detailed steps required to check this change. Following an example.
+Add the steps required to check this change. Following an example.
+ 
+1. Go to `XX >> YY >> SS`
+2. Create a new item `N` with the info `X`
+3. Try to edit this item 
+4. Check if in the left menu the feature X is not so long present.
+
+## Checklist:
```
maleck13: why do we need this checklist in the template?

camilamacedo86: The Checklist is the section which we check in order to tell that it was tested by who did the PR and for another person too. It is all FH repos and Aerogear too. ( part of the standard )

PR #154, opened 2018-11-19
fix launcher sso

philbrookes:
## Motivation
Launcher SSO was malfunctioning

- [x] Finished task
- [ ] Reviewed and approved


evals/roles/rhsso/templates/keycloak-realm.json.j2
```diff
@@ -195,6 +214,133 @@
...
+                            "userinfo.token.claim": "true",
+                            "user.attribute": "firstName",
+                            "id.token.claim": "true",
+                            "access.token.claim": "true",
+                            "claim.name": "given_name",
+                            "jsonType.label": "String"
+                        }
+                    },
+                    {
+                        "name": "username",
```
maleck13: do we actually need all this stuff? or is it default? Have you just exported the client as json and added it here? Wondering could it be cut down seems quite noisey

philbrookes: yeah, I think it can be reduced actually, I had forgotten I'd put all this back in as part of my trouble-shooting, I'll test removing it all.

philbrookes: works with this removed.

PR #158, opened 2018-11-19
add http protocol

pb82:
## Motivation

axios behaves  strangely if the protocol is missing. Instead of just adding it to the webapp i think it's better to keep such details (http or https?) to the installer.

### DIFF
--- a/evals/roles/gitea/tasks/main.yml
+++ b/evals/roles/gitea/tasks/main.yml
@@ -107,10 +107,6 @@
   shell: oc get ingress --namespace=gitea --selector='app=gitea' -o jsonpath='{.items[0].spec.rules[0].host}'
   register: gitea_ingress_host
 
-- name: Set gitea host as webapp env var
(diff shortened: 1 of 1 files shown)


evals/roles/gitea/tasks/main.yml
```diff
@@ -108,7 +108,7 @@
   register: gitea_ingress_host
 
 - name: Set gitea host as webapp env var
-  shell: oc set env dc/tutorial-web-app GITEA_HOST="{{ gitea_ingress_host.stdout }}" -n {{ webapp_namespace }} --overwrite=true
+  shell: oc set env dc/tutorial-web-app GITEA_HOST="http://{{ gitea_ingress_host.stdout }}" -n {{ webapp_namespace }} --overwrite=true
   when: gitea_ingress_host.stdout != ''
```
JameelB: Should this also check for if the webapp is installed? This is going to fail if it can't find the tutorial-web-app deployment config.

pb82: Done

PR #170, opened 2018-11-20
update broker template to point to template in master branch

JameelB:
## What
Update the managed service broker template to point to the latest template version.

## Why
Fuse is not being deleted properly and is hanging on as `marked for deletion`. This is because the managed service broker is pointing at an old template that does not contain the `GET Namespace` permission in its cluster role `managed-service`. 

This permission is required for the fix that was added in for version 1.0.2 of the managed service broker to fix broker deprovision issues: https://github.com/integr8ly/managed-service-broker/pull/13

## Verification Steps
1. Run the installer
2. Ensure the Managed service broker is provisioned successfully
3. Ensure the clusterrole, `managed-service`, created has the permission to get namespaces under the apiGroup `""`
4. Ensure that Fuse is successfully deprovisioned even after the Fuse namespace was already deleted.

## Progress

- [x] Update the template to point to the template in the master branch

### DIFF
--- a/evals/roles/msbroker/defaults/main.yml
+++ b/evals/roles/msbroker/defaults/main.yml
@@ -2,7 +2,7 @@
 msbroker_namespace: managed-service-broker
 msbroker_image_org: quay.io/integreatly
 msbroker_image_tag: 1.0.2
-msbroker_template_url: https://raw.githubusercontent.com/integr8ly/managed-service-broker/066322c1be62267ebbe54b9828d8fca2a3cf5633/templates/broker.template.yaml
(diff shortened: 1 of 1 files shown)


philbrookes:
is this the best idea? I feel more comfortable pointing at a commit hash, so we have a determined install process to test, rather than an install process that could potentially change after testing.

odra:
I think we wanted to user "master" references on master branch and then change it to  a commit hash/tag/branch on release branches

JameelB:
I was thinking the same as @odra. Would also prevent having issues like this where links gets outdated on master.

philbrookes:
ok, that's fine. Cheers for the explanation!

PR #172, opened 2018-11-20
fix(unable to install): Issue 162 - base64 -d is an invalid command/param

camilamacedo86:
## Motivation
Add references to relevant tickets or a short description about what motivated you do it. (E.g Trello Card AND/OR ISSUE: https://github.com/integr8ly/tutorial-web-app/issues)

## What
Add an short answer for: What was done in this PR? (E.g Don't allow users has access to the feature X.)

## Why
Add an short answer for: Why it was done? (E.g The feature X was deprecated.)

## How
Add an short answer for: How it was done? (E.g By removing this feature from ... OR By removing just the button but not its implementation ... ) 

## Verification Steps
 
1. Go to `XX >> YY >> SS`
2. Create a new item `N` with the info `X`
3. Try to edit this item 
4. Check if in the left menu the feature X is not so long present.

## Checklist:

## Progress

- [x] Finished task
- [ ] TODO

## Additional Notes

PS.: Add images and/or .gifs to illustrate what was changed if this pull request modifies the appearance/output of something presented to the users.


odra:
it will fail on linux because it uses a lowecase"-d".

It makes more sense to use the native ansible base64 module (`b64decode` and `b64encode`) so it is platform agnostic:

Example:
```
-
  hosts: localhost
  connection: local
  tasks:
    - debug:
        msg: "Encode: {{ 'mystr' | b64encode  }}"
    -
      debug:
        msg: "Decode: {{ 'bXlzdHI=' | b64decode }}"
```

camilamacedo86:
@odra thank you for the suggestion. I am testing it now. :-)

pmccarthy:
As @odra mentioned, OSX requires `-D` while RHEL requires `-d`. We are only supporting the installer on RHEL for now as this is where 99% of our installs take place (running from bastion hosts). If we merge this as is, it will break the installer on RHEL. Running the installer locally is not recommended due to potential differences in local environments.

camilamacedo86:
@pmccarthy 

Hi change here is use the native ansible base64 module (b64decode and b64encode) which I understand that will work in all SO. I tested it in the MacOSx and worked fine. Why do you think it will not work in the RHEL?

pmccarthy:
@camilamacedo86 Apologies, I hadn't seen the updated changes so you can ignore my previous comment

camilamacedo86:
@pmccarthy it is fine. really tks for your collab.

Issue #142, opened 2018-11-13
Uninstallation fails on 3.11


tremes:
## Description
Uninstallation fails on OpenShift 3.11

## Expected Behavior
Successful uninstallation on OS 3.11

## Actual Behavior
Uninstallation fails with following failure because `oc env` command is deprecated in 3.11
TASK [gitea : Delete Gitea token env var from the webapp] **********************
fatal: [127.0.0.1]: FAILED! => {"changed": true, "cmd": "oc env dc/tutorial-web-app GITEA_TOKEN- -n webapp", "delta": "0:00:00.184137", "end": "2018-11-12 12:01:54.853411", "msg": "non-zero return code", "rc": 1, "start": "2018-11-12 12:01:54.669274", "stderr": "Error: unknown command \"env\" for \"oc\"\n\nDid you mean this?\n\tex\n\nRun 'oc --help' for usage.", "stderr_lines": ["Error: unknown command \"env\" for \"oc\"", "", "Did you mean this?", "\tex", "", "Run 'oc --help' for usage."], "stdout": "", "stdout_lines": []}

## Steps to Reproduce
Run uninstall playbook on OS 3.11

PR #174, opened 2018-11-21
use oc set env intead of oc env

JameelB:
## What
Update `oc env` command to `oc set env` for removing the gitea token env var from the web app deployment config.

## Why
`oc env` is deprecated in 3.11 which is causing the uninstaller to fail. We should use `oc set env` instead which should also work for previous oc versions.

## Verification Steps
1. Run the uninstall playbook
2. Ensure that integreatly is uninstalled successfully

## Progress

- [x] Update `oc env` command to `oc set env`

## Additional Notes
Resolves https://github.com/integr8ly/installation/issues/142

### DIFF
--- a/evals/roles/gitea/tasks/uninstall.yml
+++ b/evals/roles/gitea/tasks/uninstall.yml
@@ -11,5 +11,5 @@
   failed_when: false
   
 - name: "Delete Gitea token env var from the webapp"
-  shell: oc env dc/tutorial-web-app GITEA_TOKEN- -n {{ webapp_namespace }}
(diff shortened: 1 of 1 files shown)


odra:
Should we handle both cases (3.10.x and 3.11.x) or are we only supporting 3.11?

JameelB:
@odra This should also work for [3.9.x](https://docs.openshift.com/container-platform/3.9/dev_guide/environment_variables.html),  [3.10.x](https://docs.openshift.com/container-platform/3.10/dev_guide/environment_variables.html) and [3.11.x](https://docs.openshift.com/container-platform/3.11/dev_guide/environment_variables.html) as per their documentation.

Issue #155, opened 2018-11-19
openshift users are not deleted in uninstall


philbrookes:
## Description
When a cluster has the uninstall playbook executed against, running `oc get users` shows any eval users that have ever been used. However they are referencing a keycloak UUID which no longer exists as keycloak is uninstalled.

## Expected Behaviour
any user in openshift that was created as a result of integr8ly should be removed during the uninstall.

i.e. `evalsXX@example.com`, `evals-admin@example.com`, `admin@example.com` and so on.

## Actual Behaviour
After installing integr8ly and logging into some eval users, those users persist through the uninstall. 

## Steps to reproduce
- Provision integr8ly against a fresh cluster
- Login as `evals01@example.com`
- Run uninstall playbook
- execute `oc get users` and see `evals01@example.com`

PR #176, opened 2018-11-21
remove users and idenitities created by rhsso during uninstall

JameelB:
## What
Add tasks to the `rhsso:uninstall` playbook to delete users and identities that was created/used by the rhsso in the integreatly environment.

## Why
When a user logs into the OpenShift console or any services that uses the RHSSO in the integreatly environment, it creates a user and identity resource in OpenShift which references a keycloak UUID. However these resources are not cleaned up properly during uninstallation and may reference keycloak UUID which no longer exists as a result. 

## Verification Steps
1. Have a cluster which has integreatly installed.
2. Login to the OpenShift console using rhsso credentials. (i.e. `admin@example.com`, `evals@example.com`, etc.)
3. Run `oc get users`. This should show users that was used to login via rhsso.
4. Run `oc get identites`. This should show the user idenitites that was used to login via rhsso which references Keycloak UUIDs.
3. Run the uninstall playbook
4. Ensure that all users and identities used to login via rhsso is removed from OpenShift.

## Progress

- [x] Add task for removing users that was used to login to OpenShift via rhsso.
- [x] Add task for removing user identities of users that was used to login to OpenShift via rhsso.

## Additional Notes
Resolves https://github.com/integr8ly/installation/issues/155

### DIFF
--- a/evals/playbooks/uninstall.yml
+++ b/evals/playbooks/uninstall.yml
@@ -76,9 +76,6 @@
       failed_when: output.stderr != '' and 'not found' not in output.stderr and 'The system is ensuring all content is removed from this namespace.' not in output.stderr
       changed_when: output.rc == 0
     - 
-      name: Delete users
-      shell: oc delete user admin@example.com evals@example.com
-    - 
       name: Uninstall the webapp
       include_role:
         name: webapp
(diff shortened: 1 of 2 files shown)


JameelB:
This PR should also fix https://github.com/integr8ly/installation/issues/178

Issue #179, opened 2018-11-21
Uninstall playbook corrupting master-config file when SSO identity provider doesn't exist


pmccarthy:
## Description
Following a failed install where the RHSSO identity provider does not get added to the master-config file on the Openshift master nodes, should the uninstall playbook get executed, it will corrupt the master-config file as it tries to remove the identity provider even though it isn't present in the master-config file.

The end result is that the config file gets corrupted so when the master services are restarted by the uninstall playbook, the master nodes go into an error state.

## Expected Behavior
The uninstall playbook should check to see if the RHSSO identity provider exists in the master-config file before trying to remove it. If the provider does not exist, it should skip the config update and not restart master services.

## Steps to reproduce

- Comment out the install playbook code block [1] that adds the SSO identity provider
- Run the install playbook
- Run the uninstall playbook - this should result in the master-config file getting corrupted and break the master nodes upon restart.

[1] https://github.com/integr8ly/installation/blob/master/evals/playbooks/install.yml#L225-L237

pmccarthy:
@tremes FYI

tremes:
Thanks @pmccarthy. As I said I think this has the highest priority - at least from QE point of view (we simply can't destroy the cluster with each unsuccessful install/uninstall :-))

trepel:
A bit related PR (not fixing this)
https://github.com/integr8ly/installation/pull/187

It might be good to review/incorporate that change while doing this.

tremes:
I would suggest to create some general issue like "Make uninstall script idempotent". WDYT guys? Because now I am facing issue "The error was: 'enmasse_namespace' is undefined".

tremes:
I didn't verify it  but this might be already fixed by https://github.com/integr8ly/installation/commit/ef2c791450a2e83c1ab025d4c59465c962ba9393

trepel:
This has been fixed by https://github.com/integr8ly/installation/pull/190

PR #187, opened 2018-11-22
Fixed keycloak resources removal if not exist

trepel:
## Motivation
Uninstall should not fail if the resources don't exist at all.

## What
It makes sure that the oc delete commands do not fail if the resource types are missing altogether.

## Why
Uninstall should run successfully if Integreatly installation has not finished successfully so it was installed only partially (or not at all).

## How
If removal of keycloak resources fails with message that such resource type does not exist, the uninstall ignores this error and continues.

## Verification Steps
Run the uninstall playbook on cluster where integreatly has not been installed previously. Beware of this bug: https://github.com/integr8ly/installation/issues/179. Currently the uninstall playbook kills the cluster if you did this :)

## Checklist:

## Progress

- [x] Finished task

## Additional Notes

None


trepel:
thanks for quick review @philbrookes

Issue #177, opened 2018-11-21
Installation fails in TASK [launcher : Generate Launcher SSO auth token for admin user] 


tremes:
## Description
Installation (currently installing from master - f26656022a0206e80a2095d5f4a36acaa6c7c8a8) fails with:
```
[...]

PR #188, opened 2018-11-22
[WIP] use secure routes

philbrookes:
## Motivation
https://github.com/integr8ly/installation/issues/177#issuecomment-441012320


philbrookes:
This was related to an install on a cluster that was using an old template.

PR #190, opened 2018-11-22
only reconfigure master if install has altered it

philbrookes:
Uninstall currently changes the master-config whether or not it is required, which causes it to break when run against a cluster where the initial install did not succeed.


evals/roles/rhsso/tasks/remove_identityprovider.yml
```diff
@@ -1,40 +1,49 @@
 ---
-- 
-  name: Remove RH-SSO identity provider from master config
-  shell: "sed -i.bak '/# BEGIN GENERATED BY SSO INSTALLER/,/# END GENERATED BY SSO INSTALLER/d' {{ rhsso_openshift_master_config_path }}"
-  become: yes
+-
+  name: Check master-config has been modified by the installer
+  shell: "cat {{ rhsso_openshift_master_config_path }} | grep 'GENERATED BY SSO INSTALLER' | wc -l"
+  register: output
```
trepel: @philbrookes I thinks `become: yes` needs to be used here as well.

philbrookes:
@trepel, I run the uninstall playbook as master on the master node. I was under the impression this was our standard way to execute now.

PR #191, opened 2018-11-22
Deploying Che with persistent volume by default

pmccarthy:
## Motivation
Che takes a considerable amount of time to scale down and back up again

## What
Prevent Che from scaling down/up on a deployment by setting the volume mount in the default che-server template

## Why
To increase overall time it takes to run the installer and prevent timeout issues occurring in PDS deployments

## How
Set the volume mount in the default che-server template deployment config

## Verification Steps
Run install playbook and ensure that Che comes up as expected with a volume mount of `/data`

## Checklist:

- [x] Code has been verified on RHPDS

## Progress

- [x] Finished task
- [ ] TODO


odra:
Code looks good - I give it a try.

PR #196, opened 2018-11-23
fix undefined variable issue

maleck13:
## How
Set a default namespace
## Verification Steps
run uninstall on cluster should not fail on enmasse


maleck13:
@tremes could you give this a try

tremes:
@maleck13 sure but it will take some time....

PR #198, opened 2018-11-23
move commit to var

philbrookes:
## Motivation
move commit / branch / tag to var

### DIFF
--- a/evals/roles/msbroker/defaults/main.yml
+++ b/evals/roles/msbroker/defaults/main.yml
@@ -2,10 +2,11 @@
 msbroker_namespace: managed-service-broker
 msbroker_image_org: quay.io/integreatly
 msbroker_image_tag: 1.0.2
-msbroker_template_url: https://raw.githubusercontent.com/integr8ly/managed-service-broker/master/templates/broker.template.yaml
+msbroker_git_ref: master
+msbroker_template_url: "https://raw.githubusercontent.com/integr8ly/managed-service-broker/{{ msbroker_git_ref }}/templates/broker.template.yaml"
 msbroker_clusterrole: managed-service
 msbroker_clusterrolebinding: default-cluster-account-managed-service
 msbroker_deployment_name: msb
 msbroker_servicebroker_name: managed-service-broker
 required_crds:
-  - https://raw.githubusercontent.com/integr8ly/syndesis/master/install/operator/deploy/syndesis-crd.yml
+  - "https://raw.githubusercontent.com/integr8ly/syndesis/{{ msbroker_git_ref }}/install/operator/deploy/syndesis-crd.yml"


pb82:
`ref` is more generic than `commit` or `tag` so seems like a good term to use (even though we probably only ever refer to tags).

philbrookes:
@pb82 @mikenairn cool, tried to apply some formalisation to those vars now

evals/roles/gitea/defaults/main.yml
```diff
@@ -1,11 +1,11 @@
...
-  - https://raw.githubusercontent.com/integr8ly/gitea-operator/{{gitea_resource_tag}}/deploy/service_account.yaml
-  - https://raw.githubusercontent.com/integr8ly/gitea-operator/{{gitea_resource_tag}}/deploy/role.yaml
-  - https://raw.githubusercontent.com/integr8ly/gitea-operator/{{gitea_resource_tag}}/deploy/role_binding.yaml
-  - https://raw.githubusercontent.com/integr8ly/gitea-operator/{{gitea_resource_tag}}/deploy/crds/crd.yaml
-  - https://raw.githubusercontent.com/integr8ly/gitea-operator/{{gitea_resource_tag}}/deploy/operator.yaml
+  - https://raw.githubusercontent.com/integr8ly/gitea-operator/{{gitea_git_ref}}/deploy/service_account.yaml
+  - https://raw.githubusercontent.com/integr8ly/gitea-operator/{{gitea_git_ref}}/deploy/role.yaml
+  - https://raw.githubusercontent.com/integr8ly/gitea-operator/{{gitea_git_ref}}/deploy/role_binding.yaml
+  - https://raw.githubusercontent.com/integr8ly/gitea-operator/{{gitea_git_ref}}/deploy/crds/crd.yaml
+  - https://raw.githubusercontent.com/integr8ly/gitea-operator/{{gitea_git_ref}}/deploy/operator.yaml
```
mikenairn: @pb82 Has a PR open doing this same thing here https://github.com/integr8ly/installation/pull/201/files

evals/roles/walkthroughs/defaults/main.yml
```diff
@@ -2,7 +2,7 @@
 # defaults file for walkthroughs
 crud_spboot_runtime_version: 1.3-8
 crud_spboot_source_repo_url: "https://github.com/integr8ly/spboot-example.git"
-crud_spboot_source_repo_ref: "master"
+crud_spboot_git_ref: "master"
```
mikenairn: What is this and where is it used? Is this OK to just rename?

philbrookes: that's a good question, I just did a find and replace, but if it's not used it can go, I guess! Leave it with me.

mikenairn: Maybe we should just revert this for now, not even sure what it is. The msbroker changes look fine, although i have not tried them.

philbrookes:
@mikenairn it seems your review went stale as I forgot about this PR, could you re-review it please?

PR #201, opened 2018-11-23
move operator.yml into installer

pb82:
Move `operator.yml` into the installer and rename `gitea_release_tag` to `gitea_git_ref`.

### DIFF
--- a/evals/roles/gitea/defaults/main.yml
+++ b/evals/roles/gitea/defaults/main.yml
@@ -1,10 +1,12 @@
 gitea_namespace: gitea
-gitea_git_ref: 0.0.2
+gitea_git_ref: v0.0.2
+gitea_operator_image_url: quay.io/integreatly/gitea-operator
+gitea_operator_image_tag: 0.0.2
 gitea_resources:
-  - https://raw.githubusercontent.com/integr8ly/gitea-operator/v{{gitea_git_ref}}/deploy/service_account.yaml
-  - https://raw.githubusercontent.com/integr8ly/gitea-operator/v{{gitea_git_ref}}/deploy/role.yaml
-  - https://raw.githubusercontent.com/integr8ly/gitea-operator/v{{gitea_git_ref}}/deploy/role_binding.yaml
-  - https://raw.githubusercontent.com/integr8ly/gitea-operator/v{{gitea_git_ref}}/deploy/crds/crd.yaml
+  - https://raw.githubusercontent.com/integr8ly/gitea-operator/{{gitea_git_ref}}/deploy/service_account.yaml
+  - https://raw.githubusercontent.com/integr8ly/gitea-operator/{{gitea_git_ref}}/deploy/role.yaml
+  - https://raw.githubusercontent.com/integr8ly/gitea-operator/{{gitea_git_ref}}/deploy/role_binding.yaml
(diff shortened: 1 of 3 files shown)


pb82:
@mikenairn Currently testing this on a new rhpds cluster

evals/roles/gitea/defaults/main.yml
```diff
@@ -1,11 +1,10 @@
...
 gitea_resources:
-  - https://raw.githubusercontent.com/integr8ly/gitea-operator/{{gitea_resource_tag}}/deploy/service_account.yaml
-  - https://raw.githubusercontent.com/integr8ly/gitea-operator/{{gitea_resource_tag}}/deploy/role.yaml
-  - https://raw.githubusercontent.com/integr8ly/gitea-operator/{{gitea_resource_tag}}/deploy/role_binding.yaml
-  - https://raw.githubusercontent.com/integr8ly/gitea-operator/{{gitea_resource_tag}}/deploy/crds/crd.yaml
-  - https://raw.githubusercontent.com/integr8ly/gitea-operator/{{gitea_resource_tag}}/deploy/operator.yaml
+  - https://raw.githubusercontent.com/integr8ly/gitea-operator/v{{gitea_git_ref}}/deploy/service_account.yaml
+  - https://raw.githubusercontent.com/integr8ly/gitea-operator/v{{gitea_git_ref}}/deploy/role.yaml
+  - https://raw.githubusercontent.com/integr8ly/gitea-operator/v{{gitea_git_ref}}/deploy/role_binding.yaml
+  - https://raw.githubusercontent.com/integr8ly/gitea-operator/v{{gitea_git_ref}}/deploy/crds/crd.yaml
```
mikenairn: Any reason you moved the "v" here, i think that should be left where it was, so that can be any git ref i.e. commit hash, bran ch, tag etc..

mikenairn: We are going to be creating a release branch on the installation repo for every new minor release. Would it be more appropriate for the master branch here to always reference the master branch of gitea and release branches will be updated as appropriate?

evals/roles/gitea/templates/operator.yml.j2
```diff
@@ -0,0 +1,37 @@
...
+      name: gitea-operator
+  template:
+    metadata:
+      labels:
+        name: gitea-operator
+    spec:
+      serviceAccountName: gitea-operator
+      containers:
+        - name: gitea-operator
+          image: quay.io/integreatly/gitea-operator:{{ gitea_git_ref }}
```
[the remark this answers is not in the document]
mikenairn: Can we also add variables for everything like https://github.com/integr8ly/installation/blob/master/evals/roles/webapp/templates/operator.yaml.j2#L18

evals/roles/gitea/defaults/main.yml
pb82: The git tags are called `vX.YZ` while the docker images tags only have `X.Y.Z`. We should decide on a common format there too.

mikenairn: Yes, i see why this was changed now having looked at the operator.yml file. Other components in the installer have had these things done with different variables, one for the git ref and one for the image tag, but possibly it could be done with 1 if we standardised our format as you said.

pb82: But then it would no longer be a `git_ref`. I can use two variables for now.

PR #204, opened 2018-11-23
 Add missing step to login into master before running the script

camilamacedo86:
## Motivation
Add missing step to login into master before running the script

## What
Add missing step to login into master before running the script

## Why
Explain how to install

## How
Add info in the README


camilamacedo86:
@maleck13 tks 👍

PR #207, opened 2018-11-24
fix indentation error

maleck13:
## Motivation
Accidentally broke master with a syntax error added after verification step

## Verification Steps
Installation should execute correctly


camilamacedo86:
Checking it now.

PR #209, opened 2018-11-24
reverts the che change and adds s default for the enmasse ns

maleck13:
## Motivation
Fix an issue introduced in issue pr 186 also add a default EnMasse ns

## Verification Steps

Run the installation, it should not fail


maleck13:
verified that the installer runs correctly now and was able to login to the webapp

PR #216, opened 2018-11-27
install manifest 

maleck13:
@mikenairn this is how I am thinking of proceeding with the manifest wanted to share and ensure we are not overlapping

I think we need a follow on task to do the externalisation of the resources / templates. When we do that the version here will be the tag to pull the resource from

**Validation**

Run the installer and ensure it installs everything expected
Ensure there is a manifest and inventory secret in the webapp namespace
Check the inventory looks ok and the manifest looks ok
``` 
oc get secret inventory -n webapp --template '{{index .data "generated_inventory"}}'  | base64 -d ```
oc get secret manifest -n webapp --template '{{index .data "generated_manifest"}}'  | base64 -d 

```

Delete EnMasse and set the manifest var to false run the installer no EnMasse should be installed

Do the same for the other components **Note on che you have to set launcher to false also as they are coupled together**


evals/inventories/group_vars/all/manifest.yaml
```diff
@@ -5,4 +5,10 @@ threescale: true
 threescale_operator_image_tag: '0.0.4'
 threescale_operator_image: 'quay.io/integreatly/3scale-operator'
 threescale_repo: 'https://github.com/integr8ly/3scale-operator'
```
mikenairn: I think it might be better to name this "threescale_operator_repo" to keep it consistent with the others, "threescale_operator" is really the component name here. The likes of the webapp where there is a component and operator will also require this kind of distinction in the naming.

maleck13:
@philbrookes @mikenairn  @pmccarthy Would be good to start getting  a review on this. I know it needs a rebase (pretty complicated one as we seem to have split up install for some reason) but the logic will stay the same

maleck13:
[(5) Installer Customisation](https://trello.com/c/Olk3Fnq7/234-8-installer-customisation)

philbrookes:
:eyes:

maleck13:
Going to squash to 1 commit to help with rebase

camilamacedo86:
Hi @maleck13,

Please, could fill the first comment following the PR template? In this way, will be easier help in this review and also this information can be checked/tracked in the future if be required to understand what, why, and how was done it and the steps required to test it.

maleck13:
@mikenairn Would be good if you got chance to take a look at this

evals/inventories/group_vars/all/manifest.yaml
```diff
@@ -0,0 +1,69 @@
+---
+
+# no particular version of threescale here as it is controlled by the operator. changing the operator version will change the 3scale version
+threescale: true
+threescale_operator_release_tag: '0.0.4'
+threescale_operator_resources: 'https://raw.githubusercontent.com/integr8ly/3scale-operator/{{threescale_operator_release_tag}}/deploy'
```
mikenairn: Are these *_resources variables actually used?

One thing i'm uncertain of here is how you can just add all resources from the operator since there will be a cr in there more than likely that we wouldn't want to create i'd imagine

evals/inventories/group_vars/all/manifest.yaml
```diff
@@ -0,0 +1,69 @@
+---
+
+# no particular version of threescale here as it is controlled by the operator. changing the operator version will change the 3scale version
+threescale: true
+threescale_operator_release_tag: '0.0.4'
```
mikenairn: A single variable i think is fine. However, as long as we use it for both an image tag and a git ref we will need to ensure both exist, which now with the automation we have they should, but not necessarily during dev/testing.

Based on the discussion yesterday though around how these things will be updated, this will probably only be a short term issue since the actual image version will be baked into the operator.yaml and pulled from git.

evals/inventories/group_vars/all/manifest.yaml
maleck13: Some of them are used. They will be important for externalising things. I imagine in the playbook we will have a list of resources to use and then loop over them

mikenairn: Yes, some are used for sure. I assume you are not talking about just adding everything that is in that directory then, but rather another variable will have a list of items you want to create from there?

Do these vars even need to be in here then? Do they not make more sense in the individual roles? It doesn't seem like something that should ever need overriding from the default.

evals/inventories/group_vars/all/manifest.yaml
```diff
@@ -0,0 +1,69 @@
...
+
+
+launcher: true
+# https://github.com/fabric8-launcher/launcher-openshift-templates/issues/46
+# we will need to set this during a release to the latest commit hash on master to ensure it doesn't change under us example: c1efdf1
+launcher_backend_image_tag: 'Set launcher sso host var'
+# we will need to set this during a release to the latest commit hash on master to ensure it doesn't change under us example: 0a941c1
+launcher_frontent_image_tag: 'latest'
+launcher_version: 'master' # note we are using master as this is what is expected by launcher https://github.com/fabric8-launcher/launcher-openshift-templates/issues/46
+launcher_template: 'https://raw.githubusercontent.com/f abric8-launcher/launcher-openshift-templates/{{launcher_version}}/openshift/launcher-template.yaml'
```
mikenairn: space in here "f abric8-launcher" -> fabric8-launcher

mikenairn:
@maleck13 These are the changes i reverted to avoid confusion of where images/git resources were being pulled from https://github.com/integr8ly/installation/pull/216/commits/eb799eeeae8932069fc24bfcd949fa35da13b511

Based on the follow up work we have planned these things will be removed/updated on a component basis as progress this new work about removing checked in resources.

maleck13:
@mikenairn those changes all look reasonable

mikenairn:
@maleck13 Needed rebased yet again, will check an install/uninstall, if all works OK, will approve and merge before the next flood of changes come in.

PR #218, opened 2018-11-29
split up install playbooks

pb82:
When provisioning a RHPDS cluster with Openshift 3.10, it will also install Ansible 2.4. This version of ansible has a bug and will fail when too many instances of `include_role` or `include_tasks` are used: https://github.com/ansible/ansible/issues/23609

While we do say that we require Ansible 2.6+ (https://github.com/integr8ly/installation#prerequisites) we might also want to support 3.10 on RHPDS.

This PR splits up the install playbook into smaller playbooks to reduce the number of includes. This makes it work on Ansible 2.4.


trepel:
Reviewing. I am running on both 3.10 PDS (ansible 2.4.6.0) and 3.11 PDS (ansible 2.6.3). If it is ok I will approve this.

pb82:
@trepel thanks for verifying!

PR #220, opened 2018-11-29
INT-3 patch for some change

jasonmadigan:
INT-3 patch for some change

# This is a test- ignore, I'll delete later.


jasonmadigan:
Test - ignore.

PR #222, opened 2018-11-30
Update webapp cr template params to match template in operator 

mikenairn:
* Update webapp cr template params to match template in operator  https://github.com/integr8ly/tutorial-web-app-operator/blob/master/deploy/template/tutorial-web-app.yml#L5

Currently the webapp image and tag are not overridden in the operator and it always uses master which is the default https://github.com/integr8ly/tutorial-web-app-operator/blob/master/deploy/template/tutorial-web-app.yml#L19:

webappimage


mikenairn:
@odra We need to update these ones as well https://github.com/integr8ly/tutorial-web-app-operator/blob/master/deploy/cr.yaml#L12-L14

PR #228, opened 2018-12-04
Add components meta data file.

mikenairn:
## Motivation

In order to automate the process of creating new integrealy releases, we need some way of knowing what goes into a release and what should be updated.

## What

Add components meta data file.

### DIFF
--- a/COMPONENTS.yaml
+++ b/COMPONENTS.yaml
@@ -46,6 +46,8 @@ components:
     ansible:
       role: webapp
       imageTagVar: webapp_image_tag
+  - name: tutorial-web-app-walkthroughs
+    gitUrl: git@github.com:integr8ly/tutorial-web-app-walkthroughs.git
   - name: apicurio-operator
     gitUrl: git@github.com:integr8ly/apicurio-operator.git
     imageUrl: quay.io/integreatly/apicurio-operator
--- a/COMPONENTS.yaml
+++ b/COMPONENTS.yaml
@@ -1,0 +1,54 @@
+---
+
+components:
+  - name: 3scale-operator
+    gitUrl: git@github.com:integr8ly/3scale-operator.git
+    imageUrl: quay.io/integreatly/3scale-operator
+    ansible:
+      role: 3scale
+      imageTagVar: threescale_operator_image_tag
+  - name: gitea-operator
+    gitUrl: git@github.com:integr8ly/gitea-operator.git
+    imageUrl: quay.io/integreatly/gitea-operator
+    ansible:
+      role: gitea
+      imageTagVar: gitea_operator_image_tag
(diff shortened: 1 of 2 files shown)


COMPONENTS.yaml
```diff
@@ -0,0 +1,54 @@
+---
+
```
mikenairn: An optional global vars file can be specified here:

```
globalVariablesFile:  evals/inventories/group_vars/all/manifest.yaml
```

If no ansible role is specified it will look to update the components variables in this file instead (role will take precedence if it exists, so should be removed if you want to update this global file instead.)

This will need to be done as part of https://github.com/integr8ly/installation/pull/216

COMPONENTS.yaml
```diff
@@ -0,0 +1,54 @@
...
+components:
+  - name: 3scale-operator
+    gitUrl: git@github.com:integr8ly/3scale-operator.git
+    imageUrl: quay.io/integreatly/3scale-operator
+    ansible:
+      role: 3scale
+      imageTagVar: threescale_operator_image_tag
```
mikenairn: role can be removed when we are updating a single manifest file.

mikenairn:
File can be named anything, went with COMPONENTS.yaml in the root of the repo since we had done that previously for other things. Happy to rename/move if we feel it's distracting here. It's a meta data file for a release, so maybe it should be moved into a "release" directory? Seemed little point to adding into the separate "release" repo https://github.com/integr8ly/release, since it's tied to the installation repo and should be versioned along with it.

camilamacedo86:
Hi @mikenairn, 

Please, could you add here in the first comment the info as requested in the PR template? See [](https://github.com/integr8ly/installation/blob/master/.github/PULL_REQUEST_TEMPLATE.md) in this way, we are able to have a better understand of the changes as how it was made and test/check it. For now, for we are able to move forward and in the future if be required for some reason came back to check it. Wdyt? 

Also, is it finish to be tested/checked? If not, could you add WIP in the title in this way we will know this info too :-)

COMPONENTS.yaml
```diff
@@ -0,0 +1,54 @@
...
+    imageUrl: quay.io/integreatly/managed-service-broker
+    ansible:
+      role: msbroker
+      imageTagVar: msbroker_image_tag
+      gitTagVar: msbroker_git_ref
+  - name: syndesis
+    gitUrl: git@github.com:integr8ly/syndesis.git
+    ansible:
+      role: msbroker
+      gitTagVar: msbroker_git_ref
```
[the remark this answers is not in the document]
maleck13: hmm not sure why we would ref a fork here, that may be left over from early days. It will be going away with the work @philbrookes  is doing currently to upgrade fuse online

PR #240, opened 2018-12-06
Add WALKTHROUGH_LOCATIONS env to the webapp operator params

aidenkeating:
There have been changes made in the web app to allow for the location
of walkthroughs to be specified, instead of bundling the walkthroughs
with the web app itself. To specify the walkthrough repo location we
need to set the WALKTHROUGH_LOCATIONS env var.

This change provides a WALKTHROUGH_LOCATIONS parameter to the web
app operator so that it can set the env var when provisioning the
web app.

It also allows for the value of the WALKTHROUGH_LOCATIONS parameter
to be overridden using the 'webapp_walkthrough_locations' var in
Ansible.


aidenkeating:
Related to: https://github.com/integr8ly/tutorial-web-app-operator/pull/14

PR #261, opened 2018-12-12
fix apicurio deletion

odra:
fix apicurio removal by waiting for its cr deletion.

### DIFF
--- a/evals/roles/apicurio/tasks/uninstall.yml
+++ b/evals/roles/apicurio/tasks/uninstall.yml
@@ -1,11 +1,20 @@
 ---
 - name: "Delete Apicurio Deployment Resource"
-  shell: oc delete apicuriodeployment integreatly-apicuriodeployment -n {{ apicurio_namespace | default('apicurio') }}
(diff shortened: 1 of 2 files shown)


evals/roles/apicurio/tasks/uninstall.yml
```diff
@@ -1,11 +1,20 @@
 ---
 - name: "Delete Apicurio Deployment Resource"
-  shell: oc delete apicuriodeployment integreatly-apicuriodeployment -n {{ apicurio_namespace | default('apicurio') }}
+  shell: oc delete apicuriodeployment {{ apicurio_name }} -n {{ apicurio_namespace }}
```
pb82: The name of the CR type `apicuriodeployment` is a bit misleading given that a `deployment` is a very established term in Kube land. Maybe just call it `apicurio`? What do others think about CR naming?

odra: I agree with that but it would require changing the operator code itself which is why I would prefer to address it in a  different pr.

tremes: Agree that the name is a little bit misleading but also agree to adress this in different pr

PR #264, opened 2018-12-14
read master config to match indentation of existing items

jessesarn:
## Motivation
Issue with installation on OSD documented in https://issues.jboss.org/browse/INTLY-334

## What
Avoid hard-coded assumption of indentation level/chars and instead determine the chars used for existing items under the corsAllowedOrigins of actual master-config file.

## Why
This ensures compatibility and consistent behavior with OSD, Pony, RHPDS, and theoretically any other cluster which may use a different number of spaces, and/or tabs, etc.

## How
I had originally wrote the necessary regular expressions needed to use the replace module with dynamic replacement via backrefs instead of blockinfile.  While I was successful in doing so, the regex was already horribly difficult to follow and in order to ensure idepotency, it was only going to get much worse.  Therefore, I separated the logic into a couple steps:
1. read the master config from remote host into variable via slurp (which ends up b64 encoded)
2. decode the contents from b64
3. remove all comment lines from the config (which can be indented differently than the actual items around it without causing problems)
4. extract whatever indentation was used for the existing items under corsAllowedOrigins
5. prepend the same indentation before the line to be inserted
6. use the same blockinfile with markers, etc. as before but no longer using an explicit block indentation via block: |#

## Verification Steps
1. Run Integreatly installer on any cluster which should report the master-config has been changed
2. make sure that the master api services aren't taken offline
3. Run the installer again which should report the master-config as unchanged.
4. Uninstall Integreatly
5. Edit the master-config.yaml and add/remove spaces for all lines in the corsAllowedOrigins section (to match OSD, use 2 spaces in front of all lines, to match Pony/RHPDS, don't use any spaces)
6. Add some comments with different indentation at the top and bottom of the real contents. (To match OSD, there is a comment 
7. Repeat steps 1-3 above

## Checklist:

## Progress

- [x] Finished task
- [ ] TODO


evals/roles/webapp/tasks/cors.yml
```diff
@@ -1,11 +1,19 @@
 ---
-- name: Add webapp route to corsAllowedOrigins in master config
+- name: Read master config
+  slurp:
+    src: /home/jsarnovs/work/master-config.yaml
```
odra: Needs to be rolled back to ` "{{ webapp_openshift_master_config_path }}"`

odra: slurp is also failing, not sure if I am running it from a bastion server:
```
TASK [webapp : Read master config] *******************************************************************************************************************
fatal: [master1.lrossett-cc67.internal]: FAILED! => {"changed": false, "msg": "file not found: /etc/origin/master/master-config.yaml"}
```

odra: Ok it worked, what's missing is a `become: yes` property since only root users can access that file.

odra:
I've applied the changed I mentioned and ran the installer (3x), tested the web app with 3 different users and ran the uninstall playbook - happy to merge this once the changes are addressed.

evals/roles/webapp/tasks/cors.yml
jessesarn: Excellent.  Thanks @odra.  I've fixed both of th issues you pointed out

odra:
Looks good, just running it a last time to be sure everything is in place.

PR #245, opened 2018-12-06
WIP MW enablement

maleck13:
## Motivation
Upgrades of integration components for middleware enablement

## What
update 3scale, fuse, enmasse and amq

## Verification Steps

Run the installer
run through the walkthroughs
## Checklist:

- [ x ] Code has been tested locally by PR requester
- [ ] Changes have been successfully verified by another team member

## Progress

- [ ] Finished task
- [ ] TODO

PR #268, opened 2018-12-18
INTLY-311: Update 3Scale to 2.4

JameelB:
## Motivation
Upgrade 3Scale to 2.4

## What
Isolate relevant 3scale changes from the following PR: https://github.com/integr8ly/installation/pull/245

## Verification Steps
1. Run the installer
2. Ensure that the 3scale operator gets deployed successfully
3. Ensure that 3Scale is deployed successfully
4. Ensure that you are able to login to the 3Scale dashboard.
5. Ensure that all users have the correct username and emails
6. Ensure that all routes are correct

## Checklist:

## Additional Notes
Do not merge PR until the following changes has been merged:
- Update of walkthrough 2. https://github.com/integr8ly/tutorial-web-app-walkthroughs/pull/10
- Updates to the 3scale operator for upgrading to 2.4. https://github.com/integr8ly/3scale-operator/pull/16

### DIFF
--- a/evals/inventories/group_vars/all/manifest.yaml
+++ b/evals/inventories/group_vars/all/manifest.yaml
@@ -1,11 +1,11 @@
 ---
 
-# no particular version of threescale here as it is controlled by the operator. changing the operator version will change the 3scale version
+# threescale version is controlled by the operator. Changing the operator version will change the 3scale version
 threescale: true
-threescale_operator_release_tag: '0.0.4'
+threescale_operator_release_tag: '0.0.5' # This will need to be updated to point to the correct tag once the changes in the 3scale-operator has been merged.
 threescale_operator_resources: 'https://raw.githubusercontent.com/integr8ly/3scale-operator/{{threescale_operator_release_tag}}/deploy'
-threescale_version: '2.2.0.GA' # not currently used but is the version installed by the operator
-#threescale_template: 'https://raw.githubusercontent.com/3scale/3scale-amp-openshift-templates/{{threescale_version}}/amp/amp.yml'
+threescale_version: '2.4.0.GA' # not currently used but is the version installed by the operator
+threescale_template: 'https://raw.githubusercontent.com/3scale/3scale-amp-openshift-templates/{{threescale_version}}/amp/amp.yml'
 
 amq: true
 amq_release_tag: '72-1.1.GA'
--- a/evals/inventories/group_vars/all/manifest.yaml
+++ b/evals/inventories/group_vars/all/manifest.yaml
@@ -2,7 +2,7 @@
 
 # threescale version is controlled by the operator. Changing the operator version will change the 3scale version
 threescale: true
-threescale_operator_release_tag: '0.0.5' # This will need to be updated to point to the correct tag once the changes in the 3scale-operator has been merged.
+threescale_operator_release_tag: 'master'
 threescale_operator_resources: 'https://raw.githubusercontent.com/integr8ly/3scale-operator/{{threescale_operator_release_tag}}/deploy'
 threescale_version: '2.4.0.GA' # not currently used but is the version installed by the operator
 threescale_template: 'https://raw.githubusercontent.com/3scale/3scale-amp-openshift-templates/{{threescale_version}}/amp/amp.yml'
--- a/evals/roles/3scale/defaults/main.yml
+++ b/evals/roles/3scale/defaults/main.yml
@@ -36,12 +36,12 @@ rhsso_seed_users_count: "{{ eval_seed_users_count }}"
 rhsso_seed_users_password: Password1
 
 threescale_cluster_admin_email: admin@example.com
-threescale_cluster_admin_username: "{{ threescale_cluster_admin_email }}"
+threescale_cluster_admin_username: "admin"
 
 #Template Params
 threescale_pvc_size_mysql: 100Gi
 threescale_pvc_rwx_storageclassname: efs
 
 #Operator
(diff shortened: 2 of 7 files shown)


evals/roles/3scale/defaults/main.yml
```diff
@@ -36,12 +36,12 @@ rhsso_seed_users_count: "{{ eval_seed_users_count }}"
 rhsso_seed_users_password: Password1
 
 threescale_cluster_admin_email: admin@example.com
-threescale_cluster_admin_username: "{{ threescale_cluster_admin_email }}"
```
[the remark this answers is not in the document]
JameelB: There's currently an issue with 3Scale 2.4 when passing an `ADMIN_USERNAME` with an email format. This is changed to fix that. https://issues.jboss.org/browse/THREESCALE-1684

evals/inventories/group_vars/all/manifest.yaml
```diff
@@ -2,10 +2,7 @@
 
 # no particular version of threescale here as it is controlled by the operator. changing the operator version will change the 3scale version
 threescale: true
-threescale_operator_release_tag: '0.0.4'
-threescale_operator_resources: 'https://raw.githubusercontent.com/integr8ly/3scale-operator/{{threescale_operator_release_tag}}/deploy'
-threescale_version: '2.2.0.GA' # not currently used but is the version installed by the operator
-#threescale_template: 'https://raw.githubusercontent.com/3scale/3scale-amp-openshift-templates/{{threescale_version}}/amp/amp.yml'
```
mikenairn: Any reason you are removing these?

odra: We are removing those vars from all operators and set the version var in the CR status field from the operator itself (so users don't have control over which 3scale version is installed/deployed in the operator)

evals/roles/customisation/tasks/main.yaml
```diff
@@ -151,6 +151,16 @@
     webapp_route: "{{webapp_route.stdout}}"
   when: webapp
 
+
+- name: Get threescale admin  route
+  shell: oc get route/system-master -o template --template \{\{.spec.host\}\} -n {{ threescale_namespace }}
+  register: threescale_system_admin_route
+  when: threescale
```
mikenairn: Is this meant to be here? Looks to be the same as the one above? This is also the master route rather than the provider admin route, is that correct?

evals/inventories/group_vars/all/manifest.yaml
mikenairn: I think we should probably leave this but set it to "admin". Alternatively remove all usages of it https://github.com/integr8ly/installation/blob/master/evals/roles/3scale/templates/cr.yaml.j2#L11

evals/roles/3scale/defaults/main.yml
```diff
@@ -36,12 +36,12 @@ rhsso_seed_users_count: "{{ eval_seed_users_count }}"
...
 
 #Template Params
 threescale_pvc_size_mysql: 100Gi
 threescale_pvc_rwx_storageclassname: efs
 
 #Operator
-threescale_operator_image_url: 'quay.io/integreatly/3scale-operator'
-threescale_operator_image_tag: master
+threescale_operator_image_url: 'quay.io/jameelb/3scale-operator' # This will need to be updated to point to the correct image once the changes in the 3scale-operator has been merged.
+threescale_operator_image_tag: 0.0.5 # This will need to be updated to point to the correct tag once the changes in the 3scale-operator has been merged.
```
mikenairn: Reminder to update these. Maybe we should have the travis builds push images on forks to make testing PRs easier.

mikenairn: Also, can you set this back to master when it's merged, but set the version in the manifest to 0.0.5 if you bumped the version.

evals/inventories/group_vars/all/manifest.yaml
mikenairn: @odra Yeah sure, that's only the version one though, and i'm not sure really what's @maleck13 idea behind adding that was tbh as the operators were always going to decide what version was being installed. The comment beside it even says as much.

Unless something has been discussed elsewhere, my understanding was this manifest was overriding all the things that set versions during the installation, and the other variables are intended to be used to do this, allbeit possibly not being used at the moment. If we remove these, i'm not even sure what this file is for anymore.

odra: Apologies, i miss read the filename but I am confused now :)

I was using this file as the "single source of truth" to check for new component versions so I agree that those should be kept, at least for now.

JameelB: I removed them at first because they weren't being used, but if that was the case then yes, I agree it's best to put them back in :+1:

evals/roles/customisation/tasks/main.yaml
JameelB: Good catch, looks like a duplicate. This is the master route, will remove this :+1:

evals/inventories/group_vars/all/manifest.yaml
maleck13: While they are not currently being used. The idea of the manifest is as @mikenairn says the manifest will be the override for all components. At the moment 3scale is our own operator so these vars don't do a whole lot right now. But as we move to externalising all of the components and pulling in their resources from version control we will need these variables. We will want to remove any 3scale resources from the installer soon. The vars give information at the moment even if not directly used (IE you can look at this file and see what is installed and what version) in a future iteration we will pull the 3scale (and all products) resources from outside of the installer, these vars will also be used at that point

PR #269, opened 2018-12-19
Ignore missing resource type failure on removal

trepel:
## Motivation
To make uninstall work if the integreatly has never been installed on the cluster, or if the installation failed (partial installation), or if previous uninstall failed (partial uninstall).

## What
Ingnoring missing resource type error if you trying to remove the instances of such resource type.

## Why
See motivation.

## How
Ignoring error output if it says that such resource type does not exist. We had the same stuff already other uninstall roles:
https://github.com/integr8ly/installation/search?q=the+server+doesn&unscoped_q=the+server+doesn

## Verification Steps
Probably just review is OK since it is simple code change we have on other places already, but if you want to you can:
1. run uninstall playbook on cluster with integreatly
2. run uninstall playbook on cluster without integreatly

Note: I have already run the uninstall against my PDS
* https://integreatly-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/view/Nightly/job/clean-uninstallation-pipeline/43/parameters/
(previous two builds failed due to error that this PR is fixing.)

## Progress

- [x] Finished task

### DIFF
--- a/evals/roles/enmasse/tasks/uninstall.yml
+++ b/evals/roles/enmasse/tasks/uninstall.yml
@@ -1,8 +1,15 @@
 ---
+- name: Check that enmasse cluster service broker exists
+  shell: kubectl api-resources --api-group="servicecatalog.k8s.io" -o=name | grep clusterservicebrokers
+  register: result
+  changed_when: false
+  failed_when: false
+  
 - name: "Delete enmasse cluster service broker"
   shell: oc delete clusterservicebroker enmasse
   register: output
-  failed_when: output.stderr != '' and 'not found' not in output.stderr
+  when: result.rc == 0
+  failed_when: output.stderr != '' and 'not found' not in output.stderr and "the server doesn't have a resource type" not in output.stderr
 
 - name: Get enmasse apiServices
   shell: for i in $(oc get apiServices | grep enmasse | awk '{ print $1}');do echo "$i"; done


evals/roles/enmasse/tasks/uninstall.yml
```diff
@@ -2,7 +2,7 @@
 - name: "Delete enmasse cluster service broker"
   shell: oc delete clusterservicebroker enmasse
   register: output
-  failed_when: output.stderr != '' and 'not found' not in output.stderr
+  failed_when: output.stderr != '' and 'not found' not in output.stderr and "the server doesn't have a resource type" not in output.stderr
```
david-martin: Would it be better to see if clusterservicebroker is a supported API resource first, and only run the delete command then?

e.g. check if below command exits with 0 
```
$ kubectl api-resources --api-group="servicecatalog.k8s.io" -o=name | grep clusterservicebrokers
clusterservicebrokers.servicecatalog.k8s.io
```

odra: The above command could be added in a "pre-req" playbook within the enmasse role so it wouldn't even install or uninstall enmasse if the resource is not supported.

trepel: Has this approach any significant advantage? If yes we might want to refactor since there are similar situations in other uninstall roles as well. It is nicer and cleaner for sure.

Could it be that we keep the old way for consistency and create a JIRA for refactoring if it is worth doing it ?

odra: I think it is more reliable to do the way @david-martin mentioned + we can refactor this one in this pr and create a new jira for other components

trepel: ok @david-martin @odra , could be. I'll refactor this one.

trepel: @david-martin @odra I refactored this. Not sure if this is ok though. The zero return code means that we need to remove. The non-zero return code can mean that the resource does not need to be removed, but it can also (in theory) mean that some error happened and thus we don't really know whether we can skip the removal or not - wdyt?

@odra about the "pre-req" playbook - would it be possible that in case of some error during install or uninstall (so we end up with integreatly being partially uninstalled or partially installed which might effectively be very similar situation) the resource is not supported anymore but there are other leftover bits that would need to be removed? If yes, this would not be a good indicator whether we need to execute the uninstall playbook or not.

david-martin: > @david-martin @odra I refactored this. Not sure if this is ok though. The zero return code means that we need to remove. The non-zero return code can mean that the resource does not need to be removed, but it can also (in theory) mean that some error happened and thus we don't really know whether we can skip the removal or not - wdyt?

Ff that command failed because of a reason other than the resource not existing, there's probably something quite wrong with the cluster (e.g. api server not running)
In this case, the uninstaller should fail as it can't reliably do the uninstall.
Is that how you'd see it @trepel

trepel: Hi @david-martin , would you mind having another look? Thanks.

PR #272, opened 2018-12-20
install apicurito from template

philbrookes:
## Motivation
INTLY-307

PR #274, opened 2019-01-03
INTLY-307 install apicurito from template rebase

aidenkeating:
## What
Rebased version of #272 to include patches on master

## Verification Steps
* Run the installer
* Ensure Apicurito is installed instead of Apicurio
* Ensure Apicurito seems to work
* Run the webapp and ensure the ServiceInstance created links to Apicurito

## Checklist:


aidenkeating:
@david-martin @JameelB Would you mind taking a look? This is just a rebase, I've verified things work as expected. I'm not exactly sure what Apicurito should look like compared to Apicurio (It has the same logo and stuff).

Also, the ServiceInstance we provision from the web app is still called apicurio, do we want that to be updated in the MSB also?

Also, should the web app UI be updated to say Apicurito instead of Apicurio? If so, I'll make a PR for both of these.

odra:
let me give it a spin

david-martin:
@aidenkeating Provisions OK.
Getting the same dashboard when i visit the route
image


> I'm not exactly sure what Apicurito should look like compared to Apicurio (It has the same logo and stuff).

This looks like what I've seen in some demos i.e. the 'Apicurio' logo, but 'Apicurito' text in the main area. 


> Also, the ServiceInstance we provision from the web app is still called apicurio, do we want that to be updated in the MSB also?
> 
> Also, should the web app UI be updated to say Apicurito instead of Apicurio? If so, I'll make a PR for both of these.

Lets leave these both as Apicurio for now.
We can revisit and change if it's a problem later.

ping @maleck13 for thoughts on if we should make these changes for the upcoming release? (this shouldn't block this PR from merging )

PR #276, opened 2019-01-04
INTLY-413 Lock launcher frontend and backend images to tags

aidenkeating:
## Motivation
Currently, for the fabric8/launcher-frontend and fabric8/launcher-
backend images we reference the master tag. This tag is updated
every time the master branch of their respective repos are updated.
This is an issue as it could lead to breaking changes being
introduced in an Integreatly release at a later date.

## What
To resolve this, this change locks down the images to the latest
commit tags. These tags shouldn't be updated and removed, which
should make an Integreatly release more stable.

## Verification Steps
* Run the installer
* Ensure the fabric8/launcher-backend image tag is b006e49
* Ensure the fabric8/launcher-frontend image tag is c885979

## Checklist:

### DIFF
--- a/evals/inventories/group_vars/all/manifest.yaml
+++ b/evals/inventories/group_vars/all/manifest.yaml
@@ -38,7 +38,7 @@ launcher: true
 # we will need to set this during a release to the latest commit hash on master to ensure it doesn't change under us example: c1efdf1
 launcher_backend_image_tag: 'b006e49'
 # we will need to set this during a release to the latest commit hash on master to ensure it doesn't change under us example: 0a941c1
-launcher_frontend_image_tag: 'c885979'
+launcher_frontend_image_tag: 'a00064f'
 launcher_version: '7224e235226ffa42135f148bacc409e9f7f402e4'
 launcher_template: 'https://raw.githubusercontent.com/fabric8-launcher/launcher-openshift-templates/{{launcher_version}}/openshift/launcher-template.yaml'
 
--- a/evals/inventories/group_vars/all/manifest.yaml
+++ b/evals/inventories/group_vars/all/manifest.yaml
@@ -39,7 +39,7 @@ launcher: true
 launcher_backend_image_tag: 'b006e49'
 # we will need to set this during a release to the latest commit hash on master to ensure it doesn't change under us example: 0a941c1
 launcher_frontend_image_tag: 'c885979'
-launcher_version: 'master' # note we are using master as this is what is expected by launcher https://github.com/fabric8-launcher/launcher-openshift-templates/issues/46
+launcher_version: '7224e235226ffa42135f148bacc409e9f7f402e4'
 launcher_template: 'https://raw.githubusercontent.com/fabric8-launcher/launcher-openshift-templates/{{launcher_version}}/openshift/launcher-template.yaml'
 
 #not currently used but is the version installed by the operator
--- a/evals/inventories/group_vars/all/manifest.yaml
+++ b/evals/inventories/group_vars/all/manifest.yaml
@@ -36,9 +36,9 @@ fuse_online_operator_resources: 'https://raw.githubusercontent.com/syndesisio/fu
 launcher: true
 # https://github.com/fabric8-launcher/launcher-openshift-templates/issues/46
 # we will need to set this during a release to the latest commit hash on master to ensure it doesn't change under us example: c1efdf1
-launcher_backend_image_tag: 'latest'
+launcher_backend_image_tag: 'b006e49'
 # we will need to set this during a release to the latest commit hash on master to ensure it doesn't change under us example: 0a941c1
-launcher_frontend_image_tag: 'latest'
+launcher_frontend_image_tag: 'c885979'
 launcher_version: 'master' # note we are using master as this is what is expected by launcher https://github.com/fabric8-launcher/launcher-openshift-templates/issues/46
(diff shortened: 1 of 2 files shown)


evals/inventories/group_vars/all/manifest.yaml
```diff
@@ -36,9 +36,9 @@ fuse_online_operator_resources: 'https://raw.githubusercontent.com/syndesisio/fu
 launcher: true
 # https://github.com/fabric8-launcher/launcher-openshift-templates/issues/46
 # we will need to set this during a release to the latest commit hash on master to ensure it doesn't change under us example: c1efdf1
-launcher_backend_image_tag: 'latest'
+launcher_backend_image_tag: 'b006e49'
 # we will need to set this during a release to the latest commit hash on master to ensure it doesn't change under us example: 0a941c1
-launcher_frontend_image_tag: 'latest'
+launcher_frontend_image_tag: 'c885979'
```
david-martin: is this not already handled here for frontend? https://github.com/integr8ly/installation/blob/master/evals/roles/launcher/defaults/main.yml#L4

aidenkeating: @david-martin Didn't even notice that, but I thought it'd be overridden by the group vars. Should the source of truth for these values be the defaults or the manifest?

Also, I'm not familiar with what this value is used for https://github.com/integr8ly/installation/pull/276/files/23d73ef9e1dbe4193d549e635171da6b44769819#diff-6a42777fbc39a86ab76eb44f9a58690fR42 , judging by the comment it should be left at master, is that correct?

tremes: Good question. I find this very confusing as well.

david-martin: > judging by the comment it should be left at master

No. that comment can be removed. It is not the right approach anymore.
Versions should be locked down on master of the installation repo for all components

tremes:
I am going to try this out.

evals/inventories/group_vars/all/manifest.yaml
david-martin: > Should the source of truth for these values be the defaults or the manifest

I don't know. Is this manifest.yml even used?

aidenkeating: It does seem to be used, but the values in `defaults.yml` will override them. At the moment we're only overriding the frontend version. I will lock `launcher_version` to the latest commit.

To me, it makes sense to keep all the versions together, which makes me think it should be removed from `defaults.yml`.

mikenairn: hmm, i didn't think any of these values were actually modifying real values, looks like i was wrong. This should fix the issue, but this manifest file being here, as it is currently, is confusing. 

The intention of this file as i understand it is to lock down specific versions of everything in one place, but as yet none of these values are actually used in the roles (apart from these ones apparently).

aidenkeating:
@tremes It looks like the launcher-frontend value is being overridden elsewhere. It should be `c885979 `. I've resolved that locally, just testing it now

aidenkeating:
@tremes @david-martin I've removed the launcher-frontend tag from `defaults.yml` (launcher-backend wasn't defined in there) and updated the tag values in `manifest.yml` to use the latest commits. Seems to work as expected.

evals/roles/launcher/defaults/main.yml
```diff
@@ -1,8 +1,6 @@
 ---
 launcher_namespace: "{{ eval_launcher_namespace | default('launcher') }}"
 
-launcher_frontend_image_tag: a00064f
```
david-martin: should we use this tag or 'c885979'?
What's the difference, and which were we actually using to date?

evals/inventories/group_vars/all/manifest.yaml
maleck13: @mikenairn is correct. The idea behind the manifest file is to record and control in one place what is installed and what version. Currently not all the values here are used, which I can see ca  cause confusion. Perhaps there is a way to convey this and reduce the confusion? 
At the time of adding this file, it would have been too big a piece of work to refactor the entire installer in one big bang change. So only values that could be easily plugged in where changed. But all of the versions were recorded. 
We intend on changing the installer to source all external resources from VCS so this file will become more and more central as we move to that approach. 
I am open to opinions on how to improve this.

tremes: OK cool. I was thinking about some test comparing particular manifest version to "real runtime" version of an image but I think this needs an explicit definition of versions in the manifest file - e.g Che version (I mean I don't want to parse some "third party" image streams or other yaml files just to get some version string).
Anyway I would like to look into it again and/or we can discuss it.

evals/roles/launcher/defaults/main.yml
aidenkeating: Currently we are using `a00064f `, I'll use `a00064f` for now as we know it works.

aidenkeating:
@mikenairn Are you happy with the changes here? I can confirm these values are used by the launcher role

mikenairn:
@aidenkeating If it's doing what it needs to be doing for a 1.1.0 release then i'm happy.

aidenkeating:
@david-martin @maleck13 I've retested this on a fresh cluster, everything seems to work as expected. It needs an up-to-date approval to merge, mind approving if there's no further work?

PR #279, opened 2019-01-07
Intly 404 release version fixes

mikenairn:
## What
Remove master/latest references where possible and ensure when a release is created that specific versions are being set.

## Why

To avoid breaking changes being made in previous integreatly releases.

## Verification Steps

master:
1. Deploy this branch to an OpenShift cluster
1. Verify everything is working and is using the desired master versions of each product/operator/whatever

Release:
1. Create a new release using the release pipeline
2. Deploy the release to an OpenShift cluster
3. Verify everything is being pulled from a release tag, an explicit commit or version (Not master or latest)

## Checklist:

## Progress

- [x] Finished task
- [ ] TODO

##Related PRs:

https://github.com/integr8ly/tutorial-web-app-operator/pull/18


COMPONENTS.yaml
```diff
@@ -41,6 +36,9 @@ components:
       imageTagVar: webapp_operator_image_tag
   - name: tutorial-web-app-walkthroughs
     gitUrl: git@github.com:integr8ly/tutorial-web-app-walkthroughs.git
+    ansible:
+      role: webapp
+      gitTagVar: webapp_walkthrough_git_ref
```
mikenairn: Unsure what to do here at the moment. I might removes these webapp/walkthough repos entirely and act as if it is fully de-coupled which would mean all the branching/tagging/updating of this repo will need to be done manually.

evals/roles/msbroker/defaults/main.yml
```diff
@@ -9,4 +9,4 @@ msbroker_clusterrolebinding: default-cluster-account-managed-service
 msbroker_deployment_name: msb
 msbroker_servicebroker_name: managed-service-broker
 required_crds:
-  - "https://raw.githubusercontent.com/integr8ly/syndesis/{{ msbroker_git_ref }}/install/operator/deploy/syndesis-crd.yml"
+  - "https://raw.githubusercontent.com/syndesisio/syndesis/syndesis-1.5.8.fuse-720001/install/operator/deploy/syndesis-crd.yml"
```
mikenairn: The crd in the upstream repo never seems to have changed and is exactly the same in all releases. Do we know what release version we should be referencing here? @JameelB?

evals/roles/walkthroughs/defaults/main.yml
```diff
@@ -2,19 +2,22 @@
...
-crud_spboot_source_repo_ref: "master"
+crud_spboot_source_repo_ref: "edf0286"
 crud_spboot_source_repo_dir: "."
 crud_spboot_artifact_copy_args: "'*.jar'"
 crud_spboot_github_webhook_secret: ""
 crud_spboot_maven_mirror_url: ""
 
 msg_work_queue_node_source_repo_url: "https://github.com/integr8ly/nodejs-messaging-work-queue.git"
-msg_work_queue_node_source_repo_ref: "deploy-frontend-without-stock"
+msg_work_queue_node_source_repo_ref: "d30a0be"
```
mikenairn: The majority of these variables don;t appear to be used. @david-martin @aidenkeating Do you know if these repo/ref variables are used anywhere? It seems to me that it just uses images defined in the templates rather than using git.

evals/roles/walkthroughs/defaults/main.yml
```diff
@@ -2,19 +2,22 @@
...
 msg_work_queue_node_service_name: messaging-service
 msg_work_queue_node_service_host: messaging.enmasse-eval.svc
 msg_work_queue_node_service_port: "5672"
 msg_work_queue_node_service_user: "username"
-msg_work_queue_node_service_password: "password"
\ No newline at end of file
+msg_work_queue_node_service_password: "password"
+
+location_soap_source_repo_url: "https://github.com/integr8ly/location-soap.git"
+location_soap_source_repo_ref: "4c97875"
```
mikenairn: The only one that does use git. didn't have the variables defined? Added this based on the current commit at the HEAD of master.

maleck13:
👀

evals/roles/webapp/defaults/main.yml
```diff
@@ -4,7 +4,8 @@ webapp_extra_services_configmap: extra-services
 webapp_client_id: tutorial-web-app
 webapp_route: tutorial-web-app
 webapp_app_label: tutorial-web-app
-webapp_walkthrough_locations: https://github.com/integr8ly/tutorial-web-app-walkthroughs.git#master
+webapp_walkthrough_git_ref: master
+webapp_walkthrough_locations: https://github.com/integr8ly/tutorial-web-app-walkthroughs.git#{{webapp_walkthrough_git_ref}}
```
mikenairn: I was goign to just remove this entirely since the webapp itself should be deciding what version of these are getting included (Like how we removed the webapp image param that started all this), but neither the operator or webapp currently default this value in a way that we can set a version.

evals/roles/msbroker/defaults/main.yml
maleck13: we are using the ```1.5``` tag in the installer

maleck13: https://raw.githubusercontent.com/syndesisio/fuse-online-install/1.5/resources/fuse-online-image-streams.yml

mikenairn: @maleck13 That's a different repo, this one doesn't have a 1.5 tag https://github.com/syndesisio/syndesis

JameelB: as @maleck13 mentioned above we're using tag `1.5`. The [fuse-online-install](https://github.com/syndesisio/fuse-online-install) repo has resources for the productised version.
https://raw.githubusercontent.com/syndesisio/fuse-online-install/1.5/resources/syndesis-crd.yml

mikenairn: Ok so we forked an entirely different repo

evals/roles/walkthroughs/defaults/main.yml
maleck13: The image is already locked down to ``` 3.0.0 ``` see  https://github.com/integr8ly/installation/blob/master/evals/roles/walkthroughs/templates/messaging_work_queue_nodejs.yml#L95 so if we update that line to use the above var and set the var to 3.0.0 should be in good shape

mikenairn: Created separate *_image_tag variables to match how we have naming of other variables in this repo.

evals/roles/walkthroughs/defaults/main.yml
```diff
@@ -1,20 +1,25 @@
 ---
 # defaults file for walkthroughs
+crud_spboot_image_tag: "latest"
```
mikenairn: @david-martin @maleck13 I think we should tag this thing something else in quay but i'm not sure what makes sense? Any ideas?

maleck13:
@mikenairn trying this out on my cluster. Not sure how to do the second part of the verification

evals/roles/launcher/defaults/main.yml
```diff
@@ -33,7 +33,7 @@ launcher_sso_keycloak_client_id: launcher-openshift-users
 launcher_github_default_scopes: "admin:repo_hook,read:org,public_repo"
 
 launcher_catalog_git_repo: https://github.com/integr8ly/launcher-booster-catalog
-launcher_catalog_git_ref: master
+launcher_catalog_git_ref: 033429c
```
maleck13: This seems to need a branch in order to succeed

odra: giving it a try using a release tag

COMPONENTS.yaml
mikenairn: This has been removed, see https://issues.jboss.org/browse/INTLY-464

evals/roles/webapp/defaults/main.yml
```diff
@@ -4,7 +4,6 @@ webapp_extra_services_configmap: extra-services
 webapp_client_id: tutorial-web-app
 webapp_route: tutorial-web-app
 webapp_app_label: tutorial-web-app
-webapp_walkthrough_locations: https://github.com/integr8ly/tutorial-web-app-walkthroughs.git#master
```
mikenairn: This is no longer required since the webapp operator will now have these values set to the correct value by default, see https://github.com/integr8ly/tutorial-web-app-operator/pull/18.

PR #283, opened 2019-01-08
find secure route dynamically

philbrookes:
## Motivation
INTLY-458

## Verification:
https://master.pbrookes-b14d.openshiftworkshop.com/console

evals01 @ https://tutorial-web-app-webapp.apps.pbrookes-b14d.openshiftworkshop.com/
evals01 @ http://launcher-launcher.apps.pbrookes-b14d.openshiftworkshop.com/launch/

### DIFF
--- a/evals/roles/customisation/tasks/main.yaml
+++ b/evals/roles/customisation/tasks/main.yaml
@@ -94,9 +94,13 @@
     components: "{{components}} + [{{che_manifest}}] "
   when: che and launcher
 
+- name: Find encrypted RH-SSO route
+  shell: for route in $(oc get routes -n {{ rhsso_namespace}} | awk '{print $1}' | grep -v 'NAME'); do term=$(oc get route $route -n {{ rhsso_namespace }} -o template --template \{\{.spec.tls.termination\}\}); if [ "$term" == "edge" ] || [ "$term" == "reencrypt" ]; then echo $route; break; fi; done
+  register: rhsso_secure_route_name
+  failed_when: rhsso_secure_route_name.stdout == ''
 
 - name: Get RH-SSO secure route
-  shell: oc get route/sso -o template --template \{\{.spec.host\}\} -n {{ rhsso_namespace }}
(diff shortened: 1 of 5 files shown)


evals/roles/customisation/tasks/main.yaml
```diff
@@ -94,9 +94,13 @@
     components: "{{components}} + [{{che_manifest}}] "
   when: che and launcher
 
+- name: Find encrypted RH-SSO route
+  shell: for route in $(oc get routes -n {{ rhsso_namespace}} | awk '{print $1}' | grep -v 'NAME'); do term=$(oc get route $route -n {{ rhsso_namespace }} -o template --template \{\{.spec.tls.termination\}\}); if [ "$term" == "edge" ] || [ "$term" == "reencrypt" ]; then echo $route; break; fi; done
```
maleck13: wow seems unfortunate that this is how we have to find a secure route. Don't suppose there are annotations added to the route or labels?

maleck13:
is there anything we can do about the repetitive nature of this?

evals/roles/customisation/tasks/main.yaml
philbrookes: yeah, I had a look at that first, didn't see anything I felt comfortable relying on, like "secure: true"

philbrookes:
@maleck13 we could possibly make a python library that does this and call to that? Want me to start working on that now? We can merge this to unblock the release in the mean time?

JameelB:
the launcher sso route is also being used here: https://github.com/integr8ly/installation/blob/master/evals/playbooks/install_services.yml#L55 I think the Che deployment fails if it also doesn't get the correct launcher sso route.

philbrookes:
@jameelb oh thanks, I did overlook that, strange that it still worked for me then. I'll update it now.

philbrookes:
@JameelB this is updated now, and re-provisioned to the same cluster for verification.

maleck13:
verifying on my cluster

maleck13:
was able to login to the webapp

PR #284, opened 2019-01-09
set che_keycloak_client_id to correct value

odra:
Sets `che_keycloak_client_id=che-client` in che role

### DIFF
--- a/evals/playbooks/generate-customisation-inventory.yml
+++ b/evals/playbooks/generate-customisation-inventory.yml
@@ -1,9 +1,8 @@
 ---
 - hosts: localhost
   gather_facts: no
-  vars:
-    components: []
-  roles:
-    - role: customisation
-
-
+  tasks:
+    - include_role:
+        name: customisation
+      vars:
+         components: []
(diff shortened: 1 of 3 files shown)


evals/playbooks/generate-customisation-inventory.yml
```diff
@@ -1,9 +1,9 @@
...
-    components: []
-  roles:
-    - role: customisation
-
-
+  tasks:
+    - include_role:
+        name: customisation
+      vars:
+         che_keycloak_client_id: che-client
```
maleck13: is this not set as a variable somewhere already that we could re use?

odra: I just did the same way we do it in the install_services playbook

odra: also did it this way because it will be cherry-picked into the release branch right? that would require a couple of che tests

odra: on second thought nevermind.. that's already being changed in the install_services playbook

odra: ok I've changed the default var and removed the "customized" values from both playbooks

maleck13:
**Verified**

```che_keycloak_client_id="che-client"```
after installation on my cluster

@odra please ensure to cherry pick back to the release branch

PR #288, opened 2019-01-09
apicurito error handling

odra:
Fixes apicurito role issues when running the install playbook a second time.

Verification:

* Run the install playbook twice: apicurito role should not fail
* Check if apicurito app is running as expected


maleck13:
@odra could you cherry pick to v1.1 branch

PR #290, opened 2019-01-10
fixes webapp patch

odra:
Fixes webapp gitea env vars patching.

Verification:

* Delete the webapp namespace
* Run the install playbook
* The installer should wait for both webapp first deployment and for a second one when gitea role patches new env vars into it
* Check webapp deployment to see if it's correctly  using all gitea env vars

### DIFF
--- a/evals/roles/gitea/tasks/main.yml
+++ b/evals/roles/gitea/tasks/main.yml
@@ -126,9 +126,18 @@
 
 # Only set the Gitea token env var if the gitea custom resource was created
 - name: Set Gitea token env var for the webapp
-  shell: oc set env dc/tutorial-web-app GITEA_TOKEN="{{ gitea_token }}" -n {{ webapp_namespace }} --overwrite=true
-  when: check_webapp_installed_cmd.rc == 0
+  shell: "oc set env dc/tutorial-web-app \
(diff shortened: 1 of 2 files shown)


evals/roles/gitea/tasks/main.yml
```diff
@@ -126,9 +126,18 @@
 
 # Only set the Gitea token env var if the gitea custom resource was created
 - name: Set Gitea token env var for the webapp
-  shell: oc set env dc/tutorial-web-app GITEA_TOKEN="{{ gitea_token }}" -n {{ webapp_namespace }} --overwrite=true
+  shell: "oc set env dc/tutorial-web-app \
+  GITEA_TOKEN='{{ gitea_token }}' \
+  GITEA_HOST='http://{{ gitea_ingress_host.stdout }}' \
+  -n {{ webapp_namespace }} \
+  --overwrite=true"
   when: check_webapp_installed_cmd.rc == 0
```
matskiv: In the condition for GITEA_HOST there was also this part:
```gitea_ingress_host.stdout != ''```
are you sure that it's not needed?

mikenairn:
Seems to be working as expected, deployed to a fresh cluster and there are 2 successful deployments of webapp as expected:

image

PR #297, opened 2019-01-15
Stale sso pods fix

secondsun:
## Motivation
When sso is provisioning, sometimes it fails.  if it fails a stale deploy pod is leftover.  

## What
This PR adds a check for stale deploy pods and deletes them

## Why
see $Motivation

## How
See $What

## Verification Steps
Fail a rhsso deploy
confirm that a stale deploy pod is in the launcher namespace
rerun the install
the deploy should pass with the patch and fail without it

## Checklist:

- [ * ] Code has been tested locally by PR requester

## Progress

- [x] Finished task


secondsun:
@maleck13 Has more context for this.

mikenairn:
@secondsun Assuming this is no longer required. Update and re-open if it is.

Issue #180, opened 2018-11-21
Walkthrough templates persists after uninstall


JameelB:
## Description
Templates used by the walkthroughs that are created in the openshift namespace during integreatly installation are not removed after the uninstall playbook is ran. The task for removing these templates are missing from the `uninstall` playbook.

Templates used by walkthroughs:
- `location-soap`
- `spring-boot-rest-http-crud`
- `nodejs-messaging-work-queue-frontend`

## Expected Behavior
Templates created by the integreatly installer should be removed.

## Actual Behavior
Templates created by the integreatly installer persists in the openshift namespace after then uninstall playbook is ran.

## Environment

* Operating system: Fedora 28
* OpenShift version:
```
oc v3.10.34
kubernetes v1.10.0+b81c8f8
features: Basic-Auth GSSAPI Kerberos SPNEGO

Server https://master1.jbriones-1967.internal:443
openshift v3.10.34
kubernetes v1.10.0+b81c8f8

```
* Ansible version: 2.4.6.0
* Project Version/Tag: master

## Steps to reproduce
1. Run the `uninstall` playbook
2. The templates mentioned above should still exist within the `openshift` namespace.

camilamacedo86:
Hi @JameelB,

It shows be fixed following the result of a test performed today.

<img width="826" alt="screenshot 2018-12-04 at 15 43 35" src="https://user-images.githubusercontent.com/7708031/49453640-75241e80-f7db-11e8-89ca-c501f81f89f9.png">

Could you check this one? If it still and it is not solved for you, then, please, could you provide detailed the steps required to check this namespaces/projects in the *steps to reproduce* section of this issue? How are you checking that these namespaces were not removed?

maleck13:
@trepel to reproduce. Close out if already fixed

trepel:
I was able to reproduce this. I can't find any piece of code that would remove that three templases either.

Templates in question:
https://github.com/integr8ly/installation/tree/master/evals/roles/walkthroughs/templates

There is no corresponding uninstall task there, see
https://github.com/integr8ly/installation/tree/master/evals/roles/walkthroughs/tasks

trepel:
@maleck13 I can create a JIRA ticket for this, not sure what the process is. Do we prefer JIRA to GH issues now?

maleck13:
[...]

PR #302, opened 2019-01-21
remove walkthrough apps from catalog during uninstall

JameelB:
## Additional Information
Resolves https://github.com/integr8ly/installation/issues/180

Removes the templates used by the walkthroughs that are created in the openshift namespace during the installation. Reboot the template broker so that the walkthrough apps are removed from the catalog.

## Verification Steps
1. Have integreatly installed in a cluster
2. Ensure that the following templates can be seen in the `openshift` namespace:
    - `location-soap`
    - `spring-boot-rest-http-crud`
    - `nodejs-messaging-work-queue-frontend`
3. Run the `uninstall` playbook.
4. Ensure that the templates above are removed from the `openshift` namespace
5. Ensure that the following apps are be removed from the service catalog
    - `Spring Boot - REST HTTP and CRUD`
    - `nodejs-messaging-work-queue-frontend`
    - `SOAP :: Location Contacts`

### DIFF
--- a/evals/roles/walkthroughs/tasks/uninstall.yml
+++ b/evals/roles/walkthroughs/tasks/uninstall.yml
@@ -1,0 +1,19 @@
+---
+# Remove walkthrough apps from the catalog
+- name: Remove crud spring boot app from the catalog
+  command: oc delete template/spring-boot-rest-http-crud -n openshift
+  register: output
+  failed_when: output.stderr != '' and 'not found' not in output.stderr
+  changed_when: output.rc == 0
+  
+- name: Remove location soap app from the catalog
+  command: oc delete template/location-soap -n openshift
+  register: output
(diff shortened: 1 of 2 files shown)


evals/roles/walkthroughs/tasks/uninstall.yml
```diff
@@ -0,0 +1,34 @@
+---
+# Remove walkthrough apps from the catalog
+- name: template crud spring boot example app
+  template:
+    src: crud_spboot_example.yml
+    dest: /tmp/crud_spboot_example.yml
+  
+- name: Remove crud spring boot app from the catalog
+  command: oc delete -f /tmp/crud_spboot_example.yml -n openshift
```
david-martin: @JameelB Would it be easier to reference the templates by name instead of needing the file to delete them?

e.g.

```
oc delete template spring-boot-rest-http-crud -n openshift
```

PR #303, opened 2019-01-22
Intly-530

philbrookes:
## Motivation
INTLY-530

### DIFF
--- a/evals/roles/rhsso/defaults/main.yml
+++ b/evals/roles/rhsso/defaults/main.yml
@@ -42,9 +42,9 @@ rhsso_evals_username: "{{ rhsso_evals_email }}"
 rhsso_evals_password: Password1
 
 rhsso_evals_admin_email: evals-admin@example.com
-rhsso_evals_admin_username: "{{ rhsso_evals_admin_email }}"
+rhsso_evals_admin_username: evals-admin
 rhsso_evals_admin_password: Password1
 
 rhsso_cluster_admin_email: admin@example.com
-rhsso_cluster_admin_username: "{{ rhsso_cluster_admin_email }}"
+rhsso_cluster_admin_username: admin
 rhsso_cluster_admin_password: Password1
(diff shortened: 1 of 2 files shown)


tremes:
I'll try to run the installation on my PDS cluster

philbrookes:
can be verified at:
http://gitea.apps.pbrookes-c3e1.openshiftworkshop.com/admin/users
giteaadmin / Password1

evals/roles/rhsso/defaults/main.yml
```diff
@@ -42,9 +42,9 @@ rhsso_evals_username: "{{ rhsso_evals_email }}"
 rhsso_evals_password: Password1
 
 rhsso_evals_admin_email: evals-admin@example.com
-rhsso_evals_admin_username: "{{ rhsso_evals_admin_email }}"
+rhsso_evals_admin_username: evals-admin
```
sedroche: Why do we switch to using hard-coded values here?

evals/playbooks/install_services.yml
```diff
@@ -110,6 +110,16 @@
     -
       name: Expose vars
       include_vars: "../roles/webapp/defaults/main.yml"
+    -
+      name: "Ensure postgresql:9.6 is imported in openshift namespace for enmasse"
```
sedroche: Is this part of the ticket in the PR description?

philbrookes: no, I ran into this issue while testing my own stuff, so had to add this to get the provision to work.

evals/roles/rhsso/defaults/main.yml
philbrookes: as with the other users in here, the name shouldn't be an email, just making this conform to standards.

evals/playbooks/install_services.yml
david-martin: @philbrookes I'd recommend a separate PR for this change so it can be discussed. Is there a jira for this specific issue?
Do we know all the tags of postgres that are in use in the cluster? Why 9.6 only?

philbrookes: ok, I'll move this to it's own PR, I added it to this one as my provision failed without it.

philbrookes: I don't know why it's 9.6. it is the enmasse playbook that looks for that tag

philbrookes:
requires: https://github.com/integr8ly/installation/pull/304 to merge first

PR #304, opened 2019-01-22
Import required image for enmasse

philbrookes:
## Additional Information
enmasse provision is failing when this tag is not imported.

### DIFF
--- a/evals/playbooks/install_services.yml
+++ b/evals/playbooks/install_services.yml
@@ -111,6 +111,16 @@
       name: Expose vars
       include_vars: "../roles/webapp/defaults/main.yml"
     -
+      name: "Ensure postgresql:9.6 is imported in openshift namespace for enmasse"
+      shell: oc -n openshift import-image postgresql:9.6
+      register: result
+      until: result.stdout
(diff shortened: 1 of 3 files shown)


evals/playbooks/install_services.yml
```diff
@@ -110,6 +110,16 @@
     -
       name: Expose vars
       include_vars: "../roles/webapp/defaults/main.yml"
+    -
+      name: "Ensure postgresql:9.6 is imported in openshift namespace for enmasse"
+      shell: oc -n openshift import-image postgresql:9.6
```
david-martin: would it be better to import all tags? (`--all`)
SSO seems to use 9.5 https://github.com/integr8ly/keycloak-operator/blob/0.0.2/deploy/template/sso72-x509-postgresql-persistent.json#L598
Might be worth checking what other things use postgres, and if they use this image stream

philbrookes: yeah, I did go through all the others and they were all using 9.5, but the --all flag still seems like a good idea, I'll add that now.

PR #305, opened 2019-01-22
install unzip as root user

JameelB:
## Additional Information
The command to install unzip has to be run under the root user in the QE cluster.

## Verification Steps
1. Run the installer on a QE cluster
2. Ensure that integreatly is successfully installed


matskiv:
I ended up testing it locally, because QE Pony cluster doesn't have a bastion server.
I will test on PDS once my instance is provisioned.

PR #306, opened 2019-01-23
Ensure componets are added to the secret "manifest"

sedroche:
## Additional Information
Ticket: https://issues.jboss.org/browse/INTLY-524

## Changes Made
The `components` variable as currently declared didn't allow the array to be concatenated resulting in `components` always being an empty array.
Declaring a local `component_manifests` var and adding to that that fixed the problem.
## Verification Steps

I have done a fresh provision of a cluster and the secret is now populated with the component manifests.
https://master.sedroche-2a21.openshiftworkshop.com/console/project/webapp/browse/secrets/manifest

### DIFF
--- a/evals/roles/customisation/tasks/main.yaml
+++ b/evals/roles/customisation/tasks/main.yaml
@@ -1,4 +1,8 @@
 
+- name: Create a list to store the manifest for each component that will be installed
+  set_fact:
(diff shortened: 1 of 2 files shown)


maleck13:
verified the secret in Johns cluster all looks good

PR #307, opened 2019-01-23
Ensure subsequent attempts of patching enmasse auth images succeeds

JameelB:
## Additional Information
Ensure attempts of patching enmasse auth images does not fail when there are no changes.
Versions of oc that are lower than `v3.11.0` returns an exit code 1 when attempting to patch objects with no difference between the existing object. This is a [known issue](https://github.com/kubernetes/kubernetes/issues/58212) and has been [fixed](https://github.com/openshift/origin/pull/20456) for `3.11`.

## Verification Steps
1. Run the installer twice
2. Ensure that the installation succeeds


JameelB:
Closing as our target is 3.11

PR #310, opened 2019-01-25
wait for enmasse csb to be removed before continuing uninstall

JameelB:
## Additional Information
Ensure that the enmasse cluster service broker is removed before continuing with the rest of the uninstall. 

If the enmasse playbook gets ran when the enmasse cluster service broker is in the process of being removed, the playbook won't try to recreate this. This may cause the issue below in the webapp which is a result of the enmasse clusterservicebroker not being available:

```
The instance references a ClusterServiceClass that does not exist. References a non-existent ClusterServiceClass {ClusterServiceClassExternalName:"amq-online-standard"} or there is more than one (found: 0)
```

## Verification Steps
1. Run the `uninstall` playbook
2. Ensure that it waits for the enmasse csb to be removed before continuing


sedroche:
:eye: :eye:

PR #311, opened 2019-01-25
INTLY-615 Include playbook for the application monitoring operator

aidenkeating:
Currently, the application monitoring operator is not included in
the installer.

This change includes the application monitoring operator in the
installer, including a role for the application monitoring
operator and a standalone playbook.

Verification:
- Run the playbooks/install_application_monitoring.yml playbook
against a cluster. Ensure the operator is setup correctly.
- Run the playbook again and ensure it is skipped.
- Run the playbooks/install.yml playbook and ensure the monitoring
playbook is executed too.
- Setting application_monitoring to false in the manifest should
force skipping the run of the playbook.

### DIFF
--- a/evals/playbooks/install.yml
+++ b/evals/playbooks/install.yml
@@ -20,6 +20,7 @@
 - import_playbook: "./install_infra.yml"
 - import_playbook: "./install_services.yml"
 - import_playbook: "./install_webapp.yml"
+- import_playbook: "./install_middleware_monitoring.yml"
 - import_playbook: "./generate-customisation-inventory.yml"
 
 - hosts: localhost
--- a/evals/roles/middleware_monitoring/defaults/main.yml
+++ b/evals/roles/middleware_monitoring/defaults/main.yml
@@ -1,0 +1,30 @@
+---
+monitoring_namespace: middleware-monitoring
+monitoring_tmp_dir: /tmp
+monitoring_repo_raw: "https://raw.githubusercontent.com/integr8ly/application-monitoring-operator/{{ middleware_monitoring_operator_version | default('master') }}"
+# Resources to create via the oc tool.
+monitoring_resources:
+- "{{ monitoring_repo_raw }}/deploy/operator_roles/service_account.yaml"
+- "{{ monitoring_repo_raw }}/deploy/operator_roles/role.yaml"
+- "{{ monitoring_repo_raw }}/deploy/operator_roles/role_binding.yaml"
+- "{{ monitoring_repo_raw }}/deploy/crds/ApplicationMonitoring.yaml"
+- "{{ monitoring_repo_raw }}/deploy/examples/ApplicationMonitoring.yaml"
+- "{{ monitoring_repo_raw }}/deploy/operator.yaml"
+
+monitoring_resource_templates:
+# Grafana resources
+- "grafana_crd.yml"
+- "grafana_cluster_role.yml"
+- "grafana_cluster_role_binding.yml"
+- "grafana_dashboard_crd.yml"
+# Prometheus resources
+- "prometheus_crd.yml"
+- "prometheus_alert_manager_crd.yml"
+- "prometheus_service_monitor_crd.yml"
+- "prometheus_rule_crd.yml"
+- "prometheus_cluster_role_binding.yml"
+- "prometheus_cluster_role.yml"
+- "prometheus_operator_cluster_role_binding.yml"
+- "prometheus_operator_cluster_role.yml"
+- "alert_manager_cluster_role_binding.yml"
+- "alert_manager_cluster_role.yml"
--- a/evals/roles/middleware_monitoring/templates/grafana_cluster_role_binding.yml.j2
+++ b/evals/roles/middleware_monitoring/templates/grafana_cluster_role_binding.yml.j2
@@ -1,0 +1,12 @@
+apiVersion: authorization.openshift.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: grafana-operator
+roleRef:
+  name: grafana-operator-cluster-role
+subjects:
+  - kind: ServiceAccount
+    name: grafana-serviceaccount
+    namespace: "{{ monitoring_namespace }}"
+userNames:
+  - system:serviceaccount:{{ monitoring_namespace }}:grafana-serviceaccount
--- a/evals/playbooks/install_middleware_monitoring.yml
+++ b/evals/playbooks/install_middleware_monitoring.yml
@@ -1,0 +1,6 @@
+---
+- hosts: master
+  gather_facts: no
+  roles:
+  - role: middleware_monitoring
+    when: middleware_monitoring | default(true) | bool
--- a/evals/roles/middleware_monitoring/tasks/create_resource_from_template.yml
+++ b/evals/roles/middleware_monitoring/tasks/create_resource_from_template.yml
@@ -1,0 +1,10 @@
+---
+- name: "Create resource file from template ({{ item }})"
+  template:
+    src: "{{ item }}.j2"
+    dest: "{{ monitoring_tmp_dir }}/{{ item }}"
+- name: Create resource from file
+  shell: "oc create -f {{ monitoring_tmp_dir }}/{{ item }} -n {{ monitoring_namespace }}"
+  register: monitoring_resource_create
+  failed_when: monitoring_resource_create.stderr != '' and 'AlreadyExists' not in monitoring_resource_create.stderr
+  changed_when: monitoring_resource_create.rc == 0
--- a/evals/roles/middleware_monitoring/tasks/main.yml
+++ b/evals/roles/middleware_monitoring/tasks/main.yml
@@ -1,0 +1,17 @@
+---
+- name: Check if monitoring namespace exists
+  shell: "oc get project {{ monitoring_namespace }}"
+  register: monitoring_namespace_check
+  failed_when: false
+
+- name: Create monitoring namespace
+  shell: "oc new-project {{ monitoring_namespace }}"
+  when: monitoring_namespace_check.rc != 0
+
+- include: ./create_resource_from_template.yml
+  with_items: "{{ monitoring_resource_templates }}"
+
+- name: Create required operator resources
+  shell: "oc apply -f {{ item }} -n {{ monitoring_namespace }}"
+  register: monitoring_resource_create
+  with_items: "{{ monitoring_resources }}"
--- a/evals/roles/middleware_monitoring/templates/alert_manager_cluster_role.yml.j2
+++ b/evals/roles/middleware_monitoring/templates/alert_manager_cluster_role.yml.j2
@@ -1,0 +1,19 @@
+apiVersion: authorization.openshift.io/v1
+kind: ClusterRole
+metadata:
+  name: alertmanager-application-monitoring
+rules:
+- apiGroups:
+  - authentication.k8s.io
+  attributeRestrictions: null
+  resources:
+  - tokenreviews
+  verbs:
+  - create
+- apiGroups:
+  - authorization.k8s.io
+  attributeRestrictions: null
+  resources:
+  - subjectaccessreviews
+  verbs:
+  - create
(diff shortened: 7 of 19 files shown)


aidenkeating:
@david-martin @pb82 Would you mind taking a look? The operator image seems to be private, is that intentional or can I set it to public?

evals/playbooks/install.yml
```diff
@@ -20,6 +20,7 @@
 - import_playbook: "./install_infra.yml"
 - import_playbook: "./install_services.yml"
 - import_playbook: "./install_webapp.yml"
+- import_playbook: "./install_application_monitoring.yml"
```
david-martin: lets rename the role & playbook to 'middleware monitoring',

which installs an instance of the 'application monitoring operator' configured for middleware services

evals/roles/application_monitoring/defaults/main.yml
```diff
@@ -0,0 +1,11 @@
...
+monitoring_namespace: application-monitoring
+monitoring_repo_raw: "https://raw.githubusercontent.com/integr8ly/application-monitoring-operator/{{ application_monitoring_operator_version | default('master') }}"
+# Resources to create via the oc tool.
+monitoring_resources:
+- "{{ monitoring_repo_raw }}/deploy/service_account.yaml"
+- "{{ monitoring_repo_raw }}/deploy/role.yaml"
+- "{{ monitoring_repo_raw }}/deploy/role_binding.yaml"
+- "{{ monitoring_repo_raw }}/deploy/crds/applicationmonitoring_v1alpha1_applicationmonitoring_crd.yaml"
+- "{{ monitoring_repo_raw }}/deploy/operator.yaml"
+- "{{ monitoring_repo_raw }}/deploy/crds/applicationmonitoring_v1alpha1_applicationmonitoring_cr.yaml"
```
david-martin: We'll have to create crds, cluster roles & rolebindings here too for the dependant operators i.e. grafana and prometheus.
I'll link further details once I have them

evals/roles/application_monitoring/defaults/main.yml
```diff
@@ -0,0 +1,11 @@
+---
+monitoring_namespace: application-monitoring
```
david-martin: lets use `middleware-monitoring` for the namespace

evals/roles/application_monitoring/defaults/main.yml
pb82: @aidenkeating the resources for the Grafana Operator can be found here: https://github.com/integr8ly/grafana-operator/tree/master/deploy

You need to create:

* roles/service_account.yaml
* roles/role.yaml
* roles/role_binding.yaml
* roles/cluster_role.yaml
* roles/cluster_role_binding.yaml

and then the CRDs:

* crds/Grafana.yaml
* crds/GrafanaDashboard.yaml

pb82: @aidenkeating for the Prometheus operator they seem to have all the CRDs under https://github.com/coreos/prometheus-operator/tree/master/contrib/kube-prometheus/manifests

there is a lot of stuff in there but i think you only need to create the custom resource definitions (the top few files).

There is also https://github.com/coreos/prometheus-operator/blob/master/bundle.yaml which creates the service account, role, etc.

pb82: @aidenkeating @david-martin I'm wondering if the installation of those resources (that belong to the dependent operators) should be done in the installer or in the application-monitoring-operator. What do you think?

evals/playbooks/install.yml
matskiv: If I understand it correctly, "evals/playbooks/install_application_monitoring.yml" should be removed?

evals/roles/application_monitoring/defaults/main.yml
david-martin: @pb82 The permissions required to create the crds, cluster roles & rolebindings are the reason for creating them here in the installer.
Anything other than crds, cluster roles/rolebindings for grafana & prometheus should be created by the application-monitoring-operator.

However, prometheus operator contradicts this by having a role for the serviceaccount that allows it to create CRDs.
https://github.com/coreos/prometheus-operator/blob/2d619aa636e65f458a9e1f9fe9a84a6ad3547d25/bundle.yaml#L19-L24.
It would be good to know why this approach was taken.

For the cluster roles/rolebindings, if the application-monitoring-operator were to create these for prometheus and grafana, it would mean the application-monitoring-operator serviceaccount would need permission to create cluster roles/rolebindings.
I'm not sure if the operator should have that kind of permission as it's essentially giving it cluster admin permissions.
Therefore, I think any cluster roles and rolebindings should surface all the way up to the installer and be created there.

However, I'm open to discussion on this and any suggestions.
(cc @maleck13 if you have thoughts on this too).

evals/playbooks/install.yml
aidenkeating: @matskiv I believe the comment was to update the playbook to `middleware monitoring` instead of `application monitoring` to make it more specific to what Integreatly is using the operator for. I've kept the `install_` prefix to match the other playbooks run by `install.yml` (https://github.com/integr8ly/installation/pull/311/files#diff-6de290e3507dbbe68ab781d04bb4995eR20)

@david-martin Is that correct?

aidenkeating: @matskiv Ah I see it's been duplicated now, ignore my previous comment

matskiv:
I ran the full installation with this PR, but `middleware_monitoring` wasn't added to inventory that is saved in a secret in webapp. Is that expected behavior?

matskiv:
I verified first three bullet points, but I am stuck with the last one.
Passing `-e middleware_monitoring=false` doesn't skip the new role. I don't understand how is this possible, any ideas?

aidenkeating:
@matskiv Taking a look at the `middleware_monitoring=false` issue now

evals/roles/middleware_monitoring/defaults/main.yml
```diff
@@ -0,0 +1,28 @@
...
+# Resources to create via the oc tool.
+monitoring_resources:
+- "{{ monitoring_repo_raw }}/deploy/service_account.yaml"
+- "{{ monitoring_repo_raw }}/deploy/role.yaml"
+- "{{ monitoring_repo_raw }}/deploy/role_binding.yaml"
+- "{{ monitoring_repo_raw }}/deploy/crds/applicationmonitoring_v1alpha1_applicationmonitoring_crd.yaml"
+- "{{ monitoring_repo_raw }}/deploy/operator.yaml"
+- "{{ monitoring_repo_raw }}/deploy/crds/applicationmonitoring_v1alpha1_applicationmonitoring_cr.yaml"
+# Grafana resources
+- "{{ monitoring_grafana_repo_raw }}/deploy/roles/cluster_role.yaml"
```
david-martin: @aidenkeating The namespace is hardcoded in some of these resources.

A couple of options:

* do a regex replace with sed before `oc create`
* use a templated version of the files maintained in the application-monitoring-operator repo from https://github.com/integr8ly/application-monitoring-operator/tree/applicationmonitoring-reconciliation/deploy/roles

aidenkeating: @david-martin I'd be a fan of the third option here. I'm assuming https://github.com/integr8ly/application-monitoring-operator/tree/applicationmonitoring-reconciliation/deploy/roles should only contain resources which the operator has permissions to create (so not clusterrolebindings?).

If the installer will create the clusterrolebindings (and other resources that the operator can't) then to me it makes sense to keep the templates close to the installer.

@pb82 Any opinions on this?

pb82: I'm ok with bundling the cluster roles & bindings with the installer. No reason to keep files in the operator that it can't do anything with. In the Future the OLM will hopefully take care of that.

evals/roles/middleware_monitoring/templates/grafana_cluster_role_binding.yml.j2
```diff
@@ -0,0 +1,11 @@
...
+kind: ClusterRoleBinding
+metadata:
+  name: grafana-clusterrole
+roleRef:
+  name: grafana-clusterrole
+subjects:
+  - kind: ServiceAccount
+    name: grafana-operator
+userNames:
+  - system:serviceaccount:application-monitoring:grafana-operator
```
aidenkeating: @david-martin Do these values (`application-monitoring`) also need to be templated, or are they ok being static?

evals/roles/middleware_monitoring/defaults/main.yml
aidenkeating: @david-martin @pb82 I have moved templated versions of the Grafana and Prometheus operator files to the installer. I left the Application Monitoring files referencing GitHub as we can still reference tags for them, if that makes sense?

I've also templated the namespace values in a few of the files, I'm not sure if any other values need to be templated

evals/roles/middleware_monitoring/templates/grafana_cluster_role_binding.yml.j2
david-martin: @aidenkeating this should be set to the namespace `middleware-monitoring`.
I think it should be templated if the `middleware-monitoring` namespace is configurable via a var.

evals/roles/middleware_monitoring/templates/grafana_cluster_role_binding.yml.j2
```diff
@@ -0,0 +1,11 @@
+apiVersion: authorization.openshift.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: grafana-clusterrole
+roleRef:
+  name: grafana-clusterrole
+subjects:
+  - kind: ServiceAccount
+    name: grafana-operator
```
david-martin: should a `namesapce` be specified here too?

evals/roles/middleware_monitoring/defaults/main.yml
david-martin: > I left the Application Monitoring files referencing GitHub as we can still reference tags for them, if that makes sense?

@aidenkeating I think that's reasonable to progress things now. We'll probably need to revisit it when looking to pull in new versions without having to duplicate files here.

aidenkeating:
@pb82 I've added in the alert manager resources. I assume the `{{ .Namespace }}` here is intentional so I've kept it in. https://github.com/integr8ly/application-monitoring-operator/blob/applicationmonitoring-reconciliation/deploy/roles/alertmanager-clusterrole_binding.yaml#L11

pb82:
@aidenkeating actually no. The `{{ .Namespace }}` is not intentional and needs to be replaced with the real namespace.

evals/roles/middleware_monitoring/defaults/main.yml
```diff
@@ -0,0 +1,30 @@
...
+monitoring_tmp_dir: /tmp
+monitoring_repo_raw: "https://raw.githubusercontent.com/integr8ly/application-monitoring-operator/{{ middleware_monitoring_operator_version | default('master') }}"
+# Resources to create via the oc tool.
+monitoring_resources:
+- "{{ monitoring_repo_raw }}/deploy/service_account.yaml"
+- "{{ monitoring_repo_raw }}/deploy/role.yaml"
+- "{{ monitoring_repo_raw }}/deploy/role_binding.yaml"
+- "{{ monitoring_repo_raw }}/deploy/crds/applicationmonitoring_v1alpha1_applicationmonitoring_crd.yaml"
+- "{{ monitoring_repo_raw }}/deploy/crds/applicationmonitoring_v1alpha1_applicationmonitoring_cr.yaml"
+- "{{ monitoring_repo_raw }}/deploy/operator.yaml"
```
aidenkeating: @pb82 Currently we do an `oc create` with this file. I see that it's referencing a `0.0.1` image: https://github.com/integr8ly/application-monitoring-operator/blob/master/deploy/operator.yaml#L18

Should that image be referencing `master` or some other tag? Or should we also move these files into the installer and change the tag there (same as the files below)?

pb82: It might be best to push a 0.0.1 image. I can do that. Another thing: I streamlined the name of the CRDs a bit: https://github.com/integr8ly/application-monitoring-operator/tree/master/deploy/crds

PR #313, opened 2019-01-28
Fix base64 command param in customisation.md

matskiv:
## Additional Information
Change parameter for the long version, which is the same on all platforms.
`-D` works on OSX, but on linux you need `-d`. Long version works on both.

## Verification Steps
Just check `base64 --help` on OSX and Linux


odra:
I think it makes more sense to use ansible's b64decode filter: https://docs.ansible.com/ansible/latest/user_guide/playbooks_filters.html

matskiv:
@odra This isn't a part of playbook though.

odra:
rofl just noticed it's a readme file

PR #316, opened 2019-01-28
ensure default identity provider gets deleted during install

JameelB:
## Additional Information
The default identity provider can have a different configuration in different environments. This causes the following issues below in certain environments that doesn't match our installation/uninstall:
- Default identity provider may not get deleted during installation. 
- Original identity provider configuration is replaced by a set `htpasswd` config which may not match with the original identity provider config. This can cause an issue with using the original default cluster admin user. 

This PR applies the following changes:
- Ensures default identity provider gets deleted during install and saves a copy of the config to the `/tmp` folder. 
- Add back the original config to the `master-config` file by using the copy saved during the install. 

Sample identity provider config in the `master-config.yaml` file:
```
identityProviders:
- name: htpasswd_provider
  challenge: true
  login: true
  provider:
    apiVersion: v1
    kind: HTPasswdPasswordIdentityProvider
    file: /etc/origin/master/htpasswd
masterCA: ca-bundle.crt
```
## Verification Steps

1. Run the install playbook
2. Ensure it removes the default identity provider from the `master-config.yaml` file and replaces it with sso.
3. Ensure that a copy of the original default identity provider config is saved in the `/tmp` folder. 
4. Run the uninstall playbook. 
5. Ensure that the original config is returned to the `master-config` file.


matskiv:
@JameelB if I understand this change correctly, the uninstall will fail on existing instance after we merge this? Simply because original identity provider config was not backed up during installation.
IMHO this is not a big problem, we will just need to communicate this to everyone on the team.

JameelB:
@matskiv Yes that's right, this should be communicated before merging.

sedroche:
Verifying

PR #317, opened 2019-01-29
remove user namespaces after integr8ly namespaces

philbrookes:
@sedroche

### DIFF
--- a/evals/playbooks/uninstall.yml
+++ b/evals/playbooks/uninstall.yml
@@ -1,36 +1,6 @@
 ---
 - hosts: localhost
   tasks:
-    - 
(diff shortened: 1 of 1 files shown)


maleck13:
What is the reason for the change?

philbrookes:
the block of code removing all the namespaces is deleting the rhsso namespace before it's uninstall playbook has run.

The uninstall playbook needs to delete the RHSSO resources in a particular order for it to work correctly.

This block of code was only ever intended to remove users namespaces, not integreatly services namespaces.

sedroche:
@maleck13 This is in relation to https://issues.jboss.org/browse/INTLY-592

tremes:
I am checking that on QE Pony

PR #318, opened 2019-01-29
INTLY-723 add WALKTHROUGH_LOCATIONS to cr.yaml template

tremes:
## Additional Information
See https://issues.jboss.org/browse/INTLY-723. I think this was caused by recent changes in tutorial-web-app-operator

## Verification Steps
Install and make sure the WALKTHROUGH_LOCATIONS env variable is in tutorial-web-app dc

### DIFF
--- a/evals/roles/webapp/defaults/main.yml
+++ b/evals/roles/webapp/defaults/main.yml
@@ -4,6 +4,7 @@ webapp_extra_services_configmap: extra-services
 webapp_client_id: tutorial-web-app
 webapp_route: tutorial-web-app
 webapp_app_label: tutorial-web-app
+webapp_app_walkthrough_locations: https://github.com/integr8ly/tutorial-web-app-walkthroughs.git#v1.1.1
 #operator vars
 webapp_operator_image_url: 'quay.io/integreatly/tutorial-web-app-operator'
 webapp_operator_image_tag: master
--- a/evals/roles/webapp/templates/cr.yaml.j2
+++ b/evals/roles/webapp/templates/cr.yaml.j2
@@ -12,3 +12,4 @@ spec:
       OPENSHIFT_OAUTHCLIENT_ID: "{{ webapp_client_id }}"
       OPENSHIFT_HOST: "{{ openshift_host.stdout }}"
       SSO_ROUTE: "{{ sso_route.stdout }}"
+      WALKTHROUGH_LOCATIONS: "{{ webapp_app_walkthrough_locations }}"


tremes:
@david-martin OK makes sense. So should I close this one? 
I think the web-app-operator is checking the env variables from a wrong input - basically from a deploymentConfig defined in this template, but I will double check

david-martin:
> So should I close this one?

Yes, closing.


> I think the web-app-operator is checking the env variables from a wrong input - basically from a deploymentConfig defined in this template, but I will double check

OK. That doesn't sound right then.

david-martin:
> I believe the correct solution here is to remove that env var from the DC and only have it set in the CR

That goes against discussions with @mikenairn about how best to pull in the webapp operator.
The webapp operator should be treated like any other component i.e. not treated specially because it's under integr8ly.

> This makes rolling out upgrades to the tutorial-web-app a lot easier from the installation repo.
The installation repo shouldn't be in control of the walkthroughs.
A version of the webapp comes with a predefined & tested set of walkthroughs.
Therefore, the installation repo should only care about the webapp (operator) version.

philbrookes:
hmm, interesting, I agree with what you're saying here. I changed the tutorial-web-app-operator to reconcile the DC as a result of changes to the CR, in an attempt to resolve this ticket:
https://issues.jboss.org/browse/INTLY-506 assuming that the installer would now effect this change by updating the CR.

Now I am wondering if the correct solution is to not have the operator reconciling the CR to the DC at all, as it does not create the CR - the installer does.

However, I'm a little unsure where the correct place to store the WALKTHROUGH_LOCATIONS value is then, or what the procedure should be, or how we would allow users to add their own WALKTHROUGH_LOCATIONS values.

If the 1.0.0 operator is running and has deployed a DC with the 1.0.0 tag, then when the 1.1.0 operator is deployed it would need to detect that the DC has an out of date WALKTHROUGH_LOCATIONS value, and change it to be what it now has in it's own DC template file?

If it does this, it runs the risk of removing custom WALKTHROUGH_LOCATIONS that a user may have added.

> A version of the webapp comes with a predefined & tested set of walkthroughs.

So maybe the solution is to not have this value in an env var at all, and instead use a hard-coded value for the integr8ly walkthroughs, and leave the env-var reconciled by the operator and have it only used for extra walkthroughs the user may wish to import?

david-martin:
> in an attempt to resolve this ticket:
> https://issues.jboss.org/browse/INTLY-506

OK, now it makes sense where's you're coming from.
My suggestion contradicts that.

I wonder if the value isn't set in the CR, should it use the default value for the env var?
Would that be a way to satisfy both use cases?

philbrookes:
I like this idea, but how does it handle if the web-app is showing integr8ly walkthroughs, and a user adds a walkthrough location of his own walkthrough repo, then all the integr8ly walkthroughs would vanish?

david-martin:
@philbrookes I think thats fine.
The env var allows for a comma separated list, so they can have both if they want.
However if they only want their own, there's no reason to stop that

PR #320, opened 2019-01-31
INTLY-778 Add uninstall playbook for the monitoring stack

aidenkeating:
Currently we only have an installer for the middleware monitoring
stack. However it would be good to be able to uninstall all of the
resources created by the installer in a simple way.

This change adds an uninstall playbook which removes every resource
created by the installer. It does this by re-creating the same
templates that are created by the installer, but invoking an `oc
delete` on them instead.

Verification:
- Run the install_monitoring_operator.yml playbook
- Run the uninstall_monitoring_operator.yml playbook
- Ensure the middleware-monitoring namespace is removed
- Ensure the other CRDS, ClusterRoles and ClusterRoleBindings
created by the installer are removed
- Try the same with a full uninstall (using uninstall.yml) to
ensure the overall uninstaller works too

### DIFF
--- a/evals/roles/middleware_monitoring/tasks/delete_resource_from_template.yml
+++ b/evals/roles/middleware_monitoring/tasks/delete_resource_from_template.yml
@@ -1,0 +1,10 @@
+---
+- name: "Delete resource file from template ({{ item }})"
+  template:
+    src: "{{ item }}.j2"
+    dest: "{{ monitoring_tmp_dir }}/{{ item }}"
+- name: Delete resource from file
+  shell: "oc delete -f {{ monitoring_tmp_dir }}/{{ item }} -n {{ monitoring_namespace }}"
+  register: monitoring_resource_delete
+  failed_when: monitoring_resource_delete.stderr != '' and 'NotFound' not in monitoring_resource_delete.stderr
+  changed_when: monitoring_resource_delete.rc == 0
--- a/evals/roles/middleware_monitoring/tasks/uninstall.yml
+++ b/evals/roles/middleware_monitoring/tasks/uninstall.yml
@@ -1,0 +1,14 @@
+---
+- include: ./delete_resource_from_template.yml
+  with_items: "{{ monitoring_resource_templates }}"
+
+- name: Delete required operator resources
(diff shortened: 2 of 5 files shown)


aidenkeating:
@david-martin @pb82 Mind taking a look? I'm doing some more testing on this will the full uninstaller now

david-martin:
:eyes:

evals/roles/middleware_monitoring/tasks/delete_resource_from_template.yml
```diff
@@ -0,0 +1,10 @@
+---
+- name: "Create resource file from template ({{ item }})"
+  template:
+    src: "{{ item }}.j2"
+    dest: "{{ monitoring_tmp_dir }}/{{ item }}"
+- name: Delete resource from file
```
david-martin: @aidenkeating Would it be easier and more complete to just delete the namespace?
My thinking is if the template has changed since the install, it may not clean up all resources.

evals/roles/middleware_monitoring/tasks/uninstall.yml
```diff
@@ -0,0 +1,18 @@
+---
+- include: ./delete_resource_from_template.yml
+  with_items: "{{ monitoring_resource_templates }}"
+
+- name: Create required operator resources
```
david-martin: Should say 'Delete'

evals/roles/middleware_monitoring/tasks/delete_resource_from_template.yml
aidenkeating: @david-martin Yeah I was thinking about that. 

The playbook does end up removing the namespace. However, some of the resources created by the install playbook are cluster level, so deleting the namespace doesn't remove them.

By just doing the install in reverse we don't have to keep two lists (stuff to install, stuff to delete) and we know that everything created (CRDs and all) during the install will be removed during uninstall. It's a bit slower though.

I could be missing something though?

david-martin: OK, I'm happy with that explanation.

PR #321, opened 2019-01-31
Intly 592

philbrookes:
## Additional Information
leave the sso namespace to be removed by the sso uninstall task

## Verification
- install integr8ly
- login to webapp with an eval user
- run uninstall
- see no errors during uninstall.


sedroche:
verified on pds @philbrookes

PR #326, opened 2019-02-01
Adds support for 3scale to use S3 bucket (required for OSD)

jessesarn:
## Additional Information
https://issues.jboss.org/browse/INTLY-740

## Verification Steps
Running the install without provided values for the new 3scale s3-related parameters should maintain the same behavior as it did before.

To test S3 storage,
- uncomment the 5 parameters in
https://github.com/jessesarn/installation/blob/c3c67dca8edeb2433bdb735e7f12c8ac3a430eb9/evals/inventories/group_vars/all/common.yml#L18-L22
- threescale_file_upload_storage matches up to the FILE_UPLOAD_STORAGE env variable used by the 3scale components, filesystem (standard RWX pvc-based storage), s3 (storage to s3 bucket)
- create an s3 bucket and set of access keys to read/write to the bucket (or use existing)
- fill in the other 4 parameters with bucket name, and region
- run the install
- verify that the s3 bucket is being populated with objects

### DIFF
--- a/evals/inventories/group_vars/all/common.yml
+++ b/evals/inventories/group_vars/all/common.yml
@@ -13,3 +13,11 @@ eval_launcher_namespace: launcher
 eval_launcher_sso_realm: launcher_realm
 eval_webapp_url_prefix: tutorial-web-app-webapp
 eval_seed_users_count: 50
+
+# To use S3 storage instead of filesystem (PVC-based requiring ReadWriteMany)
+#eval_threescale_file_upload_storage: s3
+#eval_threescale_storage_s3_aws_access_key: <aws_access_key>
+#eval_threescale_storage_s3_aws_secret_key: <aws_secret_key>
+#eval_threescale_storage_s3_aws_region: <aws_region>
+#eval_threescale_storage_s3_aws_bucket: <s3_bucket_name>
+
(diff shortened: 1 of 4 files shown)


evals/inventories/group_vars/all/common.yml
```diff
@@ -12,4 +12,12 @@ eval_apicurito_namespace: apicurito
 eval_launcher_namespace: launcher
 eval_launcher_sso_realm: launcher_realm
 eval_webapp_url_prefix: tutorial-web-app-webapp
-eval_seed_users_count: 50
\ No newline at end of file
+eval_seed_users_count: 50
+
+# To use S3 storage instead of filesystem (PVC-based requiring ReadWriteMany)
```
maleck13: Are these intended as documentation. Seems strange to keep the commented out?

maleck13:
**Verification**
Ensured it still installs ok on RHPDS
Ensure I can login to Webapp and 3scale

@tremes @trepel Anyone able to take this and run it against the QE cluster or similar

tremes:
@maleck13  Yeah I'll try to install it o QE cluster tomorrow.

tremes:
Installation was successful on QE cluster. I was able to login to Webapp as well as to 3scale.

PR #328, opened 2019-02-04
INTLY-805 Add metrics plugin and event listener to Keycloak

aidenkeating:
JIRA: https://issues.jboss.org/browse/INTLY-805

In order to have metrics/monitoring included for the Keycloak
service, we need to add the metrics plugin for Keycloak.

Verification:
- Run the installer, this includes the SSO playbook
- Ensure that the plugin is included in the SSO pod at '/opt/eap/providers',
named 'keycloak-metrics-api.jar'.
- Ensure that the metrics event listener has been configured in
Keycloak.


odra:
:eyes:

PR #329, opened 2019-02-04
allow threescale to enable wildcard routing

JameelB:
## Additional Information
Configure the wildcard routing policy during the installation of 3scale instead of relying on explicit routes created during installation. Changes below also include removal of wt3 routes as this walkthrough doesn't exist.

## Verification Steps
1. Run installation with param `-e eval_threescale_enable_wildcard_route=true`
   - Ensure that the wildcard route is enabled
   - Ensure that no explicit routes are created
   - Ensure that you can go through walkthrough 2
2. Run installation without the param above
   - Ensure that `wt2-*` routes are created
   - Ensure that you can go through walkthrough 2


aidenkeating:
Going to give this a try

PR #330, opened 2019-02-05
upgrade che to 6.17.1

maleck13:
## Additional Information
The PR externalises the CHE templates and resources. It also moves us to 6.17.1 which has improvements around PV creation in multi user mode which is important for OSD

IE it create a PV per workspace and only need ReadWriteOnce

There is an issue in this version when self signed certs are used. You can only create 1 workspace when self signed certs are used. This wont effect RHPDS or our own clusters as they use full ssl certs. So I think we are ok to merge this. Also it is fixed on master so will be ok after the next release. 

Context:
https://github.com/eclipse/che/issues/12024

@jessesarn Could you try this out on OSD also


maleck13:
As che released a 1.0.0 GA I will close this and open a new PR

PR #332, opened 2019-02-06
INTLY-822 Include kube state metrics install/uninstall playbooks

aidenkeating:
This change introduces Kube State Metrics into the middleware
monitoring namespace and ensures that Prometheus can find it by
creating a ServiceMonitor. It includes an install and uninstall
playbook, along with introducing Kube State Metrics into the
Integreatly install.yml playbook.

Verification:
- Run the playbooks/install_kube_state_metrics.yml playbook
- Ensure kube state metrics is in middleware-monitoring namespace
- Ensure Prometheus finds kube state metrics
- Run the playbook/uninstall_kube_state_metrics.yml playbook
- Ensure all resources (ServiceMonitor, deployments etc) are removed
- Run the Integreatly install.yml playbook and ensure it completes

### DIFF
--- a/evals/roles/kube_state_metrics/templates/kube_state_metrics_cluster_role.yml.j2
+++ b/evals/roles/kube_state_metrics/templates/kube_state_metrics_cluster_role.yml.j2
@@ -1,0 +1,46 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: "{{ kube_state_metrics_resource_prefix}}-kube-state-metrics"
+rules:
+- apiGroups: [""]
+  resources:
+  - configmaps
+  - secrets
+  - nodes
+  - pods
+  - services
+  - resourcequotas
+  - replicationcontrollers
+  - limitranges
+  - persistentvolumeclaims
+  - persistentvolumes
+  - namespaces
+  - endpoints
+  verbs: ["list", "watch"]
+- apiGroups: ["extensions"]
+  resources:
+  - daemonsets
+  - deployments
+  - replicasets
+  verbs: ["list", "watch"]
+- apiGroups: ["apps"]
+  resources:
+  - daemonsets
+  - deployments
+  - replicasets
+  - statefulsets
+  verbs: ["list", "watch"]
+- apiGroups: ["batch"]
+  resources:
+  - cronjobs
+  - jobs
+  verbs: ["list", "watch"]
+- apiGroups: ["autoscaling"]
+  resources:
+  - horizontalpodautoscalers
+  verbs: ["list", "watch"]
+- apiGroups: ["policy"]
+  resources:
+  - poddisruptionbudgets
+  verbs: ["list", "watch"]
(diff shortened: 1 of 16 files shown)


evals/roles/kube_state_metrics/templates/kube_state_metrics_cluster_role.yml.j2
```diff
@@ -0,0 +1,46 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: "{{ kube_state_metrics_resource_prefix}}-kube-state-metrics"
```
aidenkeating: This had to be included to not clash with the ClusterRole that may already exist on the cluster, we may want to change our permissions and not change the existing Kube State Metrics on the cluster

david-martin:
@aidenkeating So far so good.

image

I've verified the individual playbook install & uninstall.

I'll do a full integreatly install and give the final approval after reading through the code as well

evals/roles/kube_state_metrics/templates/kube_state_metrics_cluster_role.yml.j2
david-martin: Makes sense.

ns.yml
```diff
@@ -0,0 +1,6 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  annotations:
+    openshift.io/display-name: Yay I am dumb
+  name: moopp
```
david-martin: @aidenkeating eh, should this be here?

david-martin:
@aidenkeating What would be needed to pull in the templates from an external source?
For example, if we wanted to lock down the version of kube-state-metrics we're using, but update it in future, what's needed to make that as simple as a version/tag change in the manifest.json file?

ns.yml
aidenkeating: @david-martin Let's just pretend that never happened. Although, it appears the display name is accurate about me

aidenkeating:
@david-martin There are certain templates for cluster resources that needed their names changed in order to not overwrite the ones that are already on the cluster.

The ClusterRoleBinding and RoleBindings in their have a hardcoded namespace to `kube-sysem` in the `subjects` section, so we'd need to modify that too on our side after pulling.

Some other resources have hardcoded namespaces in the `metadata` section, the same would need to be done for them.

I think it would be possible to do. But we'd still have logic in our playbook that knows about the structure of those templates.

I mean, we could also just hook into the cluster Kube State Metrics if it's there and create it if it's not in the `kube-system` namespace. Then our playbook would just create their templates directly and create a ServiceMonitor in `kube-system`.

philbrookes:
Verified this on my PDS cluster

david-martin:
Lets hold the merge of this.
I'd like to find out if we can rely on kube-state-metrics already being available in the cluster, and hooking up to that instead of installing our own instance

david-martin:
@aidenkeating Merging this.
We may disable/remove this at a later point if and hook up to the kube-state-metrics in the cluster monitoring. For now, let's use it as it gives us what we need

PR #334, opened 2019-02-07
Include latest commit of the spring boot example app with polling

aidenkeating:
We've recently added the ability to poll the backend with the spring
boot example app. This needs the commit/image tag for the app to be
updated to allow for this.

Verification:
- Remove the old spring boot template
- Run the installer
- Ensure a new deploy of the app is polling instead of needing a refresh


sedroche:
Verifying

PR #335, opened 2019-02-07
Intly 775 update che to codeready GA 1.0.0

maleck13:
## Additional Information
In order to do this we needed to create a non system:admin user. So we create a new user and give him clusteradmin perms to run the codeready install

## Verification Steps

#RHPDS 
Run the installer from a fresh cluster with the regular installation command
Go to the codeready ns
Use the url and login with the evals01 or admin@example.com user
Ensure you can create 2 workspaces

#OSD 
As above except you need to pass ```-e 'create_cluster_admin=false' -e 'eval_seed_users_count=1' ``` flag as we don't run as system:admin on OSD


maleck13:
@StevenTobin this could be given a run on OSD to see if there is something fundamental we need to change

odra:
Getting this error: https://gist.github.com/odra/a857484b9184943c6968b7e7f2c9645d

I am running it as the default rhpds admin user instead of system:admin.

maleck13:
@odra ok with this change now. Ok to approve. I am doing the web app now

david-martin:
@pmccarthy Can you test this on a multi-AZ, multi-node (at least 2 nodes in each AZ) cluster.

maleck13:
@odra @david-martin perhaps we should approve and merge back as we know it is working RHPDS and then deal with any fall out on OSD PJS clusters as it appears in smaller changes

david-martin:
> perhaps we should approve and merge back

I don't mind if it's on the test plan for the RC

PR #336, opened 2019-02-07
fix json + removal of eval users

odra:
## Additional Information

Removes unused evals users and fix installation when specifying 0 seed users.

## Verification Steps

* Run an usual installation and test the webapp login
* Run an installation with "0" seed users: `ansible-playbook -i inventories/hosts playbooks/install.yml --e 'eval_seed_users_count=0'` - only the admin user (`admin@example.com`) should be created.

### DIFF
--- a/evals/roles/3scale/tasks/users.yml
+++ b/evals/roles/3scale/tasks/users.yml
@@ -29,12 +29,6 @@
     status_code: [200]
   when: default_admin_user_xml is defined and default_admin_user_xml.matches is defined
 
-- name: Create evaluation user
-  include: _user.yml email={{ threescale_evals_email }} username={{ threescale_evals_username }} password={{ threescale_evals_password }}
-
-- name: Create evaluation admin user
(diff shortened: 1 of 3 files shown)


JameelB:
:eyes:

maleck13:
The users should be removed from 3scale also

evals/roles/3scale/tasks/users.yml
```diff
@@ -29,12 +29,6 @@
     status_code: [200]
   when: default_admin_user_xml is defined and default_admin_user_xml.matches is defined
 
-- name: Create evaluation user
```
odra: @maleck13 I assumed none of those eval users should be created, does it make sense or we need some kind of "eval user creation" flag?

maleck13: admin@example.com is ok. But everything else should be decided by the flag that indicates the number of evals users to create ```eval_seed_users_count```

PR #337, opened 2019-02-07
Add pod count checks per namespace

david-martin:
To verify:

* Run a full install
* Go to the Prometheus Alerts page
* Verify the extra alerts are shown, and are all passing (with the exception of DeadMansSwitch) (see screenshot below)
* Scale down a pod in any of the monitored namespaces
* Verify the corresponding alert is active and yellow. It should turn Red if the pod stays down for 15 minutes
* Scale down *all* pods in any of the monitored namepsace
* Verify the corresponding alert is active
* Delete one of the monitored namespaces
* Verify the corresponding alert is active

image


david-martin:
Rebased, ready for re-review & approval.

PR #341, opened 2019-02-11
INTLY-818 - Generate launcher admin passwords

mikenairn:
Remove launcher_sso_password variable, let template generate password and then set fact from env var.

##Jira

https://issues.jboss.org/browse/INTLY-818

### DIFF
--- a/evals/roles/launcher/defaults/main.yml
+++ b/evals/roles/launcher/defaults/main.yml
@@ -4,7 +4,6 @@ launcher_namespace: "{{ eval_launcher_namespace | default('launcher') }}"
 launcher_sso_serviceclass: sso72-x509-postgresql-persistent
 launcher_sso_serviceinstance_name: launcher-sso-instance
 launcher_sso_username: admin
-launcher_sso_password: ""
 launcher_sso_plan: default
 launcher_sso_realm: launcher_realm
 launcher_sso_prefix: launcher-sso
@@ -12,7 +11,6 @@ launcher_sso_validate_certs: "{{ eval_sso_validate_certs | default('true') }}"
 
 launcher_sso_admin_username: launcheradmin
 launcher_sso_admin_email: launcheradmin@example.com
-launcher_sso_admin_password: ""
 
 launcher_sso_openshift_idp_scope: user:full
 launcher_sso_openshift_idp_client_id: launcher
@@ -23,10 +21,6 @@ launcher_sso_force_image_streams_update: true
 
 launcher_openshift_sso_route: ""
 launcher_openshift_sso_realm: "openshift"
-#What are these for?
-launcher_openshift_sso_username: "admin"
-launcher_openshift_sso_password: "admin"
-#
 
 launcher_sso_keycloak_client_id: launcher-openshift-users
 
--- a/evals/roles/launcher/defaults/main.yml
+++ b/evals/roles/launcher/defaults/main.yml
@@ -4,7 +4,7 @@ launcher_namespace: "{{ eval_launcher_namespace | default('launcher') }}"
 launcher_sso_serviceclass: sso72-x509-postgresql-persistent
 launcher_sso_serviceinstance_name: launcher-sso-instance
 launcher_sso_username: admin
-launcher_sso_password: admin
+launcher_sso_password: ""
 launcher_sso_plan: default
 launcher_sso_realm: launcher_realm
 launcher_sso_prefix: launcher-sso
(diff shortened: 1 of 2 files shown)


evals/roles/launcher/defaults/main.yml
```diff
@@ -23,8 +23,10 @@ launcher_sso_force_image_streams_update: true
 
 launcher_openshift_sso_route: ""
 launcher_openshift_sso_realm: "openshift"
+#What are these for?
 launcher_openshift_sso_username: "admin"
 launcher_openshift_sso_password: "admin"
+#
```
mikenairn: Anyone any idea if these are used? I cant see any reference to them in the installation repo.

evals/roles/launcher/defaults/main.yml
```diff
@@ -4,15 +4,15 @@ launcher_namespace: "{{ eval_launcher_namespace | default('launcher') }}"
...
+launcher_sso_password: ""
 launcher_sso_plan: default
 launcher_sso_realm: launcher_realm
 launcher_sso_prefix: launcher-sso
 launcher_sso_validate_certs: "{{ eval_sso_validate_certs | default('true') }}"
 
 launcher_sso_admin_username: launcheradmin
 launcher_sso_admin_email: launcheradmin@example.com
-launcher_sso_admin_password: admin
+launcher_sso_admin_password: ""
```
mikenairn: @aidenkeating You added this extra admin user here https://github.com/integr8ly/installation/pull/101. What was the reason for the additional admin user? Could you not just use the admin user that was created when the launcher sso was provisioned?

evals/roles/launcher/defaults/main.yml
```diff
@@ -4,15 +4,15 @@ launcher_namespace: "{{ eval_launcher_namespace | default('launcher') }}"
 launcher_sso_serviceclass: sso72-x509-postgresql-persistent
 launcher_sso_serviceinstance_name: launcher-sso-instance
 launcher_sso_username: admin
-launcher_sso_password: admin
+launcher_sso_password: ""
```
mikenairn: Do we want to remove these variables entirely from here? They aren't required herel since they are set as a fact in the provision-sso role.

evals/roles/launcher/tasks/provision-sso.yml
```diff
@@ -40,3 +40,13 @@
...
+
+- name: Retrieve launcher SSO Admin Password
+  shell: "oc get dc/launcher-sso \
+        -o jsonpath='{.spec.template.spec.containers[0].env[?(@.name==\"SSO_ADMIN_PASSWORD\")].value}' \
+        -n {{ launcher_namespace }}"
+  register: launcher_admin_password_cmd
+
+- set_fact:
+    launcher_sso_password: "{{ launcher_admin_password_cmd.stdout }}"
+    launcher_sso_admin_password: "{{ launcher_admin_password_cmd.stdout }}"
```
mikenairn: Not sure what we want to do with this additional admin user, currently i'm just setting the password to be the same as the generated admin user.

evals/roles/launcher/defaults/main.yml
aidenkeating: @mikenairn From what I remember this was added to have a user that could be used to create custom Che stacks. In Che you specify a username as an admin through an env, when that user logs in they can create custom stacks and give read permissions to other users.

I don't think it's possible for the `admin` user to do this, as it's not part of the launcher realm. I haven't looked at this in a while though.

mikenairn: OK. How would you test this user still worked in Che?

aidenkeating: @mikenairn We'd need to try to share a stack through the API, let me see if I can write a script to automate this test.

evals/roles/launcher/defaults/main.yml
odra: Yeah I think they were created in the early days of the installer, let's remove those vars.

evals/roles/launcher/tasks/provision-sso.yml
odra: I guess we just want to make it consistent and use the same password as the one used in the rhsso namespace

evals/roles/launcher/defaults/main.yml
odra: old variables that can be removed

odra:
giving it a try in my cluster.

PR #345, opened 2019-02-11
change default to customer-admin for user limited admin user

maleck13:
## Additional Information
We need to change the name of the user that is the admin with only view access. This PR defaults it to customer-admin

Bug fix for passing in bool on command line for self signed certs
Bug fix for EnMasse version in the manifest
## Verification Steps

Run the installer

There should be no evals-admin any more instead there should be a customer-admin user
You should be able to login with customer-admin to the console and view all projects
You should be able to login to 3scale


aidenkeating:
@maleck13 I have a cluster coming up now, I can verify this once it's complete

PR #348, opened 2019-02-12
INTLY-774 Use AMQ Online 1.0 and remove unused files

aidenkeating:
Update the EnMasse installer to use AMQ Online 1.0 when the
productized flag is set (which it is by default) and remove all
unused patches that were needed for pre-1.0 versions.

Verification:
- Remove any existing EnMasse namespace
- Ensure there are no enmasse-* files or dirs in /tmp
- Run the installer
- Ensure the EnMasse namespace came up ok
- Go through Walkthrough 1A


JameelB:
:eyes:

aidenkeating:
@maleck13 Mind taking a look? Giving it another run now on my cluster

evals/roles/enmasse/tasks/install.yml
```diff
@@ -62,7 +64,7 @@
   when: enmasse_resources_exist.stderr == "No resources found." and  is_productised_version | bool != True
 
 - name: "Provision EnMasse {{ enmasse_version }}"
```
JameelB: Do we still need two of these? Since you're setting the `enmasse_playbook_location` [here](https://github.com/integr8ly/installation/blob/b6bb745399d85177f6d78a66e33a5d499cf87457/evals/roles/enmasse/tasks/install.yml#L48-L54), i think we can just use [this](https://github.com/integr8ly/installation/blob/b6bb745399d85177f6d78a66e33a5d499cf87457/evals/roles/enmasse/tasks/install.yml#L60-L63 ) without any conditions

maleck13:
👁

evals/inventories/group_vars/all/manifest.yaml
```diff
@@ -35,7 +35,7 @@ enmasse_version: 'beta'
 enmasse_download_url: 'https://github.com/EnMasseProject/enmasse/releases/download/{{ enmasse_version }}/enmasse-{{ enmasse_version }}.tgz'
 # Controls whether to use a productised version. Use with 'enmasse_productised_download_url'
```
maleck13: we should set enmasse_version to 1.0.0.GA and prob remove the enmasse_download_url that is currently there

evals/roles/enmasse/tasks/install.yml
maleck13: Yeah I think we will always be installing the productised versions from now on so can remove conditionals

evals/inventories/group_vars/all/manifest.yaml
aidenkeating: Good catch. That download URL isn't used, when `is_productised_version` is `true`, the `enmasse_productised_download_url` URL is.

Do we just want to leave `enmasse_download_url` blank, or completely remove anything related to it?

maleck13: I think either use it for where the prod one is got from or remove it entirely

aidenkeating:
@maleck13 @JameelB Updated the version and removed the unneeded conditionals. I ran the installer and it was successful. Testing 1A now.

pb82:
Tested this together with https://github.com/integr8ly/installation/pull/349

It works, but a webapp change is required (it still tries to provision amq).

PR #349, opened 2019-02-12
remove amq

pb82:
Removes AMQ from the installation.

There's also a PR for removing a WT that used AMQ: https://github.com/integr8ly/tutorial-web-app-walkthroughs/pull/32


sedroche:
Verifying

PR #351, opened 2019-02-13
ensure ksm alert creation does not fail when it already exists

JameelB:
## Additional Information
Installation fails when its ran multiple times due to failure of creating ksm alerts when it already exists. Ensure that the `create ksm alerts` task does not fail if it already exists by adding a `failed_when` condition to the task.

## Verification Steps
1. Run the installer multiple times
2. Ensure each run was successful


pb82:
:eyes:

PR #352, opened 2019-02-13
Bump webapp version to 2.4.0

aidenkeating:


aidenkeating:
@pb82 Mind taking a look?

aidenkeating:
@pb82 I think this value is just metadata at the moment

mikenairn:
> LGTM. ping @mikenairn @maleck13 how is that manifest used? I can't see the installer referencing `webapp_version` anywhere.

@pb82 Not all the manifest variables are currently used. We are in the process as part of the externalisation work to make sure these variables are used though. The _version variables however are really just there for reference i think.

PR #354, opened 2019-02-14
INTLY-747 Include installer and uninstaller for managed fuse

aidenkeating:
Currently a user can provision their own instance of fuse. However
there isn't an already-running fuse instance available to them
when Integreatly starts up.

This adds a fuse which has it's subjectaccessreview configuration
removed so that any user who is authenticated in OpenShift can
access it.

This commit also includes the changes for the uninstallation of it

Verification (Preferably on a fresh cluster):
- Ensure the uninstaller is run if the cluster is not fresh
- Run the full installer
- Ensure a fuse namespace is created
- Login through the exposed route (syndesis-fuse...) as admin
- Login as a different user (an evals user)
- Ensure both logins worked successfully (you could try creating integrations
too or something if you want, to ensure fuse itself works)
- Run the uninstaller
- Ensure the fuse namespace is gone
- Ensure there are no fuse-ignite* imagestreams in the openshift namespace

### DIFF
--- a/evals/roles/fuse_managed/tasks/main.yml
+++ b/evals/roles/fuse_managed/tasks/main.yml
@@ -1,0 +1,47 @@
+---
+
+- name: Create fuse namespace if it doesn't exist
+  shell: oc new-project fuse
+  register: fuse_create_namespace
+  failed_when: fuse_create_namespace.stderr != '' and 'AlreadyExists' not in fuse_create_namespace.stderr
+
+- name: Create Syndesis CRD
+  shell: oc apply -f {{ fuse_online_crd_resources }}
+
+- name: Create Fuse image streams
+  shell: "oc replace --force -f {{ fuse_online_imagestream_resources }} -n openshift"
+  register: fuse_create_imagestream
+  failed_when: fuse_create_imagestream.stderr != '' and 'AlreadyExists' not in fuse_create_imagestream.stderr
+  changed_when: fuse_create_imagestream.rc == 0
+
+- name: Create fuse Operator resources in {{ fuse_namespace }}
+  shell: oc apply -f {{ fuse_online_operator_resources }} -n {{ fuse_namespace }}
+
+- template:
+    src: syndesis-customresource.yml.j2
+    dest: /tmp/syndesis-customresource.yml
+
+- name: Create Syndesis custom resource in {{ fuse_namespace }}
+  shell: oc apply -f /tmp/syndesis-customresource.yml -n {{ fuse_namespace }}
+
+- name: Verify Fuse deployment succeeded
+  shell: sleep 5; oc get pods -n {{ fuse_namespace }} | grep "deploy"
+  register: fuse_verify_result
+  until: not fuse_verify_result.stdout
+  retries: 50
+  delay: 10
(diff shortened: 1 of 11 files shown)


aidenkeating:
@pb82 @maleck13 Mind taking a look?

evals/roles/fuse_managed/tasks/main.yml
```diff
@@ -0,0 +1,47 @@
...
+  shell: sleep 5; oc get pods -n {{ fuse_namespace }} | grep "deploy"
+  register: fuse_verify_result
+  until: not fuse_verify_result.stdout
+  retries: 50
+  delay: 10
+  failed_when: fuse_verify_result.stdout
+  changed_when: False
+
+# Allow any authenticated user to access fuse.
+- name: Remove OpenShift SAR configuration on Fuse Managed OAuth proxy
```
aidenkeating: This is a bit hacky, but the `--openshift-sar` flag is hardcoded in the syndesis/fuse online operator, we only get to choose the namespace. I can make an issue on the fuse repo to perhaps allow the whole flag to be overridden if we want?

pb82:
:eyes:

aidenkeating:
I forgot to include the ServiceMonitor changes, for monitoring this instance, just testing them now.

pb82:
- [x] Ensure a fuse namespace is created
- [x] Login through the exposed route (syndesis-fuse...) as admin
- [x] Login as a different user (an evals user)
- [x] Ensure both logins worked successfully (you could try creating integrations
  too or something if you want, to ensure fuse itself works)
- [x] Run the uninstaller
- [x] Ensure the fuse namespace is gone
- [x] Ensure there are no fuse-ignite* imagestreams in the openshift namespace

pb82:
@aidenkeating looks good, i can retest with the service monitor.

aidenkeating:
So from taking a look at the fuse namespaces provisioned by the managed services broker, it appears we don't currently include a ServiceMonitor, just label the namespace.

@david-martin I think you added the alert/monitoring notes to the Jira ticket, could you confirm whether it's just the namespace labelling that needs to be done, or whether there should be a ServiceMonitor of some type in this namespace?

pb82:
@aidenkeating We only need to service monitor if there is a metrics endpoint that we want to expose. Not sure if fuse has one?

aidenkeating:
@pb82 Ah ok, I think the mention of adding labels in the Jira ticket is for alerting reasons, not monitoring. So it makes sense why there's no ServiceMonitor in the other fuse namespaces. So yeah, I think it's just the label change made in 19903fe that is needed.

david-martin:
> So yeah, I think it's just the label change made in [19903fe](https://github.com/integr8ly/installation/commit/19903fe857c7cfe2867fa5409c9cdc5d3d512845) that is needed.

@aidenkeating Yes. However, we should also disable or change the labeling of fuse namespaces created by the managed-service-broker so that in an OSD cluster we don't get alerts about abritrary fuse instances e.g. from walkthroughs.
We should only be concerned about the single 'special' instance

aidenkeating:
@david-martin I'll make those changes now

PR #355, opened 2019-02-14
INTLY-708 - Externalise all rhsso resources

mikenairn:
* Use manifest variables to determine version of rhsso to install
* Set current version of rhsso operator to latest release (v0.0.3)
* Removed rhsso operator from COMPONENTS, this means it will no longer be tagged during a release and must therefore be updated in the installation repo ahead of time.

Related:

https://github.com/integr8ly/keycloak-operator/pull/30


evals/roles/rhsso/templates/operator-dc.yaml
```diff
@@ -14,7 +14,7 @@ spec:
     spec:
       containers:
         - name: keycloak-operator
-          image: {{ rhsso_keycloak_operator_image_url }}:{{ rhsso_keycloak_operator_image_tag }}
+          image: {{ rhsso_operator_image_url }}:{{ rhsso_operator_release_tag }}
           ports:
           - containerPort: 60000
             name: metrics
```
mikenairn: Ideally we would pull the operator.yaml as an external resource for all the operators as well. However, this would require the operator resource image version to be updated in the operator repo, and kept up to date when new releases happen. This is probably how it should work, but not sure we have anything in place at the moment to make sure this happens.

evals/inventories/group_vars/all/manifest.yaml
```diff
@@ -51,10 +51,10 @@ launcher_template: 'https://raw.githubusercontent.com/fabric8-launcher/launcher-
 
 #not currently used but is the version installed by the operator
 # rhsso is not an optional component
-rh_sso_version: '7.2.2.GA'
+rhsso_version: '7.2.2.GA'
 #below vars not currently used but will be used to pull in the new versions of RH-SSO once we decouple the resources from the installer.
-rh_sso_opertor_release_tag: '0.0.2'
-rh_sso_operator_resources: 'https://raw.githubusercontent.com/integr8ly/keycloak-operator/{{rh_sso_opertor_release_tag}}/deploy/'
+rhsso_operator_release_tag: 'ee9f89e'
```
mikenairn: Do we want to create a new release of the rhsso operator and set a proper release tag here?

matskiv:
> Removed rhsso operator from COMPONENTS, this means it will no longer be tagged during a release and must therefore be updated in the installation repo ahead of time.
@mikenairn Why are we not going to tag it during a release?

maleck13:
👀

evals/inventories/group_vars/all/manifest.yaml
maleck13: Yes I think that would be worth while. Only takes a minute or two to create a new release in GH

evals/roles/rhsso/templates/operator-dc.yaml
maleck13: Could we create a follow up issues for this to make sure we move in that direction. I think ideally we would move with the operator releases and capture the version of the product installed from the product cr

evals/inventories/group_vars/all/manifest.yaml
mikenairn: I just realised we have an issue with getting it to build an appropriately tagged image and having it pushed to quay.io. We unfortunately can't rely on the docker build in quay.io since we need to build the operator binary as well. Currently we rely on the travis build to do this for us when we merge back to master or make any changes, but that won't get triggered when we create a release in GH. We will have this issue for all our operators, we can obviously do it manually, but that will no doubt result in errors.

mikenairn:
> > Removed rhsso operator from COMPONENTS, this means it will no longer be tagged during a release and must therefore be updated in the installation repo ahead of time.
> 
> @mikenairn Why are we not going to tag it during a release?

@matskiv The idea is that everything, including our own parts, will have the latest versions set in the manifest. If there is a new version of RHSSO it needs to be updated in this repo outside of any integeratly release creation like we would do for a new version of fuse, or che, or whatever. A release of integreatly should eventually just be a new tag on this repo and nothing else.

evals/inventories/group_vars/all/manifest.yaml
maleck13: There is an option to change the dockerfile to use one with go installed and then run a go build during the docker container creation

evals/roles/rhsso/templates/operator-dc.yaml
mikenairn: operator.yaml template has now been removed and it's pulled from the external resources URL like everything else. Image version is now set in the operator.yaml in the operators repo see https://github.com/integr8ly/keycloak-operator/pull/30

evals/inventories/group_vars/all/manifest.yaml
mikenairn: I was mistaken in thinking tags wouldn't cause a build to be triggered, they do. We can continue to use travis to create the images as when we create a new tag/release on GH it will automatically build the image for use as expected.

evals/inventories/group_vars/all/manifest.yaml
```diff
@@ -53,10 +53,10 @@ launcher_template: 'https://raw.githubusercontent.com/fabric8-launcher/launcher-
 
 #not currently used but is the version installed by the operator
 # rhsso is not an optional component
-rh_sso_version: '7.2.2.GA'
+rhsso_version: '7.2.2.GA'
 #below vars not currently used but will be used to pull in the new versions of RH-SSO once we decouple the resources from the installer.
-rh_sso_opertor_release_tag: '0.0.2'
-rh_sso_operator_resources: 'https://raw.githubusercontent.com/integr8ly/keycloak-operator/{{rh_sso_opertor_release_tag}}/deploy/'
+rhsso_operator_release_tag: 'v0.0.3'
+rhsso_operator_resources: 'https://raw.githubusercontent.com/integr8ly/keycloak-operator/{{rhsso_operator_release_tag}}/deploy/'
```
mikenairn: @maleck13 Created new v0.0.3 release see https://github.com/integr8ly/keycloak-operator/pull/30

maleck13:
I dont have a cluster to test on currently as I am helping DIL with their stuff

evals/roles/rhsso/tasks/main.yml
```diff
@@ -29,32 +29,11 @@
...
-- name: "Create operator deployment config template"
-  template:
-    src: "operator-dc.yaml"
-    dest: /tmp/operator-dc.yaml
-
-- name: "create operator deployment config"
-  shell: "oc create -f /tmp/operator-dc.yaml -n {{ rhsso_namespace }}"
-  register: rhsso_dc
-  failed_when: rhsso_dc.stderr != '' and 'AlreadyExists' not in rhsso_dc.stderr
+  shell: "oc create -f {{ item }} -n {{ rhsso_namespace }}"
```
mikenairn: @maleck13 Having gone through some of the other operators we seem to use a mix of "oc create" and "oc apply" when creating these resources. What is the preferred way to do it?

maleck13: Haven't really given that one a lot of thought. apply will also update an existing resource. So apply may be useful from a dev point of view as you can potentially keep running the same command and making differences to the resource without having to delete them all first

mikenairn:
Closed in favour of https://github.com/integr8ly/installation/pull/371

PR #356, opened 2019-02-14
Hide identity providers rather than deleting them

philbrookes:
## Additional Information
Instead of deleting the htpasswd identity provider we are just going to hide it from the login screen instead.

This way the users in that provider can be used for the install/uninstall without issue.

Related update to CI/CD: https://github.com/integr8ly/ci-cd/pull/47

## Verification
- Start a *fresh* cluster from RHPDS (it's unlikely this will work on a cluster that has old installations run on it - without some manual shaking of the cluster)
- Run the install
- Check `/etc/origin/master/master-config.yaml` has `login: false` for the htpasswd provider, and `login: true` for the sso provider.
- login on `oc` cli with opentlc-mgr account.
- at web login confirm no htpasswd login option
- login to ui with admin@example.com account
- run uninstall with `-e openshift_login=true -e openshift_username=<opentlc-mgr> -e openshift_password=<opentlc-mgr password> -e openshift_master_public_url=<your cluster url>`
*N.B.* This is not normally required, but is required to validate these changes.
- Check `/etc/origin/master/master-config.yaml` that the htpasswd has `login: true`
- Login on CLI and web with opentlc-mgr account.


maleck13:
@philbrookes what is the best way to verify this

philbrookes:
@maleck13 just added verification steps :)

philbrookes:
there's a bug in this atm, looking into it.

evals/roles/rhsso/tasks/remove_identityprovider.yml
```diff
@@ -10,10 +10,11 @@
   when: output.stdout != "0"
   block:
     -
-      name: Remove RH-SSO identity provider from master config
-      shell: "sed -i.bak '/# BEGIN GENERATED BY SSO INSTALLER/,/# END GENERATED BY SSO INSTALLER/d' {{ rhsso_openshift_master_config_path }}"
+      name: Remove identity providers from master config
+      shell: "sed -i.bak '/identityProviders/,/masterCA/{//!d}' {{ rhsso_openshift_master_config_path }}"
```
matskiv: This will delete all identity providers and later try to recreate them from `original_identity_provider_path` which is:
https://github.com/integr8ly/installation/blob/c0d36d76bca1d44829e6ad8c712eca12456843e6/evals/roles/rhsso/defaults/main.yml#L57
And I just realized that after 10 days that backup will be gone from /tmp


So while we are at it, wouldn't it be better to remove what's in here:
`'/# BEGIN GENERATED BY SSO INSTALLER/,/# END GENERATED BY SSO INSTALLER/d'`
and then do the opposite of what we are doing at the installation time - set `login: true`

philbrookes: yeah, I have already mentioned to @JameelB that we should move that file, I believe she is working on that.

I considered toggling the `false` back to `true` but that could turn on a provider that the user had originally set to hidden.

matskiv: Oh, that's actually good point. We would have also turned on identity provider that's just for SRE team(I think Steven mentioned it on standup today). I forgot about that.
We should be fine if we move it.

philbrookes: yeah, jameel is working on that in a seperate PR, so I don't think that needs to block this PR

philbrookes:
Install pipeline: https://integreatly-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/installation-pipeline-qe-pony/29/console

Uninstall pipeline: https://integreatly-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/uninstallation-pipeline-qe-pony/44/console

PR #357, opened 2019-02-14
Removing inventory_file var from enmasse install

pmccarthy:
## Additional Information
At the moment the EnMasse role uses the `inventory_file` to append a custom inventory to the end. This will not work when using dynamic inventories which Ansible tower uses.

The change here is to generate a separate inventory hosts file for EnMasse and pass that into the installer

## Verification Steps
- Run a standard integreatly install


philbrookes:
:eyes:

PR #359, opened 2019-02-14
INTLY-774 Revert MONITORING_KEY changes in the MSB

aidenkeating:
Verification:
- Run the installer
- Login to the webapp
- Check the labels on the provisioned fuse service, ensure it now is `brokered-fuse-online` instead of `middleware`
@sedroche This is the corresponding change to https://github.com/integr8ly/managed-service-broker/pull/35

### DIFF
--- a/evals/roles/msbroker/defaults/main.yml
+++ b/evals/roles/msbroker/defaults/main.yml
@@ -8,7 +8,7 @@ msbroker_clusterrole: managed-service
 msbroker_clusterrolebinding: default-cluster-account-managed-service
 msbroker_deployment_name: msb
 msbroker_servicebroker_name: managed-service-broker
-monitoring_key: middleware
+monitoring_key: brokered-fuse-online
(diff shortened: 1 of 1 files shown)


evals/roles/msbroker/defaults/main.yml
```diff
@@ -8,7 +8,6 @@ msbroker_clusterrole: managed-service
 msbroker_clusterrolebinding: default-cluster-account-managed-service
 msbroker_deployment_name: msb
 msbroker_servicebroker_name: managed-service-broker
-monitoring_key: middleware
```
david-martin: @aidenkeating rather than removing this, what about setting it to a different value so those namespaces could potentially be monitoring as a group in future?

e.g. 'brokered-fuse-online' ?

aidenkeating: @david-martin That sounds good to me, it'll save reverting the MSB changes too, just updating the Makefile in there, which is a plus

PR #361, opened 2019-02-15
Refactor fuse namespace check

aidenkeating:
Make the namespace readiness check in the installer not fail if pods take a long time to be created by the fuse operator.

We were just waiting 5 seconds before, these checks are more deterministic.


aidenkeating:
@pb82 Mind taking a look?

pb82:
:eyes:

PR #363, opened 2019-02-15
INTLY-986: Put back che uninstall (with minor improvements).

matskiv:
## Additional Information
JIRA: https://issues.jboss.org/browse/INTLY-986

## Verification Steps
Run the uninstall playbook.
Verify that codeready namespace has been removed.
Run the uninstall playbook again and check exist code(should be 0).


evals/playbooks/install_services.yml
```diff
@@ -87,17 +87,6 @@
...
-      name: Uninstall che
-      include_role:
-        name: che
-        tasks_from: uninstall
-      vars:
-        che_keycloak_user: "{{ eval_launcher_sso_admin_username }}"
-        che_keycloak_password: "{{ eval_launcher_sso_admin_password }}"
-        che_keycloak_host: "{{ eval_launcher_sso_host }}"
-      tags: ['che']
-      when: eval_action == 'uninstall' and che and launcher
```
matskiv: I don't know why this was in installation playbook, so I removed it because it didn't make sense.

matskiv:
Implementation of tasks/uninstall.yml is almost the same as before (before it was removed in #335).

evals/roles/code-ready/tasks/uninstall.yml
```diff
@@ -0,0 +1,71 @@
+---
+-
+  name: Set initial vars
+  set_fact:
+    che_keycloak_url: "{{ che_protocol }}://{{ che_keycloak_host }}"
+
+-
+  name: Delete objects labeled as app:{{ che_app_label }}
+  shell: "oc delete all -l app={{ item }} -n {{ che_namespace }}"
```
sedroche: Do we need to delete all the resources under `che_namespace`? Testing manually just deleting the namespace seems to work

matskiv: TBH I just reused what we had previously. 
I am guessing that this was done due to pesky finalizers.
I think that a lot of times namespace wasn't actually deleted because of some leftovers.

sedroche:
@matskiv Im finishing up a verification, will verify this then

evals/playbooks/install_services.yml
odra: We used to have a "all in one" playbook that would install or uninstall based on tags but that playbook is gone now so it makes sense to remove it.

evals/playbooks/uninstall.yml
```diff
@@ -124,6 +124,17 @@
...
       when: fuse_managed | default(true) | bool
+    -
+      name: Uninstall che
+      include_role:
+        name: code-ready
+        tasks_from: uninstall
+      vars:
+        che_keycloak_user: "{{ eval_launcher_sso_admin_username }}"
+        che_keycloak_password: "{{ eval_launcher_sso_admin_password }}"
+        che_keycloak_host: "{{ eval_launcher_sso_host }}"
```
sedroche: I'm seeing an error:

```bash
fatal: [127.0.0.1]: FAILED! => {"msg": "The task includes an option with an undefined variable. The error was: 'eval_launcher_sso_host' is undefined\n\nThe error appears to have been in '/root/installation/evals/roles/code-ready/tasks/uninstall.yml': line 3, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n-\n  name: Set initial vars\n  ^ here\n"}
```

This is the only mention of it I see

philbrookes:
:eyes:

odra:
giving a try as well

evals/playbooks/uninstall.yml
odra: got the same errors, this piece of code needs to be applied in the uninstall playbook: https://github.com/integr8ly/installation/blob/master/evals/playbooks/install_services.yml#L34

philbrookes: I'm seeing the same error here, it looks like it is set during install like so

https://github.com/integr8ly/installation/blob/master/evals/playbooks/install_services.yml#L60-L70

```
./playbooks/install_services.yml:      register: eval_launcher_sso_host_cmd
./playbooks/install_services.yml:        eval_launcher_sso_host: "{{ eval_launcher_sso_host_cmd.stdout }}"
```
But it is not being set during the uninstall.

matskiv: Good catch guys. I don't know why I assumed those vars were defined.
I'll fix this now.

matskiv:
Just FYI, to avoid duplicate work, Phil is helping me with verification.

evals/roles/code-ready/tasks/uninstall.yml
matskiv: In the end, based on some info about uninstall problems from Phil, I decided to simplify it and leave just namespace removal. :)

odra:
giving it  a try again

PR #367, opened 2019-02-18
ignore error when CRD does not exist

philbrookes:
## Additional Information
This code throws an error if the uninstall is run when there is no CRD to delete.


odra:
:eyes:

trepel:
@philbrookes @odra this blocks our nightly pipelines, is there anything else to do to help you with so this can be merged? Also, will this make it into 1.2.0 release?

odra:
@trepel should be on 1.2 rc yes.

trepel:
@odra what is the process here, will you just cherry-pick this to the v1.2 branch or there is more to do than that?

PR #375, opened 2019-02-19
remove finalizers before deleting NS

philbrookes:
If the uninstaller went to the deletion of the managed-service-broker too quickly, the fuse deletion does not complete. This PR makes the uninstaller wait for all user namespaces to be deleted before moving on.

https://integreatly-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/uninstallation-pipeline-qe-pony/56/console

### DIFF
--- a/evals/playbooks/uninstall.yml
+++ b/evals/playbooks/uninstall.yml
@@ -4,17 +4,11 @@
 
 - hosts: localhost
   tasks:
-    - when: keep_namespaces is undefined or not keep_namespaces | bool
-      block:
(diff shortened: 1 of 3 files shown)


tremes:
Looking & verifying

sedroche:
@philbrookes Should we move this into it's own role? Add some correct naming etc?

philbrookes:
@sedroche moved to it's own role now, if you wouldn't mind verifying this again for me

sedroche:
Looks good @philbrookes. Verified.

PR #382, opened 2019-02-20
Move master-config copy from the /tmp folder to /master

JameelB:
## Additional Information
The original identity provider config for a cluster is currently saved in the /tmp folder during [installation](https://github.com/integr8ly/installation/blob/master/evals/roles/rhsso/tasks/identityprovider.yml#L34). This needs to be moved as the contents of the `/tmp` directory may get removed over time.

The following changes have been done for the installation:
- Save a backup copy of the original identity provider configuration from the oldest master-config backed up by ansible (`master-config.yaml.<timestamp>~`). This step is only executed if the backup copy doesn't exist or has no content
- The original identity provider config backup has now been moved to `/etc/origin/master/idp-config.integreatly-original.yaml`.

The following changes have been done for the uninstallation:
- Restoration of the identity provider config using the file `/etc/origin/master/idp-config.integreatly-original.yaml`. This will only be attempted by the uninstaller if this file exists and has content.

This PR should also fix issues in regards to subsequent attempts of running the installer and uninstaller.

## Verification Steps
1. Run the uninstall playbook 
2. Ensure this runs successfully and the identity provider content was not changed.
3. Run the installer playbook
4. Ensure that the installation completes successfully and a copy of the original identity provider can be seen in the `/etc/origin/master` directory with the correct content.
5. Run the installer playbook again.
6. Ensure that the content of the identity provider copy has not been changed.
7. Run the uninstall playbook
8. Ensure the uninstall playbook runs successfully
9. Ensure that the identity provider config was restored to it's original content.


JameelB:
@trepel @odra @matskiv @philbrookes

matskiv:
👀 
I have a clean PDS cluster, so I'll do the verification as well.

evals/playbooks/uninstall.yml
```diff
@@ -128,12 +128,6 @@
         name: reboot_template_broker
 - hosts: master
   tasks:
-    -
-      name: Remove webapp cors data
-      include_role:
-        name: webapp
-        tasks_from: cors-remove
-      tags: ['webapp']
```
matskiv: Let's remove `roles/webapp/tasks/cors-remove.yml` as well then?

JameelB:
@matskiv Will need to rebase this to get this [commit](https://github.com/integr8ly/installation/commit/0eb2f9ea4d911b20de5a9951df3c28e51ed0dfa8) in for the uninstall.

evals/playbooks/uninstall.yml
JameelB: Good catch :+1:

evals/roles/rhsso/tasks/identityprovider.yml
```diff
@@ -31,16 +31,28 @@
...
+  become: yes
+
+- name: Get original master config backup filename
+  shell: "ls /etc/origin/master/ -t | grep master-config.yaml | tail -n 1"
+  register: original_master_config_backup_filename
+  become: yes
+
+# Only need to save this once when the original master config copy does not exist
+- name: Save original master config
+  shell: "cat /etc/origin/master/{{ original_master_config_backup_filename.stdout }} > {{ original_master_config_path }}"
```
matskiv: @JameelB Why aren't you using `rhsso_openshift_master_config_path` for the master-config.yaml path, as in remove_identityprovider.yml ?

JameelB: Ansible backs up the `master-config.yaml` when it changes in the `/master` directory with the format `master-config.yaml.<timestamp>`.  Instead of using the `master-config.yaml` file which changes quite a bit, we can use this file because this will always have the original content.

matskiv: Okay, thanks for the explanation 👍

JameelB:
@pmccarthy @StevenTobin @jessesarn Do you see any issues that may come up with this solution for OSD and Pony? Thanks for the verification @matskiv !

jessesarn:
> @pmccarthy @StevenTobin @jessesarn Do you see any issues that may come up with this solution for OSD and Pony? Thanks for the verification @matskiv !

Since Pony clusters are specifically for Integreatly, I think the logic should be fine; however, for OSD we have to worry about other changes made to the confirm outside of Integreatly.  The more time that has elapsed between install and uninstall, the higher the chance that other changes which we don't account for could have occurred (master nodes swapped out, keys/secrets rotated, openshift version updated, certificate changes, etc.)  Obviously the last thing we want to break is a paying customer's OSD cluster.  Also, the OpenShift SRE team will be running the Integreatly playbooks instead of us with knowledge about what things might get lost/broken.

The best ways I can think of to safely deal with this are:
1 - Using comment blocks when we need to disable instead of actually removing a section, then during uninstall, we can uncomment the sections which were block commented with our header/footer.  Any changes made to other sections of the file since the install will remain in the current working state.
2. Maybe use the date created of our backup file to compare against the other master-config backups and if we find any backups newer than ours, assume that we can't safely use our backup to restore to.  (We'll need to talk more about how to notify the user, allow them to proceed, etc.)

I'm not ecstatic about the second one since there are plenty of holes that could mess it up, but it'd at least reduce the chance of breaking a working cluster.  Maybe somebody else has some better ideas?

jessesarn:
I just thought of a 3rd option which would probably fall in between 1 and 2 which is:

3. Instead of overwriting the entire master config with the full contents of our backup file, only restore only the specific sections that we change (oauth, cors).  This wilil allow all other changes that could have been made to the rest of the file to be untouched.

JameelB:
@jessesarn I would go for option 3 with the same way we were doing it before so that we're only changing the bits we change. Do you think there will be a chance that the original contents of the identityProvider will be changed outside of our installation? I think it will be difficult to capture those changes in the backup copy.

philbrookes:
maybe we should just change the bit of code that hides all default identity providers to just look for the htpasswd one, and hide that, then in the uninstall it can remove the rh_sso block, and specifically unhide the htpasswd provider again

JameelB:
Will there only be a `htpasswd` by default within the original identityProvider config? I think @StevenTobin mentioned that a different default idp could also be there by default on OSD?

jessesarn:
There will always be at least 2 IDP's included with an OSD cluster, a hidden one which gives access to openshift-sre and the other which is the customer chosen one which can never be htpasswd since a customer wouldn't be able to log into the systems to manage users.

odra:
:eyes:

philbrookes:
:eyes:

philbrookes:
Can not verify at the moment as it fails with the error:
```
TASK [code-ready : patch for per workspace volumes] *************************************************************************************************************************************************************************************************************************************************************************
fatal: [127.0.0.1]: FAILED! => {"changed": true, "cmd": "oc set env deployment/che CHE_INFRA_KUBERNETES_PVC_PRECREATE__SUBPATHS=false -n codeready", "delta": "0:00:00.201577", "end": "2019-02-27 09:23:01.580651", "failed_when_result": true, "msg": "non-zero return code", "rc": 1, "start": "2019-02-27 09:23:01.379074", "stderr": "Error from server (NotFound): deployments.extensions \"che\" not found", "stderr_lines": ["Error from server (NotFound): deployments.extensions \"che\" not found"], "stdout": "", "stdout_lines": []}
```
It looks like the PR to master for that issue is merged now: https://github.com/integr8ly/installation/pull/401

So just rebase this branch and ping me, and I will try again.

JameelB:
@philbrookes rebased it there, mind taking a look again when you have time?

philbrookes:
:eyes:

PR #385, opened 2019-02-20
INTLY-934 Add OpenShift monitoring federation

aidenkeating:
Federate with the Prometheus in the OpenShift Monitoring from the
middleware monitoring namespace instead of using our own instance
of Kube State Metrics.

Verification:
- Run the installer and ensure it completes
- Ensure Kube State Metrics is not in middleware-metrics
- Ensure Alerts are passing (they may be pending for a while)
- Run the uninstaller and ensure it completes


aidenkeating:
@sedroche I'm going to update this PR to follow your role structure, as our PRs should probably have their logic end up in the same `config` role.

sedroche:
:eyes:

evals/roles/middleware_monitoring_config/jobs/openshift_monitoring_federation.j2
```diff
@@ -0,0 +1,23 @@
+- job_name: openshift-monitoring-federation
+  honor_labels: true
+  kubernetes_sd_configs:
+  - role: service
+    namespaces:
+      names:
+      - openshift-monitoring
```
sedroche: I think it makes sense to template this value as well

evals/roles/middleware_monitoring_config/tasks/main.yml
```diff
@@ -0,0 +1,13 @@
+---
+- name: "copy ksm template"
```
sedroche: https://github.com/integr8ly/installation/blob/ef57460fd9eea52a0c29936e32f022eba7679b0b/evals/roles/middleware_monitoring_config/tasks/main.yml#L2-L11

Probably worth separating this into a `create_ksm_alerts` task file and including it

sedroche:
@aidenkeating Looks good just a couple of comments.

sedroche:
Verifying

aidenkeating:
Closing for #386

PR #386, opened 2019-02-21
Middleware monitoring config

aidenkeating:
Rebase of #385 on #384 

@sedroche 

## Verification Steps for 3scale

1. Run this installation branch.
2. Go to the Prometheus UI. (`prometheus-route`/config  in the `middleware-monitoring` namespace) and make sure a new config is present.

config

3. Ensure there are 2 new targets at `prometheus-route`/targets
targets
4. Ensure you can run a query v the new targets. e.g. `nginx_http_connections{_3scale_component_element="staging"}` at `prometheus-route`/graph
graph

## Verification Steps for federation

1. Run the installer and ensure it completes (should already be done in 3scale)
2. Ensure Kube State Metrics is not in the `middleware-monitoring` namespace
3. Ensure an `openshift-monitoring-federation` target exists in prometheus and is up
screen shot 2019-02-21 at 10 27 17
4. Ensure Alerts are passing (they may be pending for a while)
screen shot 2019-02-08 at 21 15 24
5. Run the uninstaller and ensure it completes

PR #389, opened 2019-02-21
Include node-exporter metrics in monitoring federation

aidenkeating:
We currently have federation between middleware-monitoring and
the Prometheus instance in openshift-monitoring. However, we only
are federating information about the Kube State Metrics metrics.

We should include node-exporter metrics in this too.

Verification:
- Run the installer
- Login to the middleware-monitoring Prometheus instance
- Ensure that the `openshift-monitoring-federation` target now
has two sets of match[] params on the targets page, one should
be `{ service="node-exporter" }`.
- Try to do a query like `{ service="node-exporter" }` and ensure
you are seeing a response that has data (a lot of entries), this
can take some time.


aidenkeating:
@sedroche Mind taking a look, this was meant to be in #386

sedroche:
Verifying

PR #393, opened 2019-02-22
recreate codeowners file

philbrookes:
This has to stay, otherwise anyone from the org can approve and merge any PR and we need to keep some control over what is going into this repo.

It turns out that the emails sent automatically to codeowners have a signature that can be filtered, so we will have to rely on this solution for now.


philbrookes:
so when the codeowner file triggers a review request the body of the request email matches a regex like:
```
^.*as a code owner\.$
```
So we can filter on that for now, this doesn't help with the comment / merge / close spam but is at least progress

PR #400, opened 2019-02-25
INTLY-1079 Move gitea pod count alert rule to it's own PrometheusRule in the gitea namespace

david-martin:
This change creates the GiteaPodCount alert only if Gitea is installed (true by default).
If the var is set to `gitea=false`, both Gitea and the GiteaPodCount alert will *not* be installed.

The GiteaPodCount alert is moved from the `middleware_monitoring_config` role to it's own new role `gitea_config`

*Verification*

* Run an install with `-e 'gitea=false'` & verify the GiteaPodCount alert *is not* present in Prometheus (the gitea namespace should not exist either)
* Run an install with `-e 'gitea=true'` & verify the GiteaPodCount alert *is* present in Prometheus and the `prometheusrule` exists in the `gitea` namespace 
  * `oc get prometheusrule/gitea-prometheusrule -n gitea`

### DIFF
--- a/evals/playbooks/install_middleware_monitoring.yml
+++ b/evals/playbooks/install_middleware_monitoring.yml
@@ -8,3 +8,8 @@
     when: middleware_monitoring | default(true) | bool
   - role: 3scale_config
     when: middleware_monitoring | default(true) | bool
+  tasks:
+  - include_role:
+      name: gitea
+      tasks_from: monitoring
+    when: (middleware_monitoring | default(true) | bool) and gitea
--- a/evals/roles/gitea/tasks/monitoring.yaml
+++ b/evals/roles/gitea/tasks/monitoring.yaml
@@ -1,0 +1,27 @@
+---
+- name: Check {{ gitea_namespace }} namespace exists
+  shell: oc get namespace {{ gitea_namespace }}
+  register: gitea_namespace_check
+  failed_when: gitea_namespace_check.stderr != '' and 'not found' not in gitea_namespace_check.stderr
+
+- block:
+    - name: Check if the gitea PrometheusRule {{ gitea_prometheusrule_name }} exists in {{ gitea_namespace }} namespace
+      shell: "oc get prometheusrule {{ gitea_prometheusrule_name }} -n {{ gitea_namespace }}"
+      register: gitea_prometheusrule_exists
+      failed_when: gitea_prometheusrule_exists.stderr != '' and 'not found' not in gitea_prometheusrule_exists.stderr
+
+    - block:
+      - name: Create the gitea PrometheusRule custom resource file
+        template:
+          src: "{{ gitea_prometheusrule_name }}.yaml.j2"
+          dest: "/tmp/{{ gitea_prometheusrule_name }}.yaml"
+
+      - name: Create the gitea PrometheusRule custom resource
+        shell: "oc create -f /tmp/{{ gitea_prometheusrule_name }}.yaml -n {{ gitea_namespace }}"
+
+      - name: Delete the gitea PrometheusRule custom resource file
+        file:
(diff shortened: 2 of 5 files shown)


david-martin:
@maleck13 I'd be interested in your thoughts on 2 things in this PR:

* the pattern of a `<service>_config` role for anything configured or set up after the install of `<service>` i.e. splitting the role into install and config roles
* grouping the checks for triggering these `<service>_config` roles centrally. In this case I could see 1 per service that has alerts being added to `evals/playbooks/install_middleware_monitoring.yml`. However, this playbook may not be the best place for that as there may be other non monitoring related config/setup for each service too

evals/playbooks/install_middleware_monitoring.yml
```diff
@@ -8,3 +8,5 @@
     when: middleware_monitoring | default(true) | bool
   - role: 3scale_config
```
maleck13: so a pattern I am looking at for the backups is to have a structure like ```playbooks/backup_restore/install.yaml  playbooks/backup_restore/upgrade.yaml```

evals/roles/gitea_config/tasks/main.yaml
```diff
@@ -0,0 +1,3 @@
+---
+- name: Create PrometheusRules
```
maleck13: wondering should config be a separate dir under gitea/tasks instead of its own role

david-martin: Yes, I was thinking about this and that makes more sense to me.
I'm shying away from the original idea of having an `_config` or similar role for each product.
Tasks seems like the right ansible solution here.

In this case `config` probably isn't the right name for the task.
I'd prefer `monitoring`.

evals/playbooks/install_middleware_monitoring.yml
david-martin: Moved to a task in `gitea/tasks/monitoring.yaml`,
which is included here only if `middleware_monitoring` & `gitea` are installed

odra:
gonna give it  a try

evals/roles/gitea/tasks/monitoring.yaml
```diff
@@ -0,0 +1,27 @@
+---
+- name: Check {{ gitea_namespace }} namespace exists
+  shell: oc get namespace {{ gitea_namespace }}
+  register: gitea_namespace_check
+  failed_when: false
```
odra: Can you add an specific 'not found error" such as `gitea_namespace_check.stderr != '' and 'not found' not in gitea_namespace_check.stderr` - it could fail in case of an unexpected issue shows up  such as master is unreachable which we may want to throw an error

odra:
got this error:

```
TASK [gitea : Create the gitea PrometheusRule custom resource file] **********************************************************************************
fatal: [127.0.0.1]: FAILED! => {"changed": false, "msg": "AnsibleError: template error while templating string: unexpected char u'$' at 300. String: apiVersion: monitoring.coreos.com/v1\nkind: PrometheusRule\nmetadata:\n  labels:\n    monitoring-key: middleware\n  name: {{ gitea_prometheusrule_name }}\nspec:\n  groups:\n    - name: general.rules\n      rules:\n      - alert: GiteaPodCount\n        annotations:\n          message: Pod count for namespace {{ $labels.namespace }} is {{ printf \"%.0f\" $value }}. Expected exactly 3 pods.\n        expr: |\n          (1-absent(kube_pod_status_ready{condition=\"true\", namespace=\"gitea\"})) or sum(kube_pod_status_ready{condition=\"true\", namespace=\"gitea\"}) != 3\n        for: 5m\n        labels:\n          severity: critical"}
```

david-martin:
@odra pushed changes & added verification steps. Good for re-review.

PR #404, opened 2019-02-27
Remove codeowners file

philbrookes:
As I have added all these users to the github branch protection now, this should give us the same protection without having to ping everyone on every PR.

@jasonmadigan any thoughts?


philbrookes:
bah, I can still merge even though not in that list, so this is not working right

jasonmadigan:
if it stops Phil talking to me about email spam, I'm good with this

PR #407, opened 2019-02-27
INTLY-1153 first uninstall the config console

tremes:

### DIFF
--- a/evals/playbooks/uninstall.yml
+++ b/evals/playbooks/uninstall.yml
@@ -2,6 +2,7 @@
 
 - import_playbook: "./prerequisites.yml"
 - import_playbook: "./openshift.yml"
+- import_playbook: "./customise_web_console_uninstall.yml"
 
 - hosts: localhost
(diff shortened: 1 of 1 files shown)


tremes:
It worked on my PDS - https://integreatly-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/uninstallation-pipeline/735/console

JameelB:
I don't think this would work if the web console customization wasn't installed in the first place/has been removed before running the uninstall. I think we should first check if the console-config namespace is there before running this task.

tremes:
Ahh ok I thought the console-config is always installed.

JameelB:
I'm not sure if it will always be, but in cases where the installation fails before it reaches to the point where it installs the web console customization the uninstaller will fail. Maybe it's a bit of an edge case?

tremes:
Actually you're right. Good point. Feel free to close and update it.

PR #409, opened 2019-02-28
INTLY-1153 uninstall the config console & checks

tremes:


odra:
I can give it a try

tremes:
Thanks for review. It worked OK on my PDS cluster as well.

PR #414, opened 2019-03-03
WIP: Automate 3scale backup

aidenkeating:
Only includes MySQL at the moment

### DIFF
--- a/evals/inventories/managed.template
+++ b/evals/inventories/managed.template
@@ -1,0 +1,31 @@
+[local:vars]
+ansible_connection=local
+
+[local]
+127.0.0.1
+
+[OSEv3:children]
+master
+
+[OSEv3:vars]
+ansible_user=ec2-user
+
+[master]
+127.0.0.1
+
+[managed:children]
+master
+local
+
+[managed:vars]
+
+# core_install decides whether to run the main install playbook that sets up all of the core services
+core_install=true
+# backup_restore_install decides whether to run and install the backup and restore features
+backup_restore_install=true
+
+# The namespace where the aws credential secret will be
+aws_credential_secret_namespace= ''
+
+# the secret containing the aws credentials
+aws_credential_secret_name= ''
--- a/evals/inventories/managed.template
+++ b/evals/inventories/managed.template
@@ -25,7 +25,7 @@ core_install=true
 backup_restore_install=true
 
 # The namespace where the aws credential secret will be
-aws_credential_secret_namespace= ''
+aws_credential_secret_namespace=default
 
 # the secret containing the aws credentials
-aws_credential_secret_name= ''
+aws_credential_secret_name=s3-credentials
--- a/evals/roles/3scale/tasks/backup.yml
+++ b/evals/roles/3scale/tasks/backup.yml
@@ -34,5 +34,5 @@
     -p 'BACKEND_SECRET_NAME={{ aws_credential_secret_name }}' \
     -p 'ENCRYPTION_SECRET_NAME={{ threescale_backup_encryption_secret }}' \
     -p 'IMAGE={{ backup_image }}' \
-    -p 'CRON_SCHEDULE="{{ backup_schedule }}"' \
+    -p 'CRON_SCHEDULE={{ backup_schedule }}' \
     -p 'NAME=mysql-backup' | oc create -n {{ threescale_namespace }} -f -
--- a/evals/roles/3scale/tasks/backup.yml
+++ b/evals/roles/3scale/tasks/backup.yml
@@ -8,21 +8,17 @@
     binding_name: threescale_backup_binding
     serviceaccount_namespace: '{{ threescale_namespace }}'
 
-# MySQL secret create
+# MySQL backup
 - name: Get MySQL password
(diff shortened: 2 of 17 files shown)


pb82:
@aidenkeating I see that you create a serviceaccount, role and rolebinding. Those templates are already in the repository: https://github.com/integr8ly/backup-container-image/tree/master/templates/openshift/rbac

evals/roles/3scale/tasks/backup.yml
```diff
@@ -0,0 +1,39 @@
+---
+# Create ServiceAccount
+- name: Create ServiceAccount and role binding
```
pb82: I think we should only create the service account here and setup the cluster role and rolebinding in a separate task (that runs before any backup jobs are created).

aidenkeating: @pb82 Does the ClusterRoleBinding not need to point to the ServiceAccount namespace? So there has to be one ServiceAccount and one ClusterRoleBinding per namespace we create jobs in?

pb82: Ah yes, you're right. I wonder, would it be better to add all middleware namespaces as subjects here: https://github.com/integr8ly/backup-container-image/blob/master/templates/openshift/rbac/role-binding.yaml#L7 ?

aidenkeating: @pb82 I agree that we shouldn't duplicate the RoleBinding yaml in the installer.

I was thinking to template https://github.com/integr8ly/backup-container-image/blob/master/templates/openshift/rbac/role-binding.yaml and make the namespace a parameter. This way, if we need to add more backup namespaces/change namespace names we'll only need to modify the installation repo, not the backup repo. Any thoughts?

pb82: This would be a good solution because it keeps all the work for adding new namespaces confined to the installer repo. The `name` probably also needs to be templated. But in that case i would not leave the cluster role in the backup repo since it would become an Ansible template instead of just a yaml file. We can still lave the yaml files in the backup repo if someone wants to run one-off jobs.

aidenkeating: @pb82 Updated to use the ServiceAccount in the backup repo, I've kept the ClusterRoleBinding template in the installer as an Ansible template.

There seems to be an issue with the image for the backup container though. When using the `latest` tag it's still using an old version of the image.

I don't see any automated builds in Quay, do you know where these images are being pushed from?

pb82:
Getting `"The CronJob \"mysql-backup\" is invalid: spec.schedule: Invalid value: \"\\\"0 0 * * *\\\"\": Failed to parse int from \"0: strconv.Atoi: parsing \"\\\"0\": invalid syntax"`

Seems to be an issue with the template in the master branch.

aidenkeating:
@pb82 This may be an issue with the added backup schedule var which overrides the default "every minute" schedule, I'll have a look into this

evals/roles/3scale/tasks/backup.yml
```diff
@@ -0,0 +1,38 @@
...
+    secret_mysql_password: '{{ threescale_mysql_password.stdout }}'
+
+- name: Create the 3scale MySQL CronJob
+  shell: oc process -f {{ backup_resources_location }}/backup-cronjob-template.yaml \
+    -p 'COMPONENT=mysql' \
+    -p 'COMPONENT_SECRET_NAME={{ threescale_backup_mysql_secret }}' \
+    -p 'BACKEND_SECRET_NAME={{ aws_credential_secret_name }}' \
+    -p 'ENCRYPTION_SECRET_NAME={{ threescale_backup_encryption_secret }}' \
+    -p 'IMAGE={{ backup_image }}' \
+    -p 'CRON_SCHEDULE="{{ backup_schedule }}"' \
```
pb82: I had to remove the double quotes around `{{ backup_schedule }}` to make it create the job. Otherwise it would give me an error: ` "The CronJob \"mysql-backup\" is invalid: spec.schedule: Invalid value: \"\\\"*/1 * * * *\\\"\": Failed to parse int from \"*: strconv.Atoi: parsing \"\\\"*\": invalid syntax"`

aidenkeating: I'll update this now

evals/inventories/managed.template
```diff
@@ -0,0 +1,31 @@
...
+
+[managed:vars]
+
+# core_install decides whether to run the main install playbook that sets up all of the core services
+core_install=true
+# backup_restore_install decides whether to run and install the backup and restore features
+backup_restore_install=true
+
+# The namespace where the aws credential secret will be
+aws_credential_secret_namespace= ''
```
pb82: This should default to `default`

evals/inventories/managed.template
```diff
@@ -0,0 +1,31 @@
...
+
+# the secret containing the aws credentials
+aws_credential_secret_name= ''
```
pb82: can we go with `s3-credentials` for now as that is what i've suggested for the ansible tower work too: https://issues.jboss.org/browse/INTLY-1173

pb82:
@aidenkeating The backup job works once the secrets are in place and the double quotes around the schedule are removed.

aidenkeating:
@pb82 Updated based on your review. I'll take a look into the failing image builds for the backup container now

aidenkeating:
Closing this for a full MySQL, Postgres and Redis PR

PR #422, opened 2019-03-05
Intly 814 amq online install from git repo

mikenairn:
* Pull enmasse(AMQ Online) installation scripts from https://github.com/jboss-container-images/amq-online-images.git

The version that we were pulling from https://access.redhat.com/node/3869211/423/0 is the same as what is in the templates directory of https://github.com/jboss-container-images/amq-online-images.git, and this repo is tagged when new GA releases are created. It will be far easier to track this Git repo, than anything uploaded to access.redhat.com.

### DIFF
--- a/evals/inventories/group_vars/all/manifest.yaml
+++ b/evals/inventories/group_vars/all/manifest.yaml
@@ -23,10 +23,7 @@ che_installer: 'https://raw.githubusercontent.com/redhat-developer/codeready-wor
 #controls whether enmasse is installed or not
 enmasse: true
 enmasse_version: '1.0.0.GA'
-# This is used to indicate where to pull the artifact from for installing EnMasse.
-enmasse_download_url: https://access.redhat.com/node/3869211/423/0
-# Controls whether to use a productised version. Use with 'enmasse_productised_download_url'
-is_productised_version: true
+enmasse_git_url: https://github.com/jboss-container-images/amq-online-images.git
 
 #controls whether fuse is installed or not
 fuse: true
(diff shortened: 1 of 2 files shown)


JameelB:
:eyes:

evals/inventories/group_vars/all/manifest.yaml
```diff
@@ -23,10 +23,7 @@ che_installer: 'https://raw.githubusercontent.com/redhat-developer/codeready-wor
 #controls whether enmasse is installed or not
 enmasse: true
 enmasse_version: '1.0.0.GA'
-# This is used to indicate where to pull the artifact from for installing EnMasse.
-enmasse_download_url: https://access.redhat.com/node/3869211/423/0
-# Controls whether to use a productised version. Use with 'enmasse_productised_download_url'
-is_productised_version: true
```
mikenairn: Not sure what the point of this was tbh. It just seemed to dictate a different directory structure, but i would assume we should always be installing the same thing i.e. The thing we just pulled via git.

odra: back at the time it was done to have a different "installation path" since we were using a beta release url just in case we wanted to go back to the upstream url.

This is not needed anymore.

JameelB: @mikenairn Was just going to ask about that, I think that may be out of date. The downloaded folder only has [this](https://github.com/integr8ly/installation/blob/master/evals/roles/enmasse/tasks/install.yml#L32-L33) structure when I've downloaded it. 

It would be great to remove it if we can. Would there be a chance where we want to install a non-productised version of Enmasse?

odra: you also don't need other files such as the `prereqs.sh` one

JameelB:
Code changes looks good, trying it out on PDS

PR #423, opened 2019-03-05
Resources backup

pb82:
Sets up a CronJob to backup the kubernetes resources.

Verification:

1. Create a secret in default named s3-credentials that contains your credentials
1. Update the master URL in managed.template and set core_install=false if you've already setup integreatly
1. Set backup_schedule in manifest.yaml to '*/1 * * * *' to run each minute
1. Run: ansible-playbook -i inventories/managed.template playbooks/managed/install.yml
1. Wait a minute
1. In the default namespace
1.1 Ensure the resources-backup Pod succeeds
1.1 Check S3 and ensure the archives have been pushed up.
1.1 Delete the CronJobs in the default and 3scale namespaces (otherwise it keeps creating backups every minute)

### DIFF
--- a/evals/roles/resources_backup/tasks/main.yml
+++ b/evals/roles/resources_backup/tasks/main.yml
@@ -12,6 +12,7 @@
 - name: Create the resources backup job
   shell: oc process -f {{ backup_resources_location }}/backup-cronjob-template.yaml \
     -p 'BACKEND_SECRET_NAME={{ aws_credential_secret_name }}' \
+    -p 'COMPONENT=resources' \
     -p 'IMAGE={{ backup_image }}' \
     -p 'CRON_SCHEDULE={{ backup_schedule }}' \
     -p 'NAME=resources-backup' | oc create -n default -f -
--- a/evals/roles/resources_backup/tasks/main.yml
+++ b/evals/roles/resources_backup/tasks/main.yml
@@ -8,6 +8,10 @@
     binding_name: resources_backup_binding
     serviceaccount_namespace: default
 
+# Delete the old backup cronjob in case this is a re-installation
+- name: Remove old cronjob
+  shell: oc delete cronjob resources-backup --ignore-not-found
+
 # Kube resources backup backup
 - name: Create the resources backup job
   shell: oc process -f {{ backup_resources_location }}/backup-cronjob-template.yaml \
--- a/evals/roles/resources_backup/tasks/main.yml
+++ b/evals/roles/resources_backup/tasks/main.yml
@@ -1,0 +1,19 @@
+---
+# Create ServiceAccount in the default namespace
+- name: Create ServiceAccount and role binding
+  include_role:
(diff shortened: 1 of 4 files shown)


sedroche:
:eyes:

evals/roles/resources_backup/tasks/main.yml
```diff
@@ -0,0 +1,20 @@
...
+# Kube resources backup backup
+- name: Create the resources backup job
+  shell: oc process -f {{ backup_resources_location }}/backup-cronjob-template.yaml \
+    -p 'BACKEND_SECRET_NAME={{ aws_credential_secret_name }}' \
+    -p 'COMPONENT=resources' \
+    -p 'IMAGE={{ backup_image }}' \
+    -p 'CRON_SCHEDULE={{ backup_schedule }}' \
+    -p 'NAME=resources-backup' | oc create -n default -f -
+  register: resources_cronjob_create
+  failed_when: resources_cronjob_create.stderr != '' and 'AlreadyExists' not in resources_cronjob_create.stderr
```
sedroche: Does it make sense to delete and recreate the job? As any changed parameters wont be picked up this way.

evals/roles/resources_backup/tasks/main.yml
```diff
@@ -0,0 +1,20 @@
...
+- name: Create ServiceAccount and role binding
+  include_role:
+    name: backup
+    tasks_from: _setup_service_account.yml
+  vars:
+    binding_name: resources_backup_binding
+    serviceaccount_namespace: default
+
+# Kube resources backup backup
+- name: Create the resources backup job
```
sedroche: Just wondering if we can have a reusable task like the above task.

```
{{ backup_resources_location }}/backup-cronjob-template.yaml
 BACKEND_SECRET_NAME={{ aws_credential_secret_name }}
IMAGE={{ backup_image }}
```

will probably be the same and will need to be updated in lots of places otherwise (I see 3 so far)

sedroche:
Verified

- Cronjob executes every minute with this schedule.
- Artifact contains the kubernetes resources

evals/roles/resources_backup/tasks/main.yml
```diff
@@ -0,0 +1,20 @@
+---
+# Create ServiceAccount in the default namespace
+- name: Create ServiceAccount and role binding
```
odra: Same as above, ok for this pr but I think we should add those as part of a `backup prepare` role or something like that

pb82: It already is inside a general `backup` role, but the serviceaccount has to be created in every namespace, so that's why we invoke it from here.

evals/roles/resources_backup/tasks/main.yml
pb82: Yes, i think that makes sense. Will add a task for it.

Issue #402, opened 2019-02-26
Managed Fuse install happens for every run/re-run of the installer


david-martin:
Logs during a re-run of installer

```
TASK [fuse_managed : Create fuse namespace if it doesn't exist] ****************
changed: [127.0.0.1]

TASK [fuse_managed : Add labels to namespace] **********************************
changed: [127.0.0.1]

TASK [fuse_managed : Create Syndesis CRD] **************************************
changed: [127.0.0.1]

TASK [fuse_managed : Create Fuse image streams] ********************************
changed: [127.0.0.1]

TASK [fuse_managed : Create fuse Operator resources in fuse] *******************
changed: [127.0.0.1]

TASK [fuse_managed : template] *************************************************
ok: [127.0.0.1]

TASK [fuse_managed : Create Syndesis custom resource in fuse] ******************
changed: [127.0.0.1]

TASK [fuse_managed : Verify Fuse deployment succeeded] *************************
```
Looks like there are no `when` clauses  in https://github.com/integr8ly/installation/blob/master/evals/roles/fuse_managed/tasks/main.yml

This causes a recreate of the operator & reinstall due to `oc replace --force` being used

david-martin:
@aidenkeating Do you recall if there is a reason we use `oc replace --force` here https://github.com/integr8ly/installation/blob/master/evals/roles/fuse_managed/tasks/main.yml#L18

david-martin:
I'm not sure why yet, but running install results in a timeout on the running pod count (4 != 7)
I've seen this happen at least twice recently

```
NAME                          READY     STATUS             RESTARTS   AGE
syndesis-db-1-96fmm           0/1       Pending            0          7m
syndesis-db-1-deploy          1/1       Running            0          7m
syndesis-meta-1-4rzqc         0/1       Pending            0          7m
syndesis-meta-1-deploy        1/1       Running            0          7m
syndesis-oauthproxy-1-r6dz5   1/1       Running            0          7m
syndesis-operator-1-8jccs     1/1       Running            0          18m
syndesis-prometheus-1-dzjjp   1/1       Running            0          7m
syndesis-server-1-deploy      1/1       Running            0          7m
syndesis-server-1-rbk8j       0/1       CrashLoopBackOff   5          7m
syndesis-ui-1-fgg9j           1/1       Running            0          7m
todo-1-build                  0/1       Completed          0          7m
todo-1-ppg5b                  1/1       Running            0          7m
```

```
[...]

PR #430, opened 2019-03-06
Enmasse backup

pb82:
Creates the enmasse backup cronjob.

Sets up a CronJob to backup the kubernetes resources.

Verification:

To test that the files in the PV are backed up you need a brokered AMQ:

1. Go to the service catalog and select *AMQ Online (brokered)*. Pick the enmasse namespace and a name and provision the brokered quee.
1. In your enmasse namespace you should see an additional pod for the brokered queue.
1. Open the terminal of this pod and go to `/var/run/artemis`
1. Create a file in the, like `echo "test" > /var/run/artemis/test.txt`

Then you can verify the backup using:

1. Make sure that the s3-credentials secret exists in the default namespace
1. Update the master URL in managed.template and set core_install=false if you've already setup integreatly
1. Set backup_schedule in manifest.yaml to '*/1 * * * *' to run each minute
1. Run: ansible-playbook -i inventories/managed.template playbooks/managed/install.yml
1. Wait a minute
 
In the enmasse namespace
1.1 Ensure the postgres-backup Pod succeeds
1.1 Ensure the PV backup succeeds and the archive contains the file generated for the brokered pod.
1.1 Check S3 and ensure the archives have been pushed up.
1.1 Delete the CronJobs in all namespaces

### DIFF
--- a/evals/roles/enmasse/defaults/main.yml
+++ b/evals/roles/enmasse/defaults/main.yml
@@ -9,3 +9,4 @@ enmasse_keycloak_admin_password: admin
 enmasse_authentication_services: ["standard"]
 enmasse_clean_artifacts: true
 enmasse_enable_monitoring: false
+enmasse_backup_postgres_secret: 'enmasse-postgres-secret'
--- a/evals/roles/enmasse/defaults/main.yml
+++ b/evals/roles/enmasse/defaults/main.yml
@@ -10,5 +10,5 @@ enmasse_authentication_services: ["standard"]
 enmasse_clean_artifacts: true
 enmasse_enable_monitoring: false
 enmasse_backup_postgres_secret: 'enmasse-postgres-secret'
-enmasse_postgres_cronjob_name: 'postgres-backup'
-enmasse_pv_cronjob_name: 'pv-backup'
+enmasse_postgres_cronjob_name: 'enmasse-postgres-backup'
+enmasse_pv_cronjob_name: 'enmasse-pv-backup'
--- a/evals/roles/enmasse/defaults/main.yml
+++ b/evals/roles/enmasse/defaults/main.yml
@@ -10,3 +10,4 @@ enmasse_authentication_services: ["standard"]
 enmasse_clean_artifacts: true
 enmasse_enable_monitoring: false
 enmasse_backup_postgres_secret: 'enmasse-postgres-secret'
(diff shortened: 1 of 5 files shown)


evals/roles/enmasse/tasks/backup.yml
```diff
@@ -0,0 +1,48 @@
...
+  vars:
+    secret_name: '{{ enmasse_backup_postgres_secret }}'
+    secret_postgres_user: '{{ enmasse_postgres_username.stdout }}'
+    secret_postgres_host: postgresql.enmasse.svc
+    secret_postgres_database: '{{ enmasse_postgres_database.stdout }}'
+    secret_postgres_password: '{{ enmasse_postgres_password.stdout }}'
+
+# Delete the old backup cronjob in case this is a re-installation
+- name: Remove old cronjob
+  shell: oc delete cronjob {{ enmasse_backup_cronjob_name }} --ignore-not-found -n {{ enmasse_namespace }}
```
david-martin: @pb82 I'd prefer to avoid this pattern during install.
If the install failed for some reason, and had to be re-run, it should be idempotent i.e. not remove or recreate anything, but create anything that is missing.

This has caused issues for us in the past e.g. https://github.com/integr8ly/installation/pull/73 & https://github.com/integr8ly/installation/issues/402

pb82: Interesting, the idea was to make sure the job gets updated when the properties change but for that we can use `apply`. Will update this.

matskiv:
👀

evals/roles/enmasse/defaults/main.yml
```diff
@@ -8,4 +8,7 @@ enmasse_service_catalog: true
 enmasse_keycloak_admin_password: admin
 enmasse_authentication_services: ["standard"]
 enmasse_clean_artifacts: true
-enmasse_enable_monitoring: false
\ No newline at end of file
+enmasse_enable_monitoring: false
+enmasse_backup_postgres_secret: 'enmasse-postgres-secret'
+enmasse_postgres_cronjob_name: 'postgres-backup'
+enmasse_pv_cronjob_name: 'pv-backup'
```
matskiv: Since we are creating all CronJobs in default namespace, two names above should be prepended with 'enmasse-', just to avoid any ambiguity.

PR #434, opened 2019-03-07
Add manifest var and release script

maleck13:
## Additional Information
The release in 1.3 will only need to tag the installation repo

## Verification Steps
Run the installation
Ensure the version is populated in the manifest in the cluster
``` oc get secret manifest -n webapp --template '{{index .data "generated_manifest"}}'  | base64 -d ```

Test the release script

- start on this branch and run the release script. Ensure to set your own origin
``` 
export REMOTE=mine
./scripts/release.sh -b v1.3 -r release-1.3.25-rc1-test

```

New v1.3 branch should exist on the remote and tag created
Run Again with new rc
```
export REMOTE=mine
./scripts/release.sh -b v1.3 -r release-1.3.25-rc2-test

```

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -1,0 +1,92 @@
+#!/usr/bin/env bash
+
+REMOTE=${REMOTE:-origin}
+
+#check the current branch we are on
+currentBranch=$(git symbolic-ref --short HEAD)
+baseBranch=""
+releaseTag=""
+
+while getopts ":b:r:h" opt; do
+  echo "opt $opt"
+  case ${opt} in
+    b)
+      baseBranch=${OPTARG}
+      ;;
+    r)
+      releaseTag=${OPTARG}
+      ;;
+    h)
+      echo "This script will create a new integreatly release.
+      Flags:
+       - b <branch to cut the release from>
+       - r <the release to create (release-1.3.1)
+       - h help
+       "
+       exit
+      ;;
+    \?)
+      echo "Invalid option: -$OPTARG" >&2
+      exit 1
+      ;;
+    :)
+      echo "Option -$OPTARG requires an argument." >&2
+      exit 1
+      ;;
+  esac
+done
+
+if [[ "$baseBranch" == "" || "$releaseTag" == "" ]]
+then
+echo "Flags:
+       - b <branch to cut the release from>
+       - r <the release to create (release-1.3.1)
+       are required
+"
+exit 1
+fi
+
+echo "cutting release ${releaseTag} from branch ${REMOTE}/${baseBranch}"
+#do a fetch
+
+git fetch ${REMOTE}
+
+#check we have no local changes
+dirty=$(git ls-files -m | wc -l)
+if [[ "${dirty}" -gt "0" ]]
+then
+echo "
+ the local file system is dirty cannot continue
+"
+exit 1
+fi
+
+# check if the specified from branch already exists if it does check it out otherwise create it
+
+branchExists=$(git checkout -B ${REMOTE}/${baseBranch})
+if [[ $? > 0 ]]
+then
+echo "branch ${baseBranch} does not exist or you have local changes. Please create it before running the release"
+exit 1
+fi
+
+
+
+releaseExists=$(git tag | grep ${releaseTag} | wc -l)
+if [[ "${releaseExists}" -gt "0" ]]
+then
+echo "
+a release with that name already exists
+"
+exit 1
+fi
+
+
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by maleck13** [comment]
@mikenairn @david-martin  would be good to get a review if you get chance

File: scripts/release.sh [hunk 1]
```diff
@@ -0,0 +1,76 @@
+#!/usr/bin/env bash
+
+#check the current branch we are on
+currentBranch=$(git branch | grep \* | cut -d ' ' -f2)
```
mikenairn: You could also use `git symbolic-ref --short HEAD` or `git rev-parse --abbrev-ref HEAD`

scripts/release.sh
```diff
@@ -0,0 +1,76 @@
...
+if [[ "$carryON" == "y" ]]
+then
+echo "
+what branch do you want?"
+read  -p "Input: " branch
+echo "
+creating new branch $branch
+"
+
+git checkout -b ${branch}
```
mikenairn: I'd guess it should do a fetch somewhere near the beginning to make sure there isn't already a branch created. Any reason we are letting the user input a branch name? Just seem likely to go wrong the more we ask the person doing it to manually input things.

scripts/release.sh
```diff
@@ -0,0 +1,76 @@
...
+
+if [[ $? > 0 ]]
+then
+exit "command failed" $?
+fi
+
+fi
+
+#check we are upto date and no local changes
+dirty=$(git ls-files -m | wc -l)
```
mikenairn: Do we care about untracked files?

scripts/release.sh
```diff
@@ -0,0 +1,76 @@
...
+ the local file system is dirty cannot continue
+"
+exit
+fi
+
+echo "
+what release version do you want to create (e.g. release-v1.3.0-rc1)?"
+read  -p "Input: " release
+
+releaseExists=$(git tag | grep release-1.2.0-rc6 | wc -l)
```
mikenairn: Hard coded "release-1.2.0-rc6", assume it;s supposed to be release?

mikenairn: Again, need to force a fetch somewhere to make sure your local refs are up to date.

scripts/release.sh
```diff
@@ -0,0 +1,76 @@
...
+"
+exit
+fi
+
+#update the manifest with the release tag
+sed -i.bak -E "s/^integreatly_version: .*$/integreatly_version: $release/g"  ./evals/inventories/group_vars/all/manifest.yaml && rm ./evals/inventories/group_vars/all/manifest.yaml.bak
+
+#commit the change and push
+git commit -am "release manifest version  update for $release"
+git push origin ${branch}
```
mikenairn: Assumes the remote we are pushing to is called "origin", wouldn't always be the case if someone was running this locally.

mikenairn:
I know we discussed using a script like this offline, but having seen it, i have mixed feelings about doing it this way. As it is, it is very dependant on the person running it having everything correct locally, and knowing what format tags/branches should be in. 
Personally i'd much rather the script was more opinionated about what it was going to do when creating a new release and didn't rely on user input so much, in a similar way that the pipeline we currently have just asks for a version tag. I also think even though we have a script in here, we should consider running it in a pipeline anyway to avoid a users local setup becoming an issue. It can still be executed locally if needs be.
Might also be more user friendly if we used python instead of bash, but maybe that's subjective.

maleck13:
@mikenairn Perhaps the right option here is to add flags that will allow it be done non interactively. Then it could hooked in to pipeline easily?

scripts/release.sh
maleck13: good catch will change

scripts/release.sh
maleck13: I don't think so?

scripts/release.sh
maleck13: Yeah a fetch makes sense to ensure we are not creating a branch that already exists

scripts/release.sh
maleck13: fair point. Can be changed to cover it by a flag

scripts/release.sh
maleck13: We ask for a gitref currently in the release pipeline

david-martin:
@maleck13 Does this look right?

```
oc get secret manifest -n webapp --template '{{index .data "generated_manifest"}}' | base64 -d
{
    "version": "master",
    "components":[
```
The version is at the top level of the manifest.

Also, I've pushed a fix for the branch creation/check.

maleck13:
@david-martin yes that looks correct

david-martin:
Merging. Lets try this out for 1.3.0

PR #438, opened 2019-03-11
Codeready backups

pb82:
Adds the codeready backup jobs.

*NOTE*: at the moment the PV backup job will fail if there are no workspaces. I'll update the backup container image in a separate PR to prevent this.

### DIFF
--- a/evals/roles/code-ready/tasks/backup.yml
+++ b/evals/roles/code-ready/tasks/backup.yml
@@ -39,6 +39,7 @@
 - name: Create the enmasse Postgres CronJob
   shell: oc process -f {{ backup_resources_location }}/backup-cronjob-template.yaml \
     -p 'COMPONENT=postgres' \
+    -p 'PRODUCT_NAME=codeready' \
     -p 'COMPONENT_SECRET_NAME={{ codeready_backup_postgres_secret }}' \
     -p 'BACKEND_SECRET_NAME={{ aws_credential_secret_name }}' \
     -p 'IMAGE={{ backup_image }}' \
@@ -49,6 +50,7 @@
 - name: Create the codeready PV CronJob
   shell: oc process -f {{ backup_resources_location }}/backup-cronjob-template.yaml \
     -p 'COMPONENT=codeready_pv' \
+    -p 'PRODUCT_NAME=codeready' \
     -p 'BACKEND_SECRET_NAME={{ aws_credential_secret_name }}' \
     -p 'IMAGE={{ backup_image }}' \
     -p 'CRON_SCHEDULE={{ backup_schedule }}' \
--- a/evals/roles/code-ready/tasks/backup.yml
+++ b/evals/roles/code-ready/tasks/backup.yml
@@ -10,19 +10,19 @@
 
 # Postgres backup
 - name: Get the name of the postgres pod
-  shell: oc get pods --selector='component=postgres' -o jsonpath='{.items[0].metadata.name}'
+  shell: oc get pods -n {{ che_namespace }} --selector='component=postgres' -o jsonpath='{.items[0].metadata.name}'
   register: codeready_postgres_pod
 
 - name: Get Postgres password
-  shell: oc exec {{ codeready_postgres_pod.stdout }} printenv | grep POSTGRESQL_PASSWORD | awk -F"=" '{print $2}'
+  shell: oc exec -n {{ che_namespace }} {{ codeready_postgres_pod.stdout }} printenv | grep POSTGRESQL_PASSWORD | awk -F"=" '{print $2}'
   register: codeready_postgres_password
 
 - name: Get Postgres user
-  shell: oc exec {{ codeready_postgres_pod.stdout }} printenv | grep POSTGRESQL_USER | awk -F"=" '{print $2}'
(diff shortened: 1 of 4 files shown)


evals/roles/code-ready/tasks/backup.yml
```diff
@@ -0,0 +1,55 @@
...
+- name: Get the name of the postgres pod
+  shell: oc get pods -n {{ che_namespace }} --selector='component=postgres' -o jsonpath='{.items[0].metadata.name}'
+  register: codeready_postgres_pod
+
+- name: Get Postgres password
+  shell: oc exec -n {{ che_namespace }} {{ codeready_postgres_pod.stdout }} printenv | grep POSTGRESQL_PASSWORD | awk -F"=" '{print $2}'
+  register: codeready_postgres_password
+
+- name: Get Postgres user
+  shell: oc exec -n {{ che_namespace }} {{ codeready_postgres_pod.stdout }} printenv | grep POSTGRESQL_USER | awk -F"=" '{print $2}'
```
sedroche: You can get these from the deployment like:
```bash
oc get deployment postgres -n codeready -o jsonpath='{ .spec.template.spec.containers[?(@.name=="postgres")].env[?(@.name=="POSTGRESQL_USER")].value }'
```

evals/roles/code-ready/tasks/backup.yml
```diff
@@ -0,0 +1,55 @@
...
+    secret_postgres_password: '{{ codeready_postgres_password.stdout }}'
+
+- name: Create the enmasse Postgres CronJob
+  shell: oc process -f {{ backup_resources_location }}/backup-cronjob-template.yaml \
+    -p 'COMPONENT=postgres' \
+    -p 'COMPONENT_SECRET_NAME={{ codeready_backup_postgres_secret }}' \
+    -p 'BACKEND_SECRET_NAME={{ aws_credential_secret_name }}' \
+    -p 'IMAGE={{ backup_image }}' \
+    -p 'CRON_SCHEDULE={{ backup_schedule }}' \
+    -p 'NAME={{ codeready_postgres_cronjob_name }}' | oc apply -n default -f -
```
odra: Can you add the following param `-p 'PRODUCT_NAME=codeready` as part of https://github.com/integr8ly/backup-container-image/pull/26

evals/roles/code-ready/tasks/backup.yml
```diff
@@ -0,0 +1,55 @@
...
+
+# PV backup
+- name: Create the codeready PV CronJob
+  shell: oc process -f {{ backup_resources_location }}/backup-cronjob-template.yaml \
+    -p 'COMPONENT=codeready_pv' \
+    -p 'BACKEND_SECRET_NAME={{ aws_credential_secret_name }}' \
+    -p 'IMAGE={{ backup_image }}' \
+    -p 'CRON_SCHEDULE={{ backup_schedule }}' \
+    -p 'NAME={{ codeready_pv_cronjob_name }}' | oc apply -n default -f -
```
odra: Same as above

evals/roles/code-ready/tasks/backup.yml
pb82: That also works but is more complicated because it involves iterating the containers. Is there an advantage to not using `exec`?

evals/roles/code-ready/tasks/backup.yml
pb82: done

evals/roles/code-ready/tasks/backup.yml
pb82: done

evals/roles/code-ready/tasks/backup.yml
sedroche: It just avoids having to query the name of the postgres pod but the result is the same so i'm ok with `exec` too

sedroche:
Verifying

evals/roles/code-ready/tasks/backup.yml
```diff
@@ -0,0 +1,58 @@
...
+    -p 'IMAGE={{ backup_image }}' \
+    -p 'CRON_SCHEDULE={{ backup_schedule }}' \
+    -p 'NAME={{ codeready_postgres_cronjob_name }}' | oc apply -n default -f -
+
+# PV backup
+- name: Create the codeready PV CronJob
+  shell: oc process -f {{ backup_resources_location }}/backup-cronjob-template.yaml \
+    -p 'COMPONENT=codeready_pv' \
+    -p 'PRODUCT_NAME=codeready' \
+    -p 'BACKEND_SECRET_NAME={{ aws_credential_secret_name }}' \
```
odra: could we set a default for for this var `aws_credential_secret_name` to "s3-credentials"?

matskiv: Can you actually use `aws_s3_backup_secret_name` var? 
In my last PR I renamed it from 3 different names to this^ single one.

matskiv: It should work after rebase :)

pb82: Ok, will update

PR #450, opened 2019-03-12
INTLY-1100 Add backup monitoring for CronJobs

aidenkeating:
Add a set of tasks for creating alerts that are triggered when
CronJobs and the Jobs they create are not in a correct state.

A summary of the alerts added in this commit:

- `JobRunningTimeExceeded` a single Job has been running for more
than five minutes, meaning it may be in debug mode in stuck.
- `CronJobSuspended` a CronJob that is being monitored has been
set to a suspended state, meaning Jobs won't be created.
- `CronJobNotRunInThreshold` a Job has not been created for a
CronJob in 25 hours, something may be wrong with creating the Job
- `CronJobExists_*` a CronJob does not exist in the cluster that
existed when the cluster was provisioned

All monitored CronJobs and Jobs are expected to have a label on
them named `monitoring-key` with the value `middleware`.

For a `CronJobExists_*` check to be created for a CronJob, the
CronJob must exist before the `monitoring.yml` tasks are run.

Install Verification:
- In `manifest.yml`, update `backup_resources_location` to point to
`https://raw.githubusercontent.com/aidenkeating/backup-container-image/INTLY-1100/templates/openshift`.
- In `manifest.yml`, update `backup_schedule` to `*/1 * * * *`.
- Run the managed install playbook.
- Ensure the install completes successfully.
- Ensure the created CronJobs have labels with the key
`monitoring-key` and the value `middleware`.
- Ensure that the Jobs created by the CronJobs have a label with
the key `monitoring-key` and the value `middleware`.
- Ensure the created CronJobs have the concurrency policy set to
`Forbid`.
- Go to the Prometheus route from the `middleware-monitoring`
namespace.
- Ensure their are the following alerts in the `Alerts` tab:
  - CronJobExists_* (One for each CronJob)
  - CronJobNotRunInThreshold
  - CronJobSuspended
  - JobRunningTimeExceeded

Alert Verification:
- Note that all alerts will take a few minutes to trigger as they
are based off of federated data which takes a while to sync
- Delete one of the CronJobs (3scale-mysql-backup) and ensure the
corresponding Alert is triggered
- `oc edit` one of the CronJobs (3scale-redis-backup), set
`.spec.suspend` to `true, save,` and ensure the CronJobSuspended
Alert is triggered for that CronJob
- `oc edit` one of the CronJobs (3scale-postgres-backup), set the
final line of `.spec.command` to `true` instead of being a blank
string. This should cause the next Job for that CronJob to go to
debug mode and sleep forever. Wait a few minutes and ensure the
`JobRunningTimeExceeded` Alert is triggered
- To test `CronJobNotRunInThreshold`, modify the PrometheusRule in
the `middleware-monitoring` namespace and change the `60*60*25`
(25 hours) to something small (`1`). You should see the Alert
triggering for the "sleeping" Job you created in the last step.
Note that due to federation, for small values, multiple Alerts
may trigger as the start times of jobs are out-of-date in our
Prometheus.

### DIFF
--- a/roles/backup/templates/backup-monitoring-alerts.yml.j2
+++ b/roles/backup/templates/backup-monitoring-alerts.yml.j2
@@ -1,0 +1,40 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  labels:
+    monitoring-key: middleware
+  name: backup-monitoring-alerts
+spec:
+  groups:
+    - name: general.rules
+      rules:
+      - alert: JobRunningTimeExceeded
+        annotations:
+          message: Job {{ '{{' }} $labels.namespace {{ '}}' }}/{{ '{{' }} $labels.job {{ '}}' }} has been running for longer than 5 minutes
+        expr: |
+          time() - (max(kube_job_status_active * ON(job) GROUP_RIGHT() kube_job_labels{label_monitoring_key="middleware"}) BY (job) * ON(job) GROUP_RIGHT() max(kube_job_status_start_time * ON(job) GROUP_RIGHT() kube_job_labels{label_monitoring_key="middleware"}) BY (job, namespace, label_cronjob_name) > 0) > 300
+        labels:
+          severity: critical
+      - alert: CronJobSuspended
+        annotations:
+          message: CronJob {{ '{{' }} $labels.namespace {{ '}}' }}/{{ '{{' }} $labels.cronjob {{ '}}' }} is suspended
+        expr: |
+          kube_cronjob_labels{ label_monitoring_key="middleware" } * ON (cronjob) GROUP_RIGHT() kube_cronjob_spec_suspend > 0
+        for: 60s
+        labels:
+          severity: critical
+      - alert: CronJobNotRunInThreshold
+        annotations:
+          message: CronJob {{ '{{' }} $labels.namespace {{ '}}' }}/{{ '{{' }} $labels.label_cronjob_name {{ '}}' }} has not started a Job in
+        expr: |
+          (time() - (max( kube_job_status_start_time * ON(job) GROUP_RIGHT() kube_job_labels{label_monitoring_key="middleware"} ) BY (job, label_cronjob_name) == ON(label_cronjob_name) GROUP_LEFT() max( kube_job_status_start_time * ON(job) GROUP_RIGHT() kube_job_labels{label_monitoring_key="middleware"} ) BY (label_cronjob_name))) > 60*60*25
+      {% for cronjob in expected_cronjobs %}
+- alert: CronJobExists_{{ cronjob }}
+        annotations:
+          message: CronJob {{ '{{' }} $labels.namespace {{ '}}' }}/{{ '{{' }} $labels.cronjob {{ '}}' }} does not exist
+        expr: |
+          absent(kube_cronjob_info{cronjob="{{ cronjob }}", namespace="default"})
+        for: 60s
+        labels:
+          severity: critical
+      {% endfor %}
--- a/roles/backup/templates/backup-monitoring-alerts.yml.j2
+++ b/roles/backup/templates/backup-monitoring-alerts.yml.j2
@@ -29,11 +29,11 @@ spec:
         expr: |
           (time() - (max( kube_job_status_start_time * ON(job) GROUP_RIGHT() kube_job_labels{label_monitoring_key="middleware"} ) BY (job, label_cronjob_name) == ON(label_cronjob_name) GROUP_LEFT() max( kube_job_status_start_time * ON(job) GROUP_RIGHT() kube_job_labels{label_monitoring_key="middleware"} ) BY (label_cronjob_name))) > 60*60*25
       {% for cronjob in expected_cronjobs %}
-- alert: CronJobExists_{{ cronjob }}
+- alert: CronJobExists_{{ cronjob.name }}
         annotations:
           message: CronJob {{ '{{' }} $labels.namespace {{ '}}' }}/{{ '{{' }} $labels.cronjob {{ '}}' }} does not exist
         expr: |
-          absent(kube_cronjob_info{cronjob="{{ cronjob }}", namespace="default"})
+          absent(kube_cronjob_info{cronjob="{{ cronjob.name }}", namespace="{{ cronjob.namespace }}"})
         for: 60s
         labels:
           severity: critical
--- a/roles/backup/templates/backup-monitoring-alerts.yml.j2
+++ b/roles/backup/templates/backup-monitoring-alerts.yml.j2
@@ -9,13 +9,15 @@ spec:
   groups:
     - name: general.rules
       rules:
+      {% for check in [{ 'time': '300', 'severity': 'warning' }, { 'time': '600', 'severity': 'critical' }] %}
       - alert: JobRunningTimeExceeded
         annotations:
-          message: Job {{ '{{' }} $labels.namespace {{ '}}' }}/{{ '{{' }} $labels.job {{ '}}' }} has been running for longer than 5 minutes
+          message: Job {{ '{{' }} $labels.namespace {{ '}}' }}/{{ '{{' }} $labels.job {{ '}}' }} has been running for longer than {{ check['time'] }} seconds
         expr: |
-          time() - (max(kube_job_status_active * ON(job) GROUP_RIGHT() kube_job_labels{label_monitoring_key="middleware"}) BY (job) * ON(job) GROUP_RIGHT() max(kube_job_status_start_time * ON(job) GROUP_RIGHT() kube_job_labels{label_monitoring_key="middleware"}) BY (job, namespace, label_cronjob_name) > 0) > 300
+          time() - (max(kube_job_status_active * ON(job) GROUP_RIGHT() kube_job_labels{label_monitoring_key="middleware"}) BY (job) * ON(job) GROUP_RIGHT() max(kube_job_status_start_time * ON(job) GROUP_RIGHT() kube_job_labels{label_monitoring_key="middleware"}) BY (job, namespace, label_cronjob_name) > 0) > {{ check['time'] }}
         labels:
-          severity: critical
+          severity: {{ check['severity'] }}
+      {% endfor %}
       - alert: CronJobSuspended
         annotations:
           message: CronJob {{ '{{' }} $labels.namespace {{ '}}' }}/{{ '{{' }} $labels.cronjob {{ '}}' }} is suspended
--- a/roles/backup/templates/backup-monitoring-alerts.yml.j2
+++ b/roles/backup/templates/backup-monitoring-alerts.yml.j2
@@ -1,3 +1,4 @@
+#jinja2: lstrip_blocks: True
 apiVersion: monitoring.coreos.com/v1
 kind: PrometheusRule
 metadata:
@@ -25,16 +26,18 @@ spec:
           severity: critical
       - alert: CronJobNotRunInThreshold
         annotations:
-          message: CronJob {{ '{{' }} $labels.namespace {{ '}}' }}/{{ '{{' }} $labels.label_cronjob_name {{ '}}' }} has not started a Job in
+          message: CronJob {{ '{{' }} $labels.namespace {{ '}}' }}/{{ '{{' }} $labels.label_cronjob_name {{ '}}' }} has not started a Job in 25 hours
         expr: |
           (time() - (max( kube_job_status_start_time * ON(job) GROUP_RIGHT() kube_job_labels{label_monitoring_key="middleware"} ) BY (job, label_cronjob_name) == ON(label_cronjob_name) GROUP_LEFT() max( kube_job_status_start_time * ON(job) GROUP_RIGHT() kube_job_labels{label_monitoring_key="middleware"} ) BY (label_cronjob_name))) > 60*60*25
-      {% for cronjob in expected_cronjobs %}
-- alert: CronJobExists_{{ cronjob.name }}
+      {% for namespace, cronjobs in expected_cronjobs.iteritems() %}
+      {% for cronjob in cronjobs %}
+      - alert: CronJobExists_{{ namespace}}_{{ cronjob }}
         annotations:
           message: CronJob {{ '{{' }} $labels.namespace {{ '}}' }}/{{ '{{' }} $labels.cronjob {{ '}}' }} does not exist
         expr: |
-          absent(kube_cronjob_info{cronjob="{{ cronjob.name }}", namespace="{{ cronjob.namespace }}"})
+          absent(kube_cronjob_info{cronjob="{{ cronjob }}", namespace="{{ namespace }}"})
         for: 60s
         labels:
           severity: critical
+      {% endfor %}
       {% endfor %}
--- a/inventories/group_vars/all/manifest.yaml
+++ b/inventories/group_vars/all/manifest.yaml
@@ -95,4 +95,4 @@ msbroker_template: 'https://raw.githubusercontent.com/integr8ly/managed-service-
 backup_version: master
 backup_resources_location: 'https://raw.githubusercontent.com/integr8ly/backup-container-image/{{ backup_version }}/templates/openshift'
 backup_image: quay.io/integreatly/backup-container:{{ backup_version }}
-backup_schedule: '0 0 * * *'
+backup_schedule: '30 2 * * *'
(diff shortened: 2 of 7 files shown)


odra:
:eyes:

aidenkeating:
@pat-cremin @david-martin I'd say you probably know Prometheus queries better than I do, mind taking a look and seeing if those look ok to you?

@odra @pb82 @sedroche Mind taking a look?

odra:
giving it a run in my cluster

odra:
getting this error:

```
TASK [backup : Create Postgres secret threescale-postgres-secret] ************************************************************************************
fatal: [127.0.0.1]: FAILED! => {"changed": true, "cmd": "oc apply -f /tmp/postgres-secret.yml -n default", "delta": "0:00:00.334504", "end": "2019-03-12 14:19:48.505813", "failed_when_result": true, "rc": 0, "start": "2019-03-12 14:19:48.171309", "stderr": "Warning: oc apply should be used on resource created by either oc create --save-config or oc apply", "stderr_lines": ["Warning: oc apply should be used on resource created by either oc create --save-config or oc apply"], "stdout": "secret/threescale-postgres-secret configured", "stdout_lines": ["secret/threescale-postgres-secret configured"]}
```

pb82:
@odra We could use `apply` instead of create. Some of the backup tasks already do this.

aidenkeating:
@odra I see what's causing it, I'll resolve this. Not sure how this didn't show up before, I haven't changed those tasks at all

roles/backup/templates/backup-monitoring-alerts.yml.j2
```diff
@@ -0,0 +1,40 @@
...
+          kube_cronjob_labels{ label_monitoring_key="middleware" } * ON (cronjob) GROUP_RIGHT() kube_cronjob_spec_suspend > 0
+        for: 60s
+        labels:
+          severity: critical
+      - alert: CronJobNotRunInThreshold
+        annotations:
+          message: CronJob {{ '{{' }} $labels.namespace {{ '}}' }}/{{ '{{' }} $labels.label_cronjob_name {{ '}}' }} has not started a Job in
+        expr: |
+          (time() - (max( kube_job_status_start_time * ON(job) GROUP_RIGHT() kube_job_labels{label_monitoring_key="middleware"} ) BY (job, label_cronjob_name) == ON(label_cronjob_name) GROUP_LEFT() max( kube_job_status_start_time * ON(job) GROUP_RIGHT() kube_job_labels{label_monitoring_key="middleware"} ) BY (label_cronjob_name))) > 60*60*25
+      {% for cronjob in expected_cronjobs %}
```
david-martin: @aidenkeating I'm not sure about this approach for building up the cronjob alerts.
If you take the following scenario:

* install runs, but the fuse online cronjob wasn't installed for some reason (error swallowed, or the cronjob was accidentally removed from the installer)
* alerts are added for all cronjobs found, which doesn't include fuse online.
* we don't get any alerts about the missing cronjob

I'd prefer if we were explicit about the jobs that should exist.
This also makes it easier & more noticeable for upgrades if a new cronjob was added i.e. this file will be updated and show up in a diff.

aidenkeating: @david-martin Sure, makes sense. I was coming from the other point-of-view, that we could end up with an unmaintained list of CronJob names and when new CronJobs were added (especially in operators) that they wouldn't be added to a list in the installer.

I'll update this to a static list.

roles/backup/templates/backup-monitoring-alerts.yml.j2
```diff
@@ -0,0 +1,40 @@
...
+  name: backup-monitoring-alerts
+spec:
+  groups:
+    - name: general.rules
+      rules:
+      - alert: JobRunningTimeExceeded
+        annotations:
+          message: Job {{ '{{' }} $labels.namespace {{ '}}' }}/{{ '{{' }} $labels.job {{ '}}' }} has been running for longer than 5 minutes
+        expr: |
+          time() - (max(kube_job_status_active * ON(job) GROUP_RIGHT() kube_job_labels{label_monitoring_key="middleware"}) BY (job) * ON(job) GROUP_RIGHT() max(kube_job_status_start_time * ON(job) GROUP_RIGHT() kube_job_labels{label_monitoring_key="middleware"}) BY (job, namespace, label_cronjob_name) > 0) > 300
```
david-martin: @aidenkeating 
Having this generic alert is OK for now, however I'd see this getting broken out into individual alert checks depending on the backup type.
I'd like to ensure the threshold is reasonable though for now based on real life backup times. 5 minutes seems quite short. Maybe @pat-cremin or @Nanyte25 could help with a reasonable backup threshold that should cause an alert if broken?

roles/backup/templates/backup-monitoring-alerts.yml.j2
```diff
@@ -0,0 +1,40 @@
...
+        annotations:
+          message: CronJob {{ '{{' }} $labels.namespace {{ '}}' }}/{{ '{{' }} $labels.cronjob {{ '}}' }} is suspended
+        expr: |
+          kube_cronjob_labels{ label_monitoring_key="middleware" } * ON (cronjob) GROUP_RIGHT() kube_cronjob_spec_suspend > 0
+        for: 60s
+        labels:
+          severity: critical
+      - alert: CronJobNotRunInThreshold
+        annotations:
+          message: CronJob {{ '{{' }} $labels.namespace {{ '}}' }}/{{ '{{' }} $labels.label_cronjob_name {{ '}}' }} has not started a Job in
```
david-martin: The threshold is missing from the end of the message here `has not started a Job in`.
Is it meant to be 25 hours? `60*60*25`
Are jobs going to be run once daily?
cc @ciaran-byrne @pat-cremin for thoughs on this

aidenkeating: @david-martin Ah yep, thanks. The default backup schedule at the moment in the installer is `0 0 * * *`, so every midnight(?) at the moment

roles/backup/templates/backup-monitoring-alerts.yml.j2
david-martin: > when new CronJobs were added (especially in operators) that they wouldn't be added to a list

This is a good point. I'd like to see how that plays out with an operator. In those cases do we trust the operator takes care of backups given the right aws credentials, or maybe the operator also includes a cronjob alert check in a PrometheusRule in that service's namespace.

@pb82 @philbrookes  Is this something you've thought about for the keycloak operator?

roles/backup/templates/backup-monitoring-alerts.yml.j2
david-martin: So 2 things I'd like to get @ciaran-byrne @pat-cremin 's input on (not directly related to this PR)

* Is once a day reasonable to backup the things?
* Is `0 0 * * *` a reasonable time to run cronjobs? I recall we used to run jobs at 2 or 3am to avoid unusual behaviour around DST changes e.g. no job running, or 2 jobs running in a 24 hr period.

roles/backup/templates/backup-monitoring-alerts.yml.j2
pb82: This would currently not work for the keycloak backup job because it does not label it's CronJob with `monitoring-key: middleware`: https://github.com/integr8ly/keycloak-operator/blob/master/pkg/keycloak/phaseHandler.go#L223

ping @philbrookes

aidenkeating: @pb82 I am working on this change now (to add labels to the CronJob and the Jobs it creates), as it's needed for the other general checks anyway.

pat-cremin: > @aidenkeating I'm not sure about this approach for building up the cronjob alerts.
> If you take the following scenario:
> 
> * install runs, but the fuse online cronjob wasn't installed for some reason (error swallowed, or the cronjob was accidentally removed from the installer)
> * alerts are added for all cronjobs found, which doesn't include fuse online.
> * we don't get any alerts about the missing cronjob
> 
> I'd prefer if we were explicit about the jobs that should exist.
> This also makes it easier & more noticeable for upgrades if a new cronjob was added i.e. this file will be updated and show up in a diff.

If you are scraping the known namespaces and creating cronjobs based on this it stands to reason that you will have an alert to ensure all alerts have been created based off the namespaces themselves, rather than the cronjobs in existence. 

Could it be done, easily, that the install can specify the number of cronjobs expected if backups are enabled and exposing that in a label or similar? That way, for a given namespace with a given label X alerts and cronjobs should exist. An alert can be created to then ensure this number matches the info scraped from the namespaces.

roles/backup/templates/backup-monitoring-alerts.yml.j2
pat-cremin: Real life backup times normally depend on dataset size. It is not uncommon, if encryption, compressions and upload are involved, for backups to take hours. This is usually customer data though(very large mongo instances). System data tends to remain on a smaller scale and will take a few minutes to complete in their entirety. Without knowing how big they will get, it's just guess work, but 5 minutes initially seems like it would be worth warning about.

roles/backup/templates/backup-monitoring-alerts.yml.j2
pat-cremin: > So 2 things I'd like to get @ciaran-byrne @pat-cremin 's input on (not directly related to this PR)
> 
> * Is once a day reasonable to backup the things?
Yes, I would see this a reasonable. Per-transactional backups are something we can consider later, but, every 24 hours as a standard seems fine.
 
> * Is `0 0 * * *` a reasonable time to run cronjobs? I recall we used to run jobs at 2 or 3am to avoid unusual behaviour around DST changes e.g. no job running, or 2 jobs running in a 24 hr period.

2.30am is what we use currently in 3.x. Twofold reasoning: 

1. Most importantly, DST changes as you alluded to.
2. Customers themselves having a tendency to schedule workloads for "midnight", causing competing workloads.

roles/backup/templates/backup-monitoring-alerts.yml.j2
aidenkeating: @pat-cremin From what I understand, the question is whether we can have a mapping between namespace names and the amount of CronJobs we expect to be in those namespaces (via a label)? For example:

The only concern I have with that is that when the alert kicks off we won't then know what CronJob isn't there. A person seeing this alert would know that a CronJob is missing in a namespace. They'd have to check the CronJobs in that namespace and have some documentation about what CronJobs should exist there (I don't know if we have that).

IMO, if we're going to have a static list we may as well specify the names of each CronJob expected and not the count so that when an alert fires a person knows that the MySQL CronJob for SSO is missing without doing any investigation and can restore it using some SOP or something.

As for whether it gets stored as a label or just a var in the installer, I don't have any preference.

Could be missing something in your comment though? Any thoughts?

roles/backup/templates/backup-monitoring-alerts.yml.j2
aidenkeating: I will update the default backup schedule to 2.30am each day, thanks for clarifying this

roles/backup/templates/backup-monitoring-alerts.yml.j2
pat-cremin: @aidenkeating Your suggestion works just as well. The label or var could allow the inherent list through standardisation of the cron names. eg. 3scale mysql with backup enabled would expect a "3scale-mysql-backup-cron", with label on the component to state it will need a backup, the logic to  create the correct number of crons and know the exact names of them follows organically.

roles/backup/templates/backup-monitoring-alerts.yml.j2
pat-cremin: @aidenkeating  If it's not clear, what I think should be done here is have a warning after 5 minutes and a critical after 10. Currently it is set to critical after 5.

aidenkeating: @pat-cremin I have a limited knowledge of prometheus. Just to confirm, is there a way to achieve this with one alert, or would this be two separate alerts with different thresholds (`300s` for warning and `600s` for critical)?

pat-cremin: As far as I am aware you can't nest alert levels within the same rule(I will double check this). You can create two rules, both with the same name and change the criteria for triggering.

pat-cremin: So, you do need two rules and then an inhibit rule to prevent both rules triggering at the same time. eg.

```
inhibit_rules:
  - source_match:
      severity: 'critical'
    target_match:
      severity: 'warning'
    equal: ['alertname']
```

aidenkeating: Thanks for clarifying this, testing this out now

aidenkeating: This worked as expected, thanks for looking into this, to apply it properly (not manually) the monitoring operator will need to be updated. I can work on this tomorrow.

aidenkeating:
@david-martin I've updated based on your comments, mind taking another look.

There is a PR in progress (https://github.com/integr8ly/application-monitoring-operator/pull/26) to inhibit duplicate alert names and one merged to handle the Keycloak CronJob labelling (https://github.com/integr8ly/keycloak-operator/pull/40). 

I'll include the bumps for these two operators in this PR also.

roles/backup/templates/backup-monitoring-alerts.yml.j2
david-martin: @aidenkeating given the new approach to cronjob alerts, I think it's important to document how this works.
I'd like to see at least a followup issue for documenting what's involved in adding a new CronJob for some component, including alerts for that cronjob.

aidenkeating: @david-martin I have created a follow-up ticket, INTLY-1330, to address this documentation and assigned it to myself.

aidenkeating:
@odra Mind taking another look at this, there have been a number of updates but the functionality of the PR/verification should be identical.

aidenkeating:
@david-martin @pb82 @matskiv I have run the install on my cluster `akeating-5136`.

pb82:
:eyes:

david-martin:
also :eyes:

aidenkeating:
Looks like a few alerts have gone off for the SSO jobs due to having overridden the image tag (for another PR). This should resolve itself in a few minutes.

aidenkeating:
@pb82 Thanks for verifying.

They don't, they're as a result of another CronJob I was trying out. I have removed the broken CronJob so those alerts should be removed in a couple of minutes. I'll confirm this.

roles/backup/templates/backup-monitoring-alerts.yml.j2
```diff
@@ -0,0 +1,45 @@
...
+metadata:
+  labels:
+    monitoring-key: middleware
+  name: backup-monitoring-alerts
+spec:
+  groups:
+    - name: general.rules
+      rules:
+      {% for check in [{ 'time': '300', 'severity': 'warning' }, { 'time': '600', 'severity': 'critical' }] %}
+      - alert: JobRunningTimeExceeded
```
david-martin: @aidenkeating If we were to inhibit the warning message when the critical is triggering, would this inhibit *all* warnings for all other Jobs?

aidenkeating: @david-martin It'll inhibit alerts with the same name. So if one JobRunningTimeExceeded is critical and the others are warning, the alert will show for the critical. Once the others become critical they will show.

I can look into separating this out. Either by making a `JobRunningTimeExceeded_<cronjob-name>` alert or by seeing if this can be handled through config in AlertManager.

I'll take a look at the AlertManager config first.

aidenkeating:
@david-martin Looks like I dismissed your reviews on a rebase

roles/backup/templates/backup-monitoring-alerts.yml.j2
aidenkeating: @david-martin I've tested out some config changes to AlertManager.

We can have it so that, if there are three Jobs causing a `JobRunningTimeExceeded` alert. Two of them have both `warning` and `critical` firing, one has only `warning`. We'll see this:

Screen Shot 2019-03-20 at 14 15 46

If that's the type of functionality we want, I can make that in the `application-monitoring-operator.`

aidenkeating: PR to address this: https://github.com/integr8ly/application-monitoring-operator/pull/28

PR #454, opened 2019-03-12
[WIP] moving service monitoring to integreatly installer

Nanyte25:
## Additional Information
PR is related to this jira https://issues.jboss.org/browse/INTLY-661

## Verification Steps
This service-monitor will need to be installed on an OSD cluster.

-->

### DIFF
--- a/roles/webapp/tasks/monitoring.yml
+++ b/roles/webapp/tasks/monitoring.yml
@@ -1,8 +1,8 @@
 ---
 - name: Create template for Service Monitor
-  tempfile:
-   src: servicemonitor.yaml.j2
-   dest: /tmp/servicemonitor.yaml.j2
+  template:
+   src: servicemonitor.yaml
+   dest: /tmp/servicemonitor.yaml
 
 - name: Create Service Monitor resource
   shell: oc apply -f /tmp/servicemonitor.yaml -n {{ webapp_namespace }}
--- a/roles/webapp/tasks/monitoring.yml
+++ b/roles/webapp/tasks/monitoring.yml
@@ -1,0 +1,9 @@
+---
+- name: Create template for Service Monitor
+  tempfile:
+   src: servicemonitor.yaml.j2
+   dest: /tmp/servicemonitor.yaml.j2
+
+- name: Create Service Monitor resource
+  shell: oc apply -f /tmp/servicemonitor.yaml -n {{ webapp_namespace }}
+  
--- a/roles/webapp/tasks/monitoring.yml
+++ b/roles/webapp/tasks/monitoring.yml
@@ -1,7 +1,7 @@
 ---
 - name: Create template for Service Monitor
   template:
-   src: servicemonitor.yaml
+   src: servicemonitor.yaml.j2
    dest: /tmp/servicemonitor.yaml
 
 - name: Create Service Monitor resource
--- a/roles/webapp/templates/servicemonitor.yaml.j2
+++ b/roles/webapp/templates/servicemonitor.yaml.j2
@@ -12,6 +12,6 @@ spec:
     scheme: http
     tlsConfig:
       insecureSkipVerify: true
-selector:
+ selector:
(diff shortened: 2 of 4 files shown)


roles/webapp/templates/servicemonitor.yaml.j2
```diff
@@ -0,0 +1,17 @@
+- apiVersion: monitoring.coreos.com/v1
```
odra: the file needs to contain only the service resource, using a "-" will make it an array and the oc command will fail

odra:
You need to additional files/changes as well: 

- `roles/webapp/tasks/monitoring.yml`: a new ansible task that will process the template and then apply using the oc cli
- include the new task in this playbook: https://github.com/integr8ly/installation/blob/master/playbooks/install_middleware_monitoring.yml#L12

roles/webapp/templates/servicemonitor.yaml.j2
Nanyte25: @odra I was just going to add the install task to the provsion-webapp-yml file instead.

odra: It needs to be added in another file and loaded from in the middleware monitoring playbook because those monitoring crds will not be ready/applied when the webapp is being installed

roles/webapp/tasks/monitoring.yml
```diff
@@ -0,0 +1,9 @@
+---
+- name: Create template for Service Monitor
+  tempfile:
```
odra: should be "template" instead of "tempfile"

roles/webapp/tasks/monitoring.yml
```diff
@@ -0,0 +1,9 @@
...
+   src: servicemonitor.yaml.j2
+   dest: /tmp/servicemonitor.yaml.j2
```
odra: change to `/tmp/servicemonitor.yaml` (since it's not a template file anymore)

roles/webapp/templates/servicemonitor.yaml.j2
```diff
@@ -0,0 +1,17 @@
...
+  spec:
+   endpoints:
+    - path: /metrics
+      port: http
+      scheme: http
+      tlsConfig:
+        insecureSkipVerify: true
+  selector:
+    matchLabels:
+      app: tutorial-web-app
```
odra: The file needs to back one indentation level from line 3 until the end of the file:
```
apiVersion: monitoring.coreos.com/v1
kind: ServiceMonitor
metadata:
 labels:
   monitoring-key: middleware
 name: tutorial-webapp-monitoring
 namespace: "{{ webapp_namespace }}"  
spec:
  endpoints:
    - path: /metrics
      port: http
      scheme: http
      tlsConfig:
        insecureSkipVerify: true
  selector:
    matchLabels:
      app: tutorial-web-app
```

roles/webapp/tasks/monitoring.yml
Nanyte25: change files name to load in /tmp

roles/webapp/templates/servicemonitor.yaml.j2
Nanyte25: @odra Can you check the indentation and see if it correct now.

roles/webapp/templates/servicemonitor.yaml.j2
```diff
@@ -0,0 +1,17 @@
...
+ name: tutorial-webapp-monitoring
+ namespace: webapp
+spec:
+ endpoints:
+  - path: /metrics
+    port: http
+    scheme: http
+    tlsConfig:
+      insecureSkipVerify: true
+selector:
```
odra: selector needs to be within the "spec" prop

Nanyte25: @odra let me know when your done testing, Thanks

PR #456, opened 2019-03-14
[WIP] backup-container update

fh-automatron:


inventories/group_vars/all/manifest.yaml
```diff
@@ -92,8 +92,8 @@ msbroker_release_tag: 'v0.0.1'
 msbroker_template: 'https://raw.githubusercontent.com/integr8ly/managed-service-broker/{{ msbroker_release_tag }}/templates/broker.template.yaml'
 
 # information about backups
-backup_version: master
-backup_image_tag: master
+backup_version: '99.99.99'
+backup_image_tag: '99.99.99'
```
mikenairn: @JameelB Might be a good idea to make these different values when testing to be sure the correct variable is updated as expected.

JameelB: After talking with @odra and @pb82, I believe the plan is for these two to have the same value. I think the plan was to have both of these vars to be the same as the release tag.

PR #459, opened 2019-03-15
remove backup image tag var

JameelB:
## Additional Information
The `backup_version` and `backup_image_tag` will always have the same value. There is no reason for there to be two vars. The `backup_version` should be used instead.

Related Discussion - https://github.com/integr8ly/ci-cd/pull/81#discussion_r265737985

## Verification Steps
1. Ensure installer runs successfully and all Integreatly services are installed correctly


pb82:
:eyes:

PR #476, opened 2019-03-21
INTLY-1385 Use exported_namespace for amqonline pod count alert

david-martin:
The `exported_namespace` label should be used on v1.2 as kube_state_metrics is installed separately.
On master (1.3) kube_state_metrics  data is federated from the cluster monitoring, which uses the `namespace` label


pb82:
:eyes:

david-martin:
Thanks @tremes.
Build is having apt issues, unrelated to this change.
Merging

PR #478, opened 2019-03-21
gitea and launcher use http routes

pb82:
Gitea and Launcher use http routes.

Verification steps:

1. install integreatly from this branch
1. Open the console, make sure the links at the top are visible
1. If the links are not visible, check the network tab in the dev tools for scripts that weren't loaded due to cert errors. Copy the urls of those scripts, open them in another tab and accept the certs.
1. Make sure all links (above the service catalog) work.


matskiv:
Installing on my cluster now.

PR #481, opened 2019-03-22
rhsso update 7.2.2.GA

fh-automatron:

### DIFF
--- a/inventories/group_vars/all/manifest.yaml
+++ b/inventories/group_vars/all/manifest.yaml
@@ -57,7 +57,7 @@ launcher_sso_template: https://raw.githubusercontent.com/jboss-container-images/
 
 #not currently used but is the version installed by the operator
 # rhsso is not an optional component
-rhsso_version: '7.2.2'
+rhsso_version: '7.2.2.GA'
 #below vars not currently used but will be used to pull in the new versions of RH-SSO once we decouple the resources from the installer.
 rhsso_imagestream_name: redhat-sso72-openshift:1.4
 rhsso_imagestream_image: registry.access.redhat.com/redhat-sso-7/sso72-openshift:1.4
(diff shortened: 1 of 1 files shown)


inventories/group_vars/all/manifest.yaml
```diff
@@ -57,11 +57,11 @@ launcher_sso_template: https://raw.githubusercontent.com/jboss-container-images/
 
 #not currently used but is the version installed by the operator
 # rhsso is not an optional component
-rhsso_version: '7.2.0.GA'
+rhsso_version: '7.2.2'
```
mikenairn: @philbrookes Does v1.2.2 of the keycloak operator install version 7.2.2 of keycloak? https://github.com/integr8ly/keycloak-operator/blob/master/pkg/keycloak/keycloak.go#L30

philbrookes: it installs `7.2.2.GA`

mikenairn: Is it possible to have this value set to "7.2.2.GA"  in the operator? https://github.com/integr8ly/keycloak-operator/blob/master/pkg/keycloak/keycloak.go#L30 That way this value would get set properly

mikenairn: That's if we actually care about the .GA on the end here. @maleck13 Think i might have asked you about this before.

philbrookes: I don't think the operator actually uses that value anywhere.

philbrookes: it's only used here: https://github.com/integr8ly/keycloak-operator/blob/9cbca1d734aeeebcb8e504ce672b21c34b641261/pkg/keycloak/phaseHandler.go#L55 and that `Status.Version` isn't used anywhere.

mikenairn: rhsso_version isn't used in the installer either, so .... doesn't seem like in either case it matters what the value actually is so we can probably just leave it as 7.2.2 and merge this PR.

maleck13: It is used. https://github.com/integr8ly/installation/blob/master/roles/customisation/tasks/main.yaml#L108

mikenairn: OK. Does it need the .GA on the end for anything? Is that significant for whatever that is doing?

PR #483, opened 2019-03-22
[WIP] Allow an ns_prefix to be specified

aidenkeating:
## Additional Information
We need to allow the `openshift-` prefix to be specified. By default `oc new-project` won't allow this. A `namespace` role has been added which will create a `namespace` and set annotations on it to workaround the `oc new-project` issue.

## Verification Steps
- Run with `-e ns_prefix='openshift-'`

### DIFF
--- a/inventories/group_vars/all/common.yml
+++ b/inventories/group_vars/all/common.yml
@@ -7,7 +7,7 @@ eval_sso_validate_certs: false
 eval_launcher_sso_realm: launcher_realm
 
 eval_apicurito_namespace: apicurito
-eval_che_namespace: codeready
+eval_che_namespace: "{{ns_prefix | default('')}}codeready"
 eval_launcher_namespace: launcher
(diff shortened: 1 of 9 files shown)


aidenkeating:
@maleck13 @matskiv Mind taking a look? Che needs the namespace to exist beforehand as it uses `oc new-project` in it's install script

matskiv:
👀

matskiv:
I think you missed one place:
https://github.com/integr8ly/installation/blob/master/roles/middleware_monitoring_config/files/kube_state_metrics_alerts.yml#L75

Installation on my cluster is in progress, I will report here as soon as it finishes.
Other than this it looks good.

matskiv:
Installation my cluster finished successfully. Related part of log - https://gist.github.com/matskiv/7611d8c031391661d05645d426d0a25a
Namespace name is correct. Display name works. Che opens fine. Links from webapp work.
👍

PR #487, opened 2019-03-22
Allow RHSSO to have ns_prefix in namespace

aidenkeating:
## Verification Steps
Run the installer with `-e ns_prefix='openshift-'` and ensure everything still works, including monitoring checks.

### DIFF
--- a/inventories/group_vars/all/common.yml
+++ b/inventories/group_vars/all/common.yml
@@ -10,7 +10,7 @@ eval_launcher_sso_realm: launcher_realm
 eval_apicurito_namespace: "{{ns_prefix | default('')}}apicurito"
 eval_che_namespace: "{{ns_prefix | default('')}}codeready"
 eval_launcher_namespace: "{{ns_prefix | default('')}}launcher"
-eval_rhsso_namespace: sso
+eval_rhsso_namespace: "{{ns_prefix | default('')}}sso"
 eval_webapp_namespace: "{{ns_prefix | default('')}}webapp"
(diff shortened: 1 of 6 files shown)


roles/namespace/tasks/create.yml
```diff
@@ -8,5 +8,6 @@
   shell: oc create namespace {{ namespace }}
   when: namespace_exists.rc != 0
 
-- name: Give namespace {{ namespace }} display name {{ che_display_name }}
-  shell: oc annotate --overwrite namespace {{ namespace }} openshift.io/display-name="{{ che_display_name }}"
+- name: Give namespace {{ namespace }} display name {{ display_name }}
```
aidenkeating: @matskiv You might need this patch in your PR also. There were some incorrect vars used here.

roles/namespace/tasks/create.yml
```diff
@@ -8,5 +8,6 @@
...
+  shell: oc annotate --overwrite namespace {{ namespace }} openshift.io/display-name="{{ display_name }}"
+  whem: display_name is defined
```
matskiv: typo^ `whem` => `when`

philbrookes:
do you think we should presume the `-`? 
e.g. `-e ns_prefix=openshift` results in `openshift-namespace` and `-e ns_prefix=""` results in `namespace`?

matskiv:
@philbrookes No, I would leave as is - simple and straight forward. Without auto-magical dash :)

maleck13:
**Verification**
Installation completed succesfully

PR #489, opened 2019-03-24
change enmasse to be able to be installed in openshift- ns

maleck13:
## Additional Information
allows a prefix to be added to the default ns name this is to ensure we can hide these ns in openshift dedicated

## Verification Steps

- provision setting the ns_prefix var to ```openshift-```
- ensure can access route and works from webapp
- run uninstall ensuring to use ```-e ns_prefix=openshift-```
- provision with no ns prefix and ensure ns is as normal (ie the default) 
- ensure can still access via webapp

### DIFF
--- a/inventories/group_vars/all/common.yml
+++ b/inventories/group_vars/all/common.yml
@@ -1,4 +1,5 @@
 ---
+ns_prefix: ''
 eval_action: install
 eval_namespace: eval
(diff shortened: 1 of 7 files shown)


aidenkeating:
@maleck13 Spinning up a cluster at the moment. Will try this out when it's ready

aidenkeating:
@maleck13 Giving a another quick run now

PR #498, opened 2019-03-25
Allow ns_prefix to be specified in backups

aidenkeating:
Verification:
- Set the backup schedule in the manifest to something like `*/1 * * * *`
- Run the managed install (`playbooks/managed/install.yml`|`inventories/managed`) with `-e ns_prefix='openshift-'`
- Ensure that any Secrets created in the backups Namespace that include service URLs are using the correct namespace.
- Ensure no Job is failing because of failing to read the secret or failing to communicate with a component


matskiv:
LGTM, I'll test once my cluster is free.

PR #500, opened 2019-03-26
INTLY-1384 Update launcher frontend version with fix

david-martin:
See https://issues.jboss.org/browse/INTLY-1384 for steps to try reproduce


david-martin:
@tremes I've verified this in my own cluster by patching the image being used.
If you want to verify as well, I'll cherrypick to v1.2 branch.

You may need to remove any users from launcher's sso so they can be re-created & re-linked so the issue can be attempted to reproduce (at the time of redirect back from github)

PR #508, opened 2019-03-27
[WIP] Use oc create in fuse managed instead of apply

### LINKED ISSUE #402: Managed Fuse install happens for every run/re-run of the installer

aidenkeating:
Currently, we are using `oc apply` in many places in the Fuse
Managed installation. This causes issues when running the installation
over an existing installation as `oc apply` will update the existing
resources.

This change changes these places to `oc create` with a check that will
not fail if the resource already exists.

This change also includes an update to the Fuse Managed checks to see
if all Pods are up. Instead of referencing a label and specifying an
exact number of Pods, this will now specify a minimum amount of Pods
that should exist in the namespace before continuing.

Verification:
- Run the installer
- Ensure it completes successfully
  - Ensure the console can be accessed by any two users in the
    cluster, to verify the OAuth proxy rules are still correct
- Run the installer again, with the Fuse Managed namespace now
there
- Ensure it completes successfully, again

Resolves #402

### DIFF
--- a/roles/fuse_managed/tasks/main.yml
+++ b/roles/fuse_managed/tasks/main.yml
@@ -13,7 +13,9 @@
   changed_when: namespace_label.rc == 0
 
 - name: Create Syndesis CRD
-  shell: oc apply -f {{ fuse_online_crd_resources }}
+  shell: oc create -f {{ fuse_online_crd_resources }}
+  register: create_crd
+  failed_when: create_crd.stderr != '' and 'AlreadyExists' not in create_crd.stderr
 
 - name: Create Fuse image streams
   shell: "oc replace --force -f {{ fuse_online_imagestream_resources }} -n openshift"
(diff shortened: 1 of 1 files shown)


roles/fuse_managed/tasks/main.yml
```diff
@@ -22,19 +24,23 @@
...
 
 - name: Create Syndesis custom resource in {{ fuse_namespace }}
-  shell: oc apply -f /tmp/syndesis-customresource.yml -n {{ fuse_namespace }}
+  shell: oc create -f /tmp/syndesis-customresource.yml -n {{ fuse_namespace }}
+  register: create_cr
+  failed_when: create_cr.stderr != '' and 'AlreadyExists' not in create_cr.stderr
 
 - name: Verify Fuse deployment succeeded
-  shell: oc get pods -n {{ fuse_namespace }} --selector="app=syndesis" -o jsonpath='{.items[*].status.containerStatuses[?(@.ready==true)].ready}' | wc -w
+  shell: oc get pods -n openshift-fuse -o jsonpath='{.items[*].status.containerStatuses[?(@.ready==true)].ready}' | wc -w | tr -d ' '
```
sedroche: Are you going for `{{ns_prefix}}{{ fuse_namespace }}` here?

aidenkeating: @sedroche `{{ fuse_namespace }}` should be enough here as the `ns_prefix` will be set in `defaults.yml` of this role. I've removed the mistakenly hardcoded namespaces though in 0d56914.

mikenairn:
@aidenkeating Update and re-open if this is still valid.

PR #513, opened 2019-03-27
Rename grafana(_dashboard)_crd.yml.j2 files to match content

matskiv:
## Additional Information
There was content x file name mismatch with this files. 
I just swapped the names.


pb82:
@matskiv so we had the dashboard CRD in grafana and the grafana CRD in the dashboard?

matskiv:
@pb82 yes, exactly.

PR #521, opened 2019-04-01
INTLY-1578 Remove managed install playbook

david-martin:
The backup playbook is now called from the main install playbook.
It is only run if the backup_restore_install flag is true.

*Why*

So we only have a single install playbook that gets run on every environment (poc, pds, osd).
Differences between clusters should be handled by inventory files rather than playbooks.


playbooks/install.yml
```diff
@@ -36,6 +36,13 @@
       shell: "htpasswd -D /etc/origin/master/htpasswd integreatly-admin"
       become: yes
       when: create_cluster_admin
+
+- hosts: managed
```
pmccarthy: Might not be within the remit of this PR but not sure we need a 'managed' host either. Regardless of which environment we run against, there will only ever be a localhost and master.

david-martin: @pmccarthy OK. So the steps to remove it would be:

* use loca as the hostl here
* use local in the managed.template
* use local in the SOP (linked pr)

Does that sound right?

maleck13:
So the flavours of installation here will be controlled by the inventory used to do the install? IE there would be a workshop inventory as well?

david-martin:
> IE there would be a workshop inventory as well

Yes, that's my understanding

@pmccarthy Are you good with these changes?

I've ran the install.yml in a cluster with the hosts.template, and the managed.template to ensure both still work, with the latter also installing backups.

playbooks/install.yml
```diff
@@ -36,6 +36,13 @@
       shell: "htpasswd -D /etc/origin/master/htpasswd integreatly-admin"
       become: yes
       when: create_cluster_admin
+
+- hosts: local
```
pmccarthy: No need to specify the host here as the back_restore playbook that gets imported has a reference to localhost already, see: https://github.com/integr8ly/installation/blob/master/playbooks/backup_restore/install.yml#L2

pmccarthy:
> So the flavours of installation here will be controlled by the inventory used to do the install? IE there would be a workshop inventory as well?

@maleck13 Yes, that's how I would hope it would be handled. They should all use a generic playbooks structure and have unique vars overridden in their respective inventory file

pmccarthy:
Again - may not be needed in this PR but it would also be nice if we removed that backup_restore directory and renamed the `playbooks/backup_restore/install.yml` file to `playbooks/install_backup.yml`. Not sure if this is referenced in any SOP/README though.

david-martin:
> Not sure if this is referenced in any SOP/README though

I'll check and update that

> removed that backup_restore directory

i'll do that

david-martin:
@pmccarthy are you good with the latest changes, and sop update in linked pr?

david-martin:
output from end of install using the managed inventory

```
TASK [include_role : backup] ***************************************************

TASK [backup : template] *******************************************************
ok: [127.0.0.1]

TASK [backup : Create CronJob Alerts] ******************************************
changed: [127.0.0.1]

PLAY [localhost] ***************************************************************

TASK [switch back to system admin] *********************************************
changed: [127.0.0.1]

TASK [debug] *******************************************************************
ok: [127.0.0.1] => {
    "msg": "All services have been provisioned successfully."
}

TASK [debug] *******************************************************************
ok: [127.0.0.1] => {
    "msg": "Please add 'https://launcher-sso-launcher.apps.waterford-8b55.openshiftworkshop.com/auth/realms/launcher_realm/broker/github/endpoint' as the Authorization callback URL of your GitHub OAuth Application."
}

PLAY RECAP *********************************************************************
127.0.0.1                  : ok=486  changed=235  unreachable=0    failed=0
master1.waterford-8b55.internal : ok=22   changed=15   unreachable=0    failed=0

^C[1]+  Done                    nohup ansible-playbook -i inventories/managed.template playbooks/install.yml -e eval_seed_users_count=1

[root@bastion installation]# oc get cronjob --all-namespaces
NAMESPACE                       NAME                        SCHEDULE     SUSPEND   ACTIVE    LAST SCHEDULE   AGE
openshift-integreatly-backups   3scale-mysql-backup         30 2 * * *   False     0         <none>          2h
openshift-integreatly-backups   3scale-postgres-backup      30 2 * * *   False     0         <none>          2h
openshift-integreatly-backups   3scale-redis-backup         30 2 * * *   False     0         <none>          2h
openshift-integreatly-backups   codeready-postgres-backup   30 2 * * *   False     0         <none>          2h
openshift-integreatly-backups   codeready-pv-backup         30 2 * * *   False     0         <none>          2h
openshift-integreatly-backups   enmasse-postgres-backup     30 2 * * *   False     0         <none>          2h
openshift-integreatly-backups   enmasse-pv-backup           30 2 * * *   False     0         <none>          2h
openshift-integreatly-backups   fuse-postgres-backup        30 2 * * *   False     0         <none>          2h
openshift-integreatly-backups   launcher-postgres-backup    30 2 * * *   False     0         <none>          2h
openshift-integreatly-backups   resources-backup            30 2 * * *   False     0         <none>          2h
openshift-logging               logging-curator             30 3 * * *   False     0         11h             5d
sso                             daily-at-midnight           30 2 * * *   False     0         <none>          2h
```

PR #524, opened 2019-04-03
add doc around creating releases

maleck13:
## Additional Information
Doc around creating new releases


maleck13:
@mikenairn @david-martin  FYI

maleck13:
I would prefer if it was in a pipeline but ok to do it locally until that work can be done

PR #525, opened 2019-04-03
INTLY-1260 Add a var for *not* running master tasks

david-martin:
This change allows for running an install without the need to
ssh to the master(s).
In this scenario, certain extra vars should be set, and other tasks
done outside of the install run.

## Verification Steps

To verify on PDS (can ssh to master during install), run an install as normal. Everything should behave as it did before. As should uninstall.

To verify on OSD (*no* ssh to master during install) do the following:

- run the install.yml playbook:
  - based on the managed.template inventory file
  - on a machine that is logged in (oc) with cluster-admin access (*not* system:admin user)
  - with values set for the following vars
    - eval_app_host (router subdomain, used for creating some routes & for 3scale)
    - openshift_master_url (master url for oauth use)
- after install, add the relevant cors rule to the master-config.yaml based on the webapp route
  - Get the webapp host using `oc get route tutorial-web-app -n webapp --template 'https://{{.spec.host}} '`
  - Add the following to master-config.yaml
```
   corsAllowedOrigins:
   - <webapp_host>
```
- add the cluster SSO identity provider to the master-config.yaml
  - Get the client secret using `oc get secret openshift-client-client -n sso --template '{{.data.install}}' | base64 --decode`
  - Get the sso host using `oc get route sso -n sso --template 'https://{{.spec.host}} '`
  - Add the following to master-config.yaml
```
   identityProviders:
   - name: rh_sso
     challenge: true
     login: true
     mappingInfo: add
     provider:
       apiVersion: v1
       kind: OpenIDIdentityProvider
       clientID: openshift-client
       clientSecret: <client_secret>
       ca: ca.crt
       urls:
         authorize: <sso_host>/auth/realms/openshift/protocol/openid-connect/auth
         token: <sso_host>/auth/realms/openshift/protocol/openid-connect/token
         userInfo: <sso_host>/auth/realms/openshift/protocol/openid-connect/userinfo
         claims:
         id:
           - sub
         preferredUsername:
           - preferred_username
          name:
           - name
         email:
           - email
```
- restart the master(s)
- you should be able to login as normal to openshift & the various services. The webapp should behave as normal (i.e. provisions things for your user by calling the openshift api from browser)


david-martin:
@pmccarthy @jessesarn Can I get your eyes on this change?

jessesarn:
I'm going to go through this with the integreatly-dev OSD cluster which should already have the changes made to the master config from previous install.  I'll let you know how it goes.

jessesarn:
Worth noting that I don't expect that logging in with rhsso provider is going to work properly at the end because the client secret will have changed; however, I should be able to still use github login to test everything and can ask OpenShift SRE to update the config so that config loop has the new secret.

jessesarn:
I had to make sure I pulled off all the the secret values that I needed to be able to run the install again with the right values for the aws s3 details. (aws access keys, bucket name, etc.) before running my slightly tweaked uninstall playbook from v1.1.0.

The first run of the install playbook stalled out at:
TASK [code-ready : install code ready valid certs]

I couldn't find any real explanation for why and gave it well over 30 mins before I gave up and killed the process.  I removed the openshift-codeready namespace and tried the install again which got past the point where it hung previously but failed at:
```
TASK [3scale : Create 3Scale SSO provider]
...
    "msg": "Status code was 503 and not [201, 422]: HTTP Error 503: Service Unavailable", 
    "pragma": "no-cache", 
    "redirected": false, 
    "status": 503, 
    "url": "https://3scale-admin.6c7f.integreatly-dev.openshiftapps.com/admin/api/account/authentication_providers.xml"
```
Trying to access that url for the authentication_providers.xml myself returned the same 503; however, when I tried the url without the path portion, I got forwarded to what looked like a working login page (https://3scale-admin.6c7f.integreatly-dev.openshiftapps.com/p/login).  I'll continue my investigation in the morning.

jessesarn:
Sorry for the long delay in getting the results I shared last week into a PR comment.  I was able to get through the steps using the new parameters as well as with ns_prefix='openshift-'.  It took 7 tries to finally get fully through the install playbook.  I saved the logs from each of the failures which I renamed to indicate what went wrong:
```
$ du -sh failure*
564K	failure1_codeready-install-certs-hang.log
592K	failure2_3scale-sso-create.log
280K	failure3_fuse-imagestream-conn-reset.log
1.2M	failure4_customisation-wrong-yq.log
412K	failure5_launcher-create-public-client.log
1.1M	failure6_backup-secret-missing.log
```
Failure #4 was because I had a completely different implementation of yq than the one that was being used.
Failure #6 is a chicken in the egg problem where the secret name is expected but the namespace where it needs to be won't exist until after the install is complete.  Workaround is to manually create the namespace (with openshift- prefix) and create the secret before running the install. [INTLY-1683](https://issues.jboss.org/browse/INTLY-1683)

All of the other failures were due to random transient issues that the installer didn't handle better with a retry/timeout, but not related to the changes from this PR.  Therefore, I feel good that this can be merged.

PR #527, opened 2019-04-04
add sso route to msb params

pb82:
Update to latest msbroker which adds a service instance for rhsso

### DIFF
--- a/roles/msbroker/tasks/main.yml
+++ b/roles/msbroker/tasks/main.yml
@@ -38,6 +38,10 @@
 - set_fact:
     fuse_dashboard_url: "https://{{ fuse_host_cmd.stdout}}"
 
+- name: Get SSO route
+  shell: oc get route sso -o template --template \{\{.spec.host\}\} -n {{ eval_rhsso_namespace }}
+  register: sso_route
+
 # Create managed-service-broker namespace
 - include_role:
     name: namespace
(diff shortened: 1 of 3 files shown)


pb82:
@maleck13 it's deployed to my cluster: https://master.pbraun-73b1.openshiftworkshop.com/console/project/webapp/overview

When you open the webapp you should see the new RH-SSO route on the overview page. @david-martin has already verified that part.

roles/msbroker/tasks/main.yml
```diff
@@ -73,6 +77,7 @@
   --param=THREESCALE_DASHBOARD_URL={{ threescale_dashboard_url }} \
   --param=APICURIO_DASHBOARD_URL={{ apicurito_dashboard_url }} \
   --param=SHARED_FUSE_DASHBOARD_URL={{ fuse_dashboard_url }} \
+  --param=SSO_URL=https://{{ sso_route.stdout }}/auth/admin/{{ sso_realm }}/console \
```
david-martin: @pb82 This should be the `openshift` realm. Is that the case here?

pb82: Yes, this is populated from https://github.com/integr8ly/installation/blob/master/roles/rhsso/defaults/main.yml#L6 through https://github.com/integr8ly/installation/pull/527/files#diff-4ea592a838e3c8063aa17dbe4ed16379R172

maleck13:
Verified the url populated in MSB was https://sso-sso.apps.pbraun-73b1.openshiftworkshop.com/auth/admin/openshift/console

PR #529, opened 2019-04-04
feat: add user-facing sso

wei-lee:
## Additional Information

https://issues.jboss.org/browse/INTLY-1447

## Verification Steps

1. Run the install playbook. You should see there is a new `user-sso` namespace created, and the keycloak operator is running there.
2. Run the uninstall playbook and the namespace should be removed.

### DIFF
--- a/inventories/group_vars/all/manifest.yaml
+++ b/inventories/group_vars/all/manifest.yaml
@@ -61,7 +61,7 @@ rhsso_version: '7.2.6.GA'
 #below vars not currently used but will be used to pull in the new versions of RH-SSO once we decouple the resources from the installer.
 rhsso_imagestream_name: redhat-sso72-openshift:1.4
 rhsso_imagestream_image: registry.access.redhat.com/redhat-sso-7/sso72-openshift:1.4
-rhsso_operator_release_tag: 'v1.3.0'
+rhsso_operator_release_tag: 'v1.3.2'
 rhsso_operator_resources: 'https://raw.githubusercontent.com/integr8ly/keycloak-operator/{{rhsso_operator_release_tag}}/deploy/'
 rhsso_namespace: 'sso'
 
@@ -86,6 +86,9 @@ apicurito_version: '0.2.18.Final'
 #controls whether nexus is installed or not
 nexus: true
 nexus_version: '2.14.11-01'
+
+#controls whether user-rhsso is installed or not
+user_rhsso: true
 
 monitoring_label_name: 'monitoring-key'
 monitoring_label_value: 'middleware'
--- a/inventories/group_vars/all/manifest.yaml
+++ b/inventories/group_vars/all/manifest.yaml
@@ -87,9 +87,6 @@ apicurito_version: '0.2.18.Final'
 nexus: true
 nexus_version: '2.14.11-01'
 
-#controls whether user-rhsso is installed or not
-user_rhsso: true
-
 monitoring_label_name: 'monitoring-key'
 monitoring_label_value: 'middleware'
 middleware_monitoring_operator_release_tag: '0.0.5'
--- a/playbooks/upgrades/install_user_rhsso.yml
+++ b/playbooks/upgrades/install_user_rhsso.yml
@@ -13,7 +13,7 @@
         sso_namespace: "{{ eval_user_rhsso_namespace }}"
         sso_namespace_display_name: "User Facing Red Hat Single Sign-On"
       tags: ['user_rhsso']
-      when: user_rhsso
+      when: user_rhsso | default(true) | bool
     -
       name: Setup backup for user rhsso
       include_role:
@@ -22,4 +22,4 @@
       vars:
         sso_namespace: "{{ eval_user_rhsso_namespace }}"
       tags: ['user_rhsso']
-      when: user_rhsso
+      when: user_rhsso | default(true) | bool
--- a/playbooks/upgrades/install_user_rhsso.yml
+++ b/playbooks/upgrades/install_user_rhsso.yml
@@ -1,0 +1,25 @@
+---
+- hosts: localhost
+  gather_facts: true
+  tasks:
+    - name: Include vars from rhsso
+      include_vars: "../../roles/rhsso/defaults/main.yml"
+    -
+      name: Install user rhsso
+      include_role:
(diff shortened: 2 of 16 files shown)


wei-lee:
@david-martin I certainly can. The only problem is that there is already a `uninstall.yml` tasks in the current rhsso role. I need to split it because there are some bit that I don't need to uninstall the user rhsso. Do you want me to leave that one there?

wei-lee:
@david-martin would it be ok if I rename the tasks to`install_sso` and `uninstall_sso`?

david-martin:
> would it be ok if I rename the tasks to`install_sso` and `uninstall_sso`

@wei-lee  Sure, that sounds fine.

david-martin:
@wei-lee 
Install looks to put things in the right place

```
oc get keycloak --all-namespaces
NAMESPACE   NAME      AGE
sso         rhsso     38m
user-sso    rhsso     12m

oc get deployment,po -n user-sso
NAME                                      DESIRED   CURRENT   UP-TO-DATE   AVAILABLE   AGE
deployment.extensions/keycloak-operator   1         1         1            1           25m

NAME                                     READY     STATUS    RESTARTS   AGE
pod/keycloak-operator-857c5dcdc9-6pmxz   1/1       Running   0          25m
```

However, there is no keycloak/sso instance running.
I believe this is because there is no keycloakrealm.

```
oc get keycloakrealm --all-namespaces
NAMESPACE   NAME        AGE
sso         openshift   38m
```

Would it make sense to have a `master` realm?
Or at least whatever is needed to have the keycloak instance running at this point.

@philbrookes Interested in your thoughts here on this behaviour.

wei-lee:
@david-martin Yeah, my understanding is that once a sso instance is running, the master realm will be created automatically. You don't need to create a keycloak realm for master. Given that user may or may not using this service, it might be ok to only provision the instance when a realm is actually needed. That said, I am ok with creating a realm here too. Just need to know what realm to actually create.

philbrookes:
yeah, the operator only creates the sso instances when a request for a realm is created.

philbrookes:
it's also worth noting that the rhsso templates use deploymentconfigs rather than deploys.

playbooks/upgrades/install_user_rhsso.yml
```diff
@@ -0,0 +1,16 @@
+---
```
wei-lee: @david-martin @maleck13 Should I use a version number for the upgrade playbook?

wei-lee:
@david-martin @pb82 @maleck13  This PR is ready for final review now. I have added tasks to perform backup, and alert rules for the backup job. I have also added a upgrade playbook to install it on an existing cluster.

Once you have run the installation, you will see a new namespace called "user-sso", and it will have keycloak running, and you should be able to login with the admin user credential.

The monitoring is setup for the new keycloak instance: the dashboard will have dropdown and you can select namespace from there.

Can I get some help on reviewing this PR and get it merged? Thanks.

pb82:
:eyes:

playbooks/upgrades/install_user_rhsso.yml
aidenkeating: @wei-lee @david-martin @maleck13 Did we get an offline answer to this? I was working on the 1.4 upgrade playbook and was planning to have it something like this: `playbooks/upgrades/1_4_0_upgrade.yml`.

wei-lee:
@david-martin I can't merge this PR because you have requested changes. Do you mind do a quick review and approve the PR if it's ok?

wei-lee:
@pb82 maybe, not sure there is too much benefit if we do it that way.

roles/rhsso/tasks/uninstall_sso.yml
```diff
@@ -0,0 +1,37 @@
+---
```
grdryn: For my own understanding, would the Ansible [k8s module](https://docs.ansible.com/ansible/latest/modules/k8s_module.html) have worked here? Does it work with custom resources? Are there (dis/)advantages to using `oc` over the module?

I know this is pretty much just copied from what was previously in `roles/rhsso/tasks/uninstall.yml` above -- like I said, it's just for my own understanding/curiosity :)

playbooks/upgrades/install_user_rhsso.yml
wei-lee: @aidenkeating no, I didn't get answer about this. Do you want me to leave this as it is, and you can change it afterwards, or do you want to me change it here?

aidenkeating: @wei-lee I'm happy to leave it the way it is for now, I'll change it in my PR if needed

roles/rhsso/tasks/uninstall_sso.yml
wei-lee: @grdryn yes, it will work. I believe it works with custom resource as well because you are just writing YAML in the ansible format with this module.

wei-lee:
@david-martin I have updated this PR to address the problem with where the feature flag should be defined. Can you take a look and approve this PR if that's ok?

david-martin:
@wei-lee I'll take another look at this

inventories/group_vars/all/manifest.yaml
```diff
@@ -61,7 +61,7 @@ rhsso_version: '7.2.6.GA'
 #below vars not currently used but will be used to pull in the new versions of RH-SSO once we decouple the resources from the installer.
 rhsso_imagestream_name: redhat-sso72-openshift:1.4
 rhsso_imagestream_image: registry.access.redhat.com/redhat-sso-7/sso72-openshift:1.4
-rhsso_operator_release_tag: 'v1.3.0'
+rhsso_operator_release_tag: 'v1.3.2'
```
david-martin: @wei-lee v1.3.2 points to a draft release of the keycloak operator. Is this correct? https://github.com/integr8ly/keycloak-operator/releases

The latest release is now 1.3.4. (cc @pb82 @philbrookes )

wei-lee: @david-martin there is another PR: https://github.com/integr8ly/installation/pull/549/files that will update the KC operator to the latest version. So I think it's ok to leave this one here?

PR #541, opened 2019-04-10
INTLY-1707 partial fix

tremes:
This is just partial fix of the problem, because it doesn't remove the integreatly-admin (if he/she exists) user during uninstallation.

### DIFF
--- a/playbooks/install.yml
+++ b/playbooks/install.yml
@@ -55,7 +55,7 @@
     - name: remove integreatly-admin user
       shell: "htpasswd -D /etc/origin/master/htpasswd integreatly-admin"
       become: yes
-      when: (run_master_tasks | default(true) | bool) and create_cluster_admin
+      when: (run_master_tasks | default(true) | bool) and (create_cluster_admin == true)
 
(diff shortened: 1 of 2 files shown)


trepel:
@tremes I guess it has to be fixed here as well:
https://github.com/integr8ly/installation/blob/master/playbooks/install.yml#L58 ?

tremes:
@trepel yes probably.

trepel:
@maleck13 you approved this but I don't understand, why to remove `integreatly-admin` user if the `create_cluster_admin` is set to `true`. We should remove this user when the `create_cluster_admin` is set to `false`, should not we?

david-martin:
> why to remove `integreatly-admin` user if the `create_cluster_admin` is set to `true`

@trepel My understanding of this is the integreatly-admin user is created at the beginning of the play, and removed near the end of the play. They are created & used for the install if 'create_cluster_admin' is true, and cleaned up after. I don't think 'create_cluster_admin' is meant to create a persistent user

trepel:
@david-martin ah, thanks. It makes sense now. Merging this one then.

PR #542, opened 2019-04-10
Install Prom+Grafana on application-metrics namespace

aliok:
~~Pushing for early feedback. Please don't mind the commit messages, things will be squashed.~~

~~@ziccardi, can you add some info? (I need to leave now)~~

Ready for verification.

### DIFF
--- a/roles/middleware_monitoring/templates/application_monitoring_cr.yml.j2
+++ b/roles/middleware_monitoring/templates/application_monitoring_cr.yml.j2
@@ -1,0 +1,6 @@
+apiVersion: applicationmonitoring.integreatly.org/v1alpha1
+kind: ApplicationMonitoring
+metadata:
+  name: example-applicationmonitoring{{ namespace_postfix }}
+spec:
+  labelSelector: {{ monitoring_label_value }}
--- a/roles/middleware_monitoring/templates/prometheus_operator_cluster_role_binding.yml.j2
+++ b/roles/middleware_monitoring/templates/prometheus_operator_cluster_role_binding.yml.j2
@@ -1,7 +1,7 @@
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
 metadata:
-  name: prometheus-application-monitoring-operator
+  name: prometheus-application-monitoring-operator{{ namespace_postfix }}
 roleRef:
   apiGroup: rbac.authorization.k8s.io
   kind: ClusterRole
--- a/playbooks/install.yml
+++ b/playbooks/install.yml
@@ -45,6 +45,7 @@
 - import_playbook: "./install_services.yml"
 - import_playbook: "./install_webapp.yml"
 - import_playbook: "./install_middleware_monitoring.yml"
+- import_playbook: "./install_application_metrics.yml"
 - import_playbook: "./install_user_rhsso.yml"
 - import_playbook: "./generate-customisation-inventory.yml"
 - import_playbook: "./customise_web_console_install.yml"
--- a/playbooks/install_application_metrics.yml
+++ b/playbooks/install_application_metrics.yml
@@ -1,0 +1,6 @@
+---
+- hosts: localhost
+  gather_facts: no
+  roles:
+  - role: application_monitoring
+    when: application_metrics | default(true) | bool
--- a/playbooks/uninstall.yml
+++ b/playbooks/uninstall.yml
@@ -92,6 +92,13 @@
       tags: ['middleware-monitoring']
       when: middleware_monitoring | default(true) | bool
     -
+      name: Uninstall application-monitoring
+      include_role:
+        name: application_monitoring
+        tasks_from: uninstall
+      tags: ['application-monitoring']
+      when: application_monitoring | default(true) | bool
+    -
       name: Uninstall managed Fuse
       include_role:
         name: fuse_managed
--- a/roles/application_monitoring/defaults/main.yml
+++ b/roles/application_monitoring/defaults/main.yml
@@ -1,0 +1,5 @@
+---
+app_monitoring_namespace: openshift-application-metrics
+app_monitoring_postfix: -{{ app_monitoring_namespace }}
+app_monitoring_display_name: Application Metrics
+app_monitoring_monitoring_label_value : application-monitoring
--- a/roles/application_monitoring/tasks/main.yml
+++ b/roles/application_monitoring/tasks/main.yml
@@ -1,0 +1,8 @@
+---
+- include_role:
+    name: middleware_monitoring
+  vars:
+    monitoring_namespace: "{{ app_monitoring_namespace }}"
+    namespace_postfix: "{{ app_monitoring_postfix }}"
+    monitoring_display_name: "{{ app_monitoring_display_name }}"
+    monitoring_label_value: "{{ app_monitoring_monitoring_label_value }}"
--- a/roles/application_monitoring/tasks/uninstall.yml
+++ b/roles/application_monitoring/tasks/uninstall.yml
@@ -1,0 +1,8 @@
+---
+- include_role:
+    name: middleware_monitoring
+    tasks_from: uninstall
+  vars:
+    monitoring_namespace: "{{ app_monitoring_namespace }}"
+    namespace_postfix: "{{ app_monitoring_postfix }}"
+    monitoring_display_name: "{{ app_monitoring_display_name }}"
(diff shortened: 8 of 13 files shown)


roles/middleware_monitoring/templates/prometheus_operator_cluster_role_binding.yml.j2
```diff
@@ -1,7 +1,7 @@
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
 metadata:
-  name: prometheus-application-monitoring-operator
+  name: prometheus-application-monitoring-operator-{{ monitoring_namespace }}
```
ziccardi: @wei-lee @david-martin  we added the `-{{ monitoring_namespace }}` postfix because we needed this binding to be created for the `serviceaccount:application-metrics:prometheus-application-monitoring` too, however we are not sure this is the best approach. WDYT?

ziccardi:
ping @danielpassos

ziccardi:
What has been done:

- Created the `install_application_metrics.yml` playbook and invoked from the `install.yml`. This invokes the `middleware_monitoring` role.
- Added the `-{{ monitoring_namespace }}` postfix to the `alert_manager_cluster_role_binding.yml.j2`, `prometheus_cluster_role_binding.yml.j2` and `prometheus_operator_cluster_role_binding.yml.j2` name property.

roles/middleware_monitoring/templates/prometheus_operator_cluster_role_binding.yml.j2
david-martin: @ziccardi This approach looks reasonable. 
1 thing to be aware of is how this affects new installs for the install_middleware_monitoring playbook, and how existing clusters will need to be modified to match how a new install looks.

In this case, it looks like the existing clusterRoleBindings for the middleware_monitoring namespace & serviceaccount will need to be modified (probably createed with the new name, and delete the old bindings).

wei-lee: yeah, to @david-martin's point, if we do it this way, we have to manage the update of these role bindings in an existing cluster. One way to avoid doing it is to add a "monitor_role_binding" postfix variable and set it to empty string for the existing playbook, but a different value for the new playbook. This way they won't change so we don't need to delete & recreate for existing clusters.

ziccardi: @wei-lee @david-martin 👍

aliok: >  One way to avoid doing it is to add a "monitor_role_binding_postfix" variable and set it to empty string for the existing playbook, but a different value for the new playbook. 

sounds good.

> how existing clusters will need to be modified to match how a new install looks

Out of curiosity, I guess we care about having everything identical because of maintenance and possibly automated upgrades.
I think this will not be the only time that a name or a value will change. Is there a concept already in place for handling situations like this?

david-martin: > sounds good.
> 
> >

I agree with this approach



> Is there a concept already in place for handling situations like this

There is no example of this yet, but I would expect any upgrade steps/playbook to do what is necessary to get things to the same state as what a new install looks like

aliok: Thanks!

aliok:
* Rebased and squashed
* Tested on cluster (fresh cluster)
* Uninstall is part of another JIRA

pb82:
:eyes:

pb82:
The monitoring stack is installed successfully to a namespace `Application Metrics`. However no dashboard is imported. That is because at the moment every dashboard will only ever be imported once and there is already the middleware monitoring stack. This is being fixed in another task at the moment.

pb82:
@aliok @ziccardi This also currently duplicates all the Alerts from the middleware namespaces: they are already handled in the `middleware-monitoring` namespace and importing them here too will likely cause alerts firing twice. I think we should use other labels for Prometheus and Grafana resource scanning (ServiceMonitors, PrometheusRules and GrafahaDashboards).

aliok:
Now I am thinking maybe we should create a separate role for this (e.g. application-metrics). We also need to import separate dashboards etc.
Maybe extract the bits that create related CRDs in a separate role and use that in middleware-monitoring and the new application-metrics roles. 

Alternatively, we can create the new role still but make sure that new role is executed after existing middleware-monitoring role.

@ziccardi @pb82 what do you think? cc @wei-lee

pb82:
@aliok A separate role is ok if needed. I think the problem is that at the moment all monitoring resources use the same hardcoded label (monitoring-key: middleware). We should make that configurable in the application monitoring operator. Then we could properly separate the two monitoring stacks and not have each one import the others dashboards and alerts.

wei-lee:
@aliok I agree with @pb82. It's probably doesn't really need a separate role, but if it will make it easier to work on, without duplicating existing code (you can still import tasks from the existing role), you can create a new role, and maybe pass in different configurations for the application metrics.

aliok:
Thanks to @pb82 I understand what the issue is and we will work on it together.

ziccardi:
@wei-lee @pb82 @david-martin In the end we created a new role for the application-monitoring so that it has been easier to group install and uninstall. The role includes the `middleware-monitoring` one so that we don't duplicate code.

aliok:
@pb82 could you have a look again? I pass label selector to app monitoring operator now.

roles/middleware_monitoring/templates/application_monitoring_cr.yml.j2
```diff
@@ -0,0 +1,6 @@
+apiVersion: applicationmonitoring.integreatly.org/v1alpha1
+kind: ApplicationMonitoring
+metadata:
+  name: example-applicationmonitoring
```
pb82: Can we template the name so it's either `middleware-monitoring` or `application-monitoring`?

aliok: +1

roles/middleware_monitoring/templates/application_monitoring_cr.yml.j2
```diff
@@ -0,0 +1,6 @@
+apiVersion: applicationmonitoring.integreatly.org/v1alpha1
+kind: ApplicationMonitoring
+metadata:
+  name: example-applicationmonitoring{{ namespace_postfix }}
```
aliok: @pb82 this is templated now. just like other cases, I made sure the CR name is not changed in regular middleware monitoring

pb82: Maybe also remove the `example-` from the name since it's not not really an example.

aliok: @pb82 I would love to, but see this comment above:

> https://github.com/integr8ly/installation/pull/542#discussion_r274021508

don't want to break existing stuff

pb82: Hmm, this should not have any effect on any of the created resources or roles, but we can also change it later. I'll try this PR on my cluster now.

aliok:
@pb82 tested and squashed. Was blocked by https://github.com/integr8ly/installation/pull/606 and includes those changes too.

aliok:
Rebased now again @pb82

maleck13:
@pb82 @aliok Are there new components coming in here? If so we need to add them to the manifest and we will also need to think about adding them to the Delorean work. @mikenairn

mikenairn:
>  If so we need to add them to the manifest and we will also need to think about adding them to the Delorean work.

This looks to me to be an extension to the middleware monitoring stack which is already in Delorean here https://integreatly-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/view/Delorean/job/delorean-middleware-monitoring/. Unless i'm missing something?

pb82:
@mikenairn @maleck13 This adds another monitoring stack used by Aerogear (mobile and sync) to collect and consume the non-middleware metrics and dashboards. The stack itself is identical to middleware-monitoring, it just uses different label selectors. Is this correct @aliok

aliok:
> @mikenairn @maleck13 This adds another monitoring stack used by Aerogear (mobile and sync) to collect and consume the non-middleware metrics and dashboards. The stack itself is identical to middleware-monitoring, it just uses different label selectors. Is this correct @aliok

Yes, this is correct.

aliok:
@pb82 could you please test and merge this one?

pb82:
👀

PR #548, opened 2019-04-12
reorder links

pb82:
Make sure that the Solution Explorer link is the first one in the Openshift Console. For this we need to add it to the component manifest as the first component.

Verification steps:

1. Run the installer from this branch (uninstall first if already installed)
1. Open the console (accept all certs for scripts) and verify that the first link is the Solution Explorer


matskiv:
👀

PR #556, opened 2019-04-16
INTLY-1754 Use sso namespace var for backup alert

david-martin:
## Additional Information
https://issues.jboss.org/browse/INTLY-1754

## Verification Steps

- Install with backups enabled and an ns_prefix value (doesn't matter what the value is, as long as its set to something).
- Verify the alert for `CronJobExists_sso_daily-at-midnight` is *not* triggering as it is now checking the correct namespace for the `CronJob`.
  - This can be done in Prometheus > Alerts.
  - The PromethesRule can also be checked in the `{prefix}middleware-monitoring` namespace.

No, for existing clusters, a manual fix will be applied via `oc` and `jq` to fix the namespace in the prometheusrule.

### DIFF
--- a/roles/backup/defaults/main.yml
+++ b/roles/backup/defaults/main.yml
@@ -19,7 +19,7 @@ backup_expected_cronjobs: |
       "resources-backup",
       "launcher-postgres-backup"
     ],
-    "sso": [
+    "{{eval_rhsso_namespace}}": [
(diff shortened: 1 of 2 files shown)


david-martin:
Updated to include user-sso namespace

image

david-martin:
@odra if you have time, this is ready for re-review

PR #557, opened 2019-04-16
AEROGEAR-8976 DataSync Roles

wtrocki:
## Additional Information

Following requirements from the product team, adding standard DataSync templates to integr8ly installer. https://issues.jboss.org/browse/AEROGEAR-8976

## How this works

2 new templates (Application and showcase) will be installed by default into OpenShift namespace. Templates going to be sourced from github - please let me know if that is a problem. I tried my best to follow the general structure of the existing roles and other products. Haven't really tested that so posting for early feedback.

### DIFF
--- a/playbooks/install_services.yml
+++ b/playbooks/install_services.yml
@@ -171,3 +171,10 @@
         apicurio_namespace: "{{ eval_apicurito_namespace }}"
         sso_realm: "{{ rhsso_realm }}"
       tags: ['msbroker']
+
+    - 
+      name: Install DataSync Templates
+      include_role:
+        name: datasync
+      tags: ['datasync']
+      when: datasync
--- a/roles/datasync/defaults/main.yml
+++ b/roles/datasync/defaults/main.yml
@@ -1,6 +1,6 @@
 ---
 datasync_template_namespace: openshift
-datasync_template_tag: "0.1.0"
+datasync_template_tag: "0.1.1"
 datasync_template_repository: "https://raw.githubusercontent.com/aerogear/datasync-deployment/"
 datasync_app_template: "{{ datasync_template_repository }}{{ datasync_template_tag }}/openshift/datasync-http.yml"
 datasync_showcase_template: "{{ datasync_template_repository }}{{ datasync_template_tag }}/openshift/datasync-showcase.yml"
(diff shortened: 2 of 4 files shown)


wtrocki:
ping @ziccardi @aliok

roles/datasync/defaults/main.yml
```diff
@@ -0,0 +1,6 @@
+---
+datasync_template_namespace: openshift
```
aliok: I am not sure if this is right

playbooks/install_services.yml
```diff
@@ -171,3 +171,10 @@
         apicurio_namespace: "{{ eval_apicurito_namespace }}"
         sso_realm: "{{ rhsso_realm }}"
       tags: ['msbroker']
+
+    - 
+      name: Install DataSync Templates
+      include_role:
+        name: datasync
+      tags: ['datasync']
+      when: datasync
```
aliok: This would need something like this: https://github.com/integr8ly/installation/blob/fb6b809f7986fb6078aef4fa4dc7bf250514338c/inventories/group_vars/all/manifest.yaml#L69

aliok:
@wtrocki I added some comments, not sure if they're 100% good

wtrocki:
@aliok Thanks a lot! I will need to add this in order to test it but did not wanted to enable datasync yet by default. 

I'm not sure if the integ8ly team want this to be installed now or after Mobile will become official on this platform so skipped that part. 
I'm just looking for any feedback and ability to test this stuff. 

@maleck13 is auto_test label going to help me to test that?

aliok:
@wtrocki 
> I will need to add this in order to test it but did not wanted to enable datasync yet by default.
+1, good point. But perhaps you can still add that var with value=false for better readability?

wtrocki:
@aliok Awesome! doing that now

roles/datasync/defaults/main.yml
wtrocki: In order for templates to appear in the OpenShift UI they need to be added to OpenShift namespace, but not sure if I'm 100% right either (just seen other templates in this repo doing the same)

odra: That's the right way to do it, we do the same for fuse templates: https://github.com/integr8ly/installation/blob/8af2bb98a3fe55fb3a9fa1766f845c39d1389121/roles/fuse/tasks/main.yml#L13

odra:
code looks ok, wainting for my cluster to be ready and will test it.

PR #558, opened 2019-04-17
Adding flag for handling cors install when master tasks is set to false

pmccarthy:
## Additional Information
It looks like the installer attempts to ssh to the master nodes for adding the webapp cors regardless of whether or not the `run_master_tasks` flag is set to `true` or `false`

## Verification Steps
1. Run an install with the `run_master_tasks` flag set to false.
2. Ensure that the installer does not attempt to access the master node when running the `install_webapp.yml` playbook

### DIFF
--- a/playbooks/install_webapp.yml
+++ b/playbooks/install_webapp.yml
@@ -6,7 +6,7 @@
         name: webapp
         tasks_from: cors
       tags: ['webapp', 'remote']
-      when: webapp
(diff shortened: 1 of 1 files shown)


david-martin:
Good catch @pmccarthy.
I think this might have slipped through as testing the original PR may have been done with an inventory that didn't specify a master host, and hence skipped any block meant for the master host.

PR #561, opened 2019-04-24
Prefix the webapp redirect URL in SSO if it's provided

aidenkeating:
Currently we don't take into account the ns_prefix when setting
the redirect URLs for the webapp in the cluster SSO. This results
in logging out of the webapp failing when the ns_prefix is set
during the installation.

This resolves this issue by taking the ns_prefix into account when
running the installer.

## Verification
- Run the install with `-e ns_prefix='something-'`
- Ensure you can logout of the webapp successfully, there should be no invalid redirect URL issue
- Ensure a default install without the `ns_prefix` option set still works

### DIFF
--- a/inventories/group_vars/all/common.yml
+++ b/inventories/group_vars/all/common.yml
@@ -20,7 +20,7 @@ eval_enmasse_namespace: "{{ ns_prefix | default('')}}enmasse"
 eval_user_rhsso_namespace:  "{{ns_prefix | default('')}}user-sso"
 
 eval_seed_users_count: 50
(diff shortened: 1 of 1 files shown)


aidenkeating:
@tremes This will fix the issue we were seeing

tremes:
@aidenkeating Thanks. Do you have an image or cluster to try it out?

aidenkeating:
@tremes I should have included that, `akeating-7432` has it provisioned with `ns_prefix` as `rhpds-`.

PR #562, opened 2019-04-25
fix: update datasync template version

wtrocki:
## 

DataSync template version update. 
Changed descriptions and template names.

Reference: https://github.com/aerogear/datasync-deployment/releases


wtrocki:
Verified.

PR #563, opened 2019-04-26
Intly 1791 fix msb template processing

sedroche:
## Additional Information
Jira: https://issues.jboss.org/browse/INTLY-1791

Updating the managed service broker template creation, which was not erroring, but also was not passing the `NAMESPACE` param correctly.

## Verification Steps
1. Run this branches uninstall as the ClusterRoleBinding is not deleted correctly otherwise (and then not updated)
2. Run this branch with a namespace prefix set
3. Check clusterrolebinding `default-cluster-account-managed-service` has the correct namespace and userNames
4. Test walkthrough 1 fuse creation works

### DIFF
--- a/roles/msbroker/tasks/main.yml
+++ b/roles/msbroker/tasks/main.yml
@@ -68,19 +68,18 @@
   when: not fuse_online
 
 - name: Create managed service broker template in {{ msbroker_namespace }}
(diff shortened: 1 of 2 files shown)


sedroche:
@JameelB

PR #565, opened 2019-04-26
refactored web console configs without yq

jessesarn:
## Additional Information
https://issues.jboss.org/browse/INTLY-1722
Also fixed it so that prerequisites no longer require sudo password when unzip is already available on system.  This will make it safe to remove the "-e prerequisites_install=false" from the OSD install SOP which will allow the oc version check to prevent the old version on their bastions from being used by accident.

## Verification Steps
The customise_web_console_install.yml and customise_web_console_uninstall.yml playbooks can be run directly on a cluster with Integreatly 1.3 or higher installed.

1. Use 'oc login' to connect to cluster as a user with cluster-admin role
2. Retrieve the current web console config:
```
oc get configmaps webconsole-config -n openshift-web-console --template '{{ index .data "webconsole-config.yaml" }}'
```
3. Take note of the extensions.scriptURLs
4. Run the customise web console install playbook (an inventory isn't necessary since only localhost is needed):
```
ansible-playbook playbooks/customise_web_console_install.yml
```
5. Repeat the command in step 2 and you should now see "https://...console-config......<apps_subdomain>/web-console-config.js" under scriptURLs along with any other pre-existing entries.
6. Repeat steps 4 and 5 again, making sure that a duplicate scriptURL isn't added.
7. Run the customise web console uninstall playbook:
```
ansible-playbook playbooks/customise_web_console_uninstall.yml
```
8. Repeat the command in step 2 and "https://...console-config......<apps_subdomain>/web-console-config.js" should no longer be under scriptURLs but any pre-existing entries have been maintained.
9. Repeat steps 7 and 8 without causing any problems.

No upgrade is needed

### DIFF
--- a/roles/customise-web-console/tasks/install.yml
+++ b/roles/customise-web-console/tasks/install.yml
@@ -1,6 +1,6 @@
 ---
 - name: "get manifest secret from webapp namespace"
-  shell: "oc get secret {{ customise_web_console.webapp.secret }} -n {{ eval_webapp_namespace }} -o yaml | yq r - \"data.generated_manifest\" | base64 -d"
+  shell: oc get secret {{ customise_web_console.webapp.secret }} -n {{ eval_webapp_namespace }} --template '{{ "{{" }} .data.generated_manifest | base64decode {{ "}}" }}'
   register: manifest
   failed_when: manifest.stderr != ""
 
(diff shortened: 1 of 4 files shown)


jessesarn:
@philbrookes  I especially want you to review this since you had worked on this functionality using yq and can make sure that it covers all the bases.  And note: I TOTALLY get why you wanted to use yq after realizing how many complications involved (period in key name, target yaml actually embedded as a multi-line string, configmaps not supporting partial update for just one file, etc.)

By the end, I could see there are ways it could be simplified, but focus was to just get the equivalent functionality covered.

philbrookes:
:eyes:

jessesarn:
@philbrookes ?

david-martin:
@jessesarn @philbrookes I'm seeing an uninstall issue after this landed

```TASK [customise-web-console : determine changes needed to remove script url from web console config] ***
fatal: [127.0.0.1]: FAILED! => {"msg": "The conditional check 'get_javascript_route.stdout + '/' + customise_web_console.javascript_filename in get_web_console_config_result.stdout' failed. The error was: error while evaluating conditional (get_javascript_route.stdout + '/' + customise_web_console.javascript_filename in get_web_console_config_result.stdout): 'dict object' has no attribute 'stdout'\n\nThe error appears to have been in '/root/installation/roles/customise-web-console/tasks/uninstall.yml': line 22, column 7, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n    - name: \"determine changes needed to remove script url from web console config\"\n      ^ here\n"}

PLAY RECAP *********************************************************************
127.0.0.1                  : ok=17   changed=6    unreachable=0    failed=1```

It may be the case that my install failed and didn't get as far as setting up the customisation,
and therefore the uninstall expected certain things to exist.
Does the uninstall bit need to be more lenient if it can't find what it's trying to reset?

jessesarn:
I tested all sorts scenarios regarding to what could be in the webconsole config, but had never considered what if the javascript url wasn't resolvable in the first place.  I'll set up this scenario and add some logic to handle it gracefully.

jessesarn:
It looks like this was due to a when clause missing on two tasks in the uninstall playbook.  However, @pawelpaszki reported a similar problem which pointed to availability of the walk function depending on the version of jq.  I've got both of those covered already, but I came up with two cases which would result in failure:
1. If the console-config namespace exists but the route for the javascript file doesn't.  This could happen if an install is stopped/fails right after the namespace is created but before the route is created.  This could also happen if the route was deleted by user, the uninstall was halted/failed before the namespace was deleted.
2. the first task in the uninstall where it detects if the namespace exists has console-config  hard coded whereas all of the remaining tasks have it using vars.  If it was ever overridden it would prevent uninstall from working.

I'll fix both of these before re-opening PR with all of the fixes.

PR #572, opened 2019-05-01
greatly simplified method to remove yq without jq either

jessesarn:
## Additional Information
https://issues.jboss.org/browse/INTLY-1722

Had to revert https://github.com/integr8ly/installation/pull/565 due to a few issues reported which affected idempotence as well as reliance on walk function which I (and I'm guessing @philbrookes) had stolen from jq's master branch and put in our ~/.jq so it could be used with jq 1.5.

I had said in the previous PR that I knew it could be simplified, but since I had to fix issues anyway and I had remembered pointing @pmccarthy to the yedit module written by one our own, Ken Woodson (https://github.com/kwoodson/ansible-role-yedit/) used extensively in openshift-ansible, so I took another pass at the whole thing. 
The yedit module has been added under the customise_web_console role, though its probably worth moving higher up the chain since any playbook running after it can use it, but that can happen later.  The logic using yedit is far better than jumping through hoops using jinja filters to convert back and forth between yaml/json with jq making the change in the middle.

While looking into the two issues, I discovered that the namespace check in the uninstall playbook was using a hardcoded value instead of the vars like the install playbook and the rest of the uninstall playbook.  The fix for that is included here as well; however, it also made me question whether the console-config namespace should have the ns prefix applied to it as well.  Ping @maleck13 / @david-martin for your thoughts on this. 

## Verification Steps
The customise_web_console_install.yml and customise_web_console_uninstall.yml playbooks can be run directly on a cluster with Integreatly 1.3 or higher installed.

1. Use 'oc login' to connect to cluster as a user with cluster-admin role
2. Retrieve the current web console config:
```
oc get configmaps webconsole-config -n openshift-web-console --template '{{ index .data "webconsole-config.yaml" }}'
```
3. Take note of the extensions.scriptURLs
4. Run the customise web console install playbook (an inventory isn't necessary since only localhost is needed):
```
ansible-playbook playbooks/customise_web_console_install.yml
```
5. Repeat the command in step 2 and you should now see "https://...console-config......<apps_subdomain>/web-console-config.js" under scriptURLs along with any other pre-existing entries.
6. Repeat steps 4 and 5 again, making sure that a duplicate scriptURL isn't added.
7. Run the customise web console uninstall playbook:
```
ansible-playbook playbooks/customise_web_console_uninstall.yml
```
8. Repeat the command in step 2 and "https://...console-config......<apps_subdomain>/web-console-config.js" should no longer be under scriptURLs but any pre-existing entries have been maintained.
9. Repeat steps 7 and 8 without causing any problems.

No upgrade is needed

### DIFF
--- a/roles/customise-web-console/library/yedit.yml
+++ b/roles/customise-web-console/library/yedit.yml
@@ -1,0 +1,969 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+# pylint: disable=wrong-import-order,wrong-import-position,unused-import
+
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+ANSIBLE_METADATA = {'metadata_version': '1.1',
+                    'status': ['preview'],
+                    'supported_by': 'core'}
+
+DOCUMENTATION = '''
+---
+module: yedit
+version_added: "2.6"
+short_description: Create, modify, and idempotently manage yaml files.
+description:
+  - Modify yaml files programmatically.
+options:
+  state:
+    description:
+    - State represents whether to create, modify, delete, or list yaml
+    required: true
+    default: present
+    choices: ["present", "absent", "list"]
+    aliases: []
+  debug:
+    description:
+    - Turn on debug information.
+    required: false
+    default: false
+    type: bool
+    aliases: []
+  src:
+    description:
+    - The file that is the target of the modifications.
+    required: false
+    aliases: []
+  content:
+    description:
+    - Content represents the yaml content you desire to work with.  This
+    - could be the file contents to write or the inmemory data to modify.
+    required: false
+    aliases: []
+  content_type:
+    description:
+    - The python type of the content parameter.
+    required: false
+    choices: ['yaml', 'json']
+    default: yaml
+    aliases: []
+  key:
+    description:
+    - The path to the value you wish to modify. Emtpy string means the top of
+    - the document.
+    required: false
+    default: ''
+    aliases: []
+  value:
+    description:
+    - The incoming value of parameter 'key'.
+    required: false
+    default:
+    aliases: []
+  edits:
+    description:
+    - A list of edits to perform.  These follow the same format as a single edit
+    required: false
+    aliases: []
+  value_type:
+    description:
+    - The python type of the incoming value.
+    required: false
+    default: ''
+    aliases: []
+  update:
+    description:
+    - Whether the update should be performed on a dict/hash or list/array
+    - object.
+    required: false
+    default: false
+    aliases: []
+    type: bool
+  append:
+    description:
+    - Whether to append to an array/list. When the key does not exist or is
+    - null, a new array is created. When the key is of a non-list type,
+    - nothing is done.
+    required: false
+    default: false
+    aliases: []
+    type: bool
+  index:
+    description:
+    - Used in conjunction with the update parameter.  This will update a
+    - specific index in an array/list.
+    required: false
+    aliases: []
+  curr_value:
+    description:
+    - Used in conjunction with the update parameter.  This is the current
+    - value of 'key' in the yaml file.
+    required: false
+    default: None
+    aliases: []
+  curr_value_format:
+    description:
+    - Format of the incoming current value.
+    choices: ["yaml", "json", "str"]
+    required: false
+    default: yaml
+    aliases: []
+  backup_ext:
+    description:
+    - The backup file's appended string.
+    required: false
+    aliases: []
+  backup:
+    description:
+    - Whether to make a backup copy of the current file when performing an
+    - edit.
+    required: false
+    default: false
+    aliases: []
+    type: bool
+  separator:
+    description:
+    - The separator being used when parsing strings.
+    required: false
+    default: '.'
+    aliases: []
+author:
+- "Kenny Woodson <kwoodson@redhat.com>"
+extends_documentation_fragment: []
+'''
+
+EXAMPLES = '''
+# Simple insert of key, value
+- name: insert simple key, value
+  yedit:
+    src: somefile.yml
+    key: test
+    value: somevalue
+    state: present
+# Results:
+# test: somevalue
+
+# Multilevel insert of key, value
+- name: insert simple key, value
+  yedit:
+    src: somefile.yml
+    key: a.b.c
+    value: d
+    state: present
+# Results:
+# a:
+#   b:
+#     c: d
+#
+# multiple edits at the same time
+- name: perform multiple edits
+  yedit:
+    src: somefile.yml
+    edits:
+    - key: a.b.c
+      value: d
+    - key: a.b.c.d
+      value: e
+    state: present
+# Results:
+# a:
+#   b:
+#     c:
+#       d: e
+'''
+
+import copy  # noqa: F401
+import fcntl  # noqa: F401
+import json   # noqa: F401
+import os  # noqa: F401
+import re  # noqa: F401
+import shutil  # noqa: F401
+import time  # noqa: F401
+
+try:
+    import ruamel.yaml as yaml  # noqa: F401
+except ImportError:
+    import yaml  # noqa: F401
+
+from ansible.module_utils.basic import AnsibleModule
+
+
+class YeditException(Exception):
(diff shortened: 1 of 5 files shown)


philbrookes:
:eyes:

philbrookes:
I've tried this twice and keep getting: 
```
TASK [3scale : Create 3Scale SSO provider] ******************************************************************************************************************
fatal: [127.0.0.1]: FAILED! => {"cache_control": "no-cache", "changed": false, "connection": "close", "content": "<?xml version=\"1.0\" encoding=\"UTF-8\"?><error>Your access token does not have the correct permissions</error>", "content_type": "application/xml; charset=utf-8", "date": "Thu, 02 May 2019 16:57:28 GMT", "msg": "Status code was 403 and not [201, 422]: HTTP Error 403: Forbidden", "redirected": false, "set_cookie": "1716f8b2bc7ee85639ddf647f5b264d3=cc0fc6f8544c1c195ac091bfa6fa43bb; path=/; HttpOnly", "status": 403, "strict_transport_security": "max-age=15552000", "transfer_encoding": "chunked", "url": "https://3scale-admin.apps.pbrookes-1639.openshiftworkshop.com/admin/api/account/authentication_providers.xml", "vary": "Accept-Encoding", "x_content_type_options": "nosniff", "x_frame_options": "DENY", "x_request_id": "0314e74e-0706-496a-a708-63ee0388f2de", "x_runtime": "0.086033", "x_served_by": "system-app-2-krhqz", "x_xss_protection": "1; mode=block"}
```
I'll try  on a fresh cluister tomorrow unless you update this with more info!

jessesarn:
I'm not sure how 3scale even came into play here.  I'll request another RHPDS myself to see if I can figure out anything.

philbrookes:
Tried on a fresh cluster, install worked ok, uninstall gave me:
```
TASK [rhsso : Replace webconsole configmap in openshift-web-console namespace] **********************************************************************************************************************************************************************************************************************************************
fatal: [127.0.0.1]: FAILED! => {"changed": true, "cmd": "oc replace -f /tmp/webconsole-config.yaml -n openshift-web-console", "delta": "0:00:00.165941", "end": "2019-05-03 10:13:37.147222", "msg": "non-zero return code", "rc": 1, "start": "2019-05-03 10:13:36.981281", "stderr": "error: error parsing /tmp/webconsole-config.yaml: error converting YAML to JSON: yaml: line 25: mapping values are not allowed in this context", "stderr_lines": ["error: error parsing /tmp/webconsole-config.yaml: error converting YAML to JSON: yaml: line 25: mapping values are not allowed in this context"], "stdout": "", "stdout_lines": []}
```

philbrookes:
oooh, it looks like Python has done it's "thing" to that json:
```
TASK [rhsso : Replace webconsole configmap in openshift-web-console namespace] **********************************************************************************************************************************************************************************************************************************************
fatal: [127.0.0.1]: FAILED! => {"changed": true, "cmd": "oc replace -f /tmp/webconsole-config.yaml -n openshift-web-console", "delta": "0:00:00.165941", "end": "2019-05-03 10:13:37.147222", "msg": "non-zero return code", "rc": 1, "start": "2019-05-03 10:13:36.981281", "stderr": "error: error parsing /tmp/webconsole-config.yaml: error converting YAML to JSON: yaml: line 25: mapping values are not allowed in this context", "stderr_lines": ["error: error parsing /tmp/webconsole-config.yaml: error converting YAML to JSON: yaml: line 25: mapping values are not allowed in this context"], "stdout": "", "stdout_lines": []}
```

Check out all of those `u` chars, I think Python does that when it detects a unicode string right?

philbrookes:
hmm, maybe that's a red herring, looking at the file /tmp/webconsole-config.yaml:
```
# cat /tmp/webconsole-config.yaml 
apiVersion: v1
data:
  webconsole-config.yaml: '{''failed'': False, ''state'': ''present'', ''changed'':
    True, ''result'': [{''edit'': {''kind'': ''WebConsoleConfiguration'', ''clusterInfo'':
    {''adminConsolePublicURL'': ''https://console.apps.pbrookes-2198.openshiftworkshop.com/'',
    ''logoutPublicURL'': ''https://sso-sso.apps.pbrookes-2198.openshiftworkshop.com/auth/realms/openshift/protocol/openid-connect/logout?redirect_uri=https://master.pbrookes-2198.openshiftworkshop.com:443/console'',
    ''consolePublicURL'': ''https://master.pbrookes-2198.openshiftworkshop.com/console/'',
    ''masterPublicURL'': ''https://master.pbrookes-2198.openshiftworkshop.com:443'',
    ''metricsPublicURL'': ''https://hawkular-metrics.apps.pbrookes-2198.openshiftworkshop.com/hawkular/metrics'',
    ''loggingPublicURL'': ''https://kibana.apps.pbrookes-2198.openshiftworkshop.com''},
    ''servingInfo'': {''namedCertificates'': None, ''certFile'': ''/var/serving-cert/tls.crt'',
    ''bindAddress'': ''0.0.0.0:8443'', ''bindNetwork'': ''tcp4'', ''maxRequestsInFlight'':
    0, ''keyFile'': ''/var/serving-cert/tls.key'', ''clientCA'': '''', ''requestTimeoutSeconds'':
    0}, ''apiVersion'': ''webconsole.config.openshift.io/v1'', ''extensions'': {''stylesheetURLs'':
    [], ''properties'': {}, ''scriptURLs'': []}, ''features'': {''inactivityTimeoutMinutes'':
    0, ''clusterResourceOverridesEnabled'': False}}, ''key'': ''extensions.scriptURLs''}]}'
kind: ConfigMap
metadata:
  creationTimestamp: 2019-05-03T09:17:30Z
  name: webconsole-config
  namespace: openshift-web-console
  resourceVersion: "32532"
  selfLink: /api/v1/namespaces/openshift-web-console/configmaps/webconsole-config
  uid: 46d16444-6d84-11e9-b014-06ab19d08042
      logoutPublicURL: ''
```
It seems to have that `logoutPublicURL:''` added in at the end where it doesn't belong.

philbrookes:
right, I've done more digging, this does ultimately appear to be caused by the weird escaping that Python is doing, it's upsetting the regex here: 
https://github.com/integr8ly/installation/blob/master/roles/rhsso/tasks/uninstall.yml#L21

Because if you look at the configmap: it looks like this:
```
webconsole-config.yaml: '{''failed'': False, ''state'': ''present'', ''changed'': True,
```

philbrookes:
ok, finally sorted out my cluster! It turned out that the issue is the value in the configmap should be yaml, but your change are writing it as JSON.

jessesarn:
> hmm, maybe that's a red herring, looking at the file /tmp/webconsole-config.yaml:
> 
> ```
> # cat /tmp/webconsole-config.yaml 
> 
> .... <clipped for readability>
> 
> It seems to have that `logoutPublicURL:''` added in at the end where it doesn't belong.

I used 'oc apply' since its atomic vs. the non-atomic methods:
 1. using 'oc delete' followed by 'oc create' 
 2. check existence and then use  'oc create' or 'oc replace' based on the result
Behind the scenes, 'oc apply' handles new vs. generating a patch for update.  That's great for idempotence; however, the side effect is that it saves the previous state in an annotation (kubectl.kubernetes.io/last-applied-configuration).

When the lineinifile comes along, its regex to look for logoutPublicURL finds both the line in the real config and the line in the last-applied.  The annotation is one line with newlines replaces with \n.  I know there are lots of places where the install playbook is relying on regex matches like that to do various config changes, so I guess 'oc apply' is too risky without going back through and fixing all of the other places.  So instead, I'll redo it without using 'oc apply'.

jessesarn:
> ok, finally sorted out my cluster! It turned out that the issue is the value in the configmap should be yaml, but your change are writing it as JSON.

So the JSON effect is an add-on to what is happening with the last-applied annotation.  When the last-applied value is altered by line/regex matching elsewhere and then 'oc apply' is run again, it generates a patch using the current and last applied annotation which is no longer value json or yaml so its putting it in as a big string.

philbrookes:
if it needs to stay as JSON then this bit will need to be updated to handle that too:
https://github.com/integr8ly/installation/blob/master/roles/rhsso/tasks/uninstall.yml#L21

as the regex is currently assuming it's yaml

jessesarn:
Its staying yaml format.  Using the --from-literal for the 'oc create secret' seemed to act differently when the literal value was enclosed in '' vs "" and/or empty strings in the content using '' vs "".

Using --from-file instead avoids the ambiguity and ensures the yaml contents are always treated as a multi-line string.

PR #582, opened 2019-05-03
Intly 1947

pb82:
Use the actual 1.4 tags for the upgrade playbook and make sure the alerts are recreated.

Verification steps:

1. Install the release-1.3.0 tag
2. Run the upgrade playbook
3. Check that the version of the MSB is 0.0.5
4. Check that the version of the webapp operator is 0.0.15
5. Check that the version of the webapp is 2.10.1
6. Check that the versino of the keycloak operator is 1.3.6
7. Check that the `PrometheusRules` resource in the sso namespace has the [updated rules](https://github.com/integr8ly/keycloak-operator/pull/64/files#diff-cce393af68999c5f732259d28517d0c9R45)

### DIFF
--- a/playbooks/upgrades/upgrade.yaml
+++ b/playbooks/upgrades/upgrade.yaml
@@ -4,7 +4,7 @@
   tasks:
     - set_fact:
         upgrade_msb_image: quay.io/integreatly/managed-service-broker:v0.0.5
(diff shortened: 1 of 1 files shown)


david-martin:
@pb82 can you cherry-pick to v1.4 as well

pb82:
@david-martin yes, will do

PR #591, opened 2019-05-07
Merge v1.4-rhpds to v1.4 to avoid confusion

david-martin:
Merging changes from v1.4-rhpds branch.
I'm not sure where v1.4-rhpds came from, but I'd like to delete it.
All rcs should be tagged from v1.4 only.
Having the v1.4-rhpds branch confuses things.
It will be deleted after this merge


mikenairn:
@trepel Was this branch something you created to get round that issue you were having with the integreatly release version containing "rc2"?

PR #592, opened 2019-05-07
Enable datasync by default

wtrocki:
## Motivation

Enable datasync by default, but not for managed database.

### DIFF
--- a/inventories/group_vars/all/manifest.yaml
+++ b/inventories/group_vars/all/manifest.yaml
@@ -89,9 +89,9 @@ apicurito_version: '0.2.18.Final'
 nexus: true
 nexus_version: '2.14.11-01'
 
-## Enables datasync templates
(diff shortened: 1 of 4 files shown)


wtrocki:
Thank you @david-martin. Making changes now.

inventories/group_vars/all/manifest.yaml
```diff
@@ -89,7 +89,7 @@ apicurito_version: '0.2.18.Final'
 nexus: true
 nexus_version: '2.14.11-01'
 
-datasync: false
+datasync: true
```
wei-lee: don't set the var here. Instead, use the `default` operator to set the default value inline. See https://github.com/integr8ly/installation/search?q=user_rhsso&unscoped_q=user_rhsso.

wtrocki:
It was verified already on demos cluster.

PR #593, opened 2019-05-07
upgrade AMQ Online to 1.1

philbrookes:
## Verification
- Install from v1.3 branch
- go through walkthrough 1A as admin user and an evals user
- Run upgrade playbook
- go through walkthrough 1A as admin user and an evals user


roles/enmasse/defaults/main.yml
```diff
@@ -12,3 +12,7 @@ enmasse_enable_monitoring: false
 enmasse_backup_postgres_secret: 'enmasse-postgres-secret'
 enmasse_postgres_cronjob_name: 'enmasse-postgres-backup'
 enmasse_pv_cronjob_name: 'enmasse-pv-backup'
+enmasse_upgrade_version: "1.1"
+enmasse_upgrade_directory: 'amq-online-install-1.1'
+enmasse_upgrade_zip_url: 'https://access.redhat.com/node/4098741/423/0/13779961'
+enmasse_upgrade_zip_name: 'amq-online-install-1.1.zip'
```
mikenairn: Any reason we download this zip file instead of using git like we do for the install?

roles/enmasse/defaults/main.yml
```diff
@@ -12,3 +12,7 @@ enmasse_enable_monitoring: false
 enmasse_backup_postgres_secret: 'enmasse-postgres-secret'
 enmasse_postgres_cronjob_name: 'enmasse-postgres-backup'
 enmasse_pv_cronjob_name: 'enmasse-pv-backup'
+enmasse_upgrade_version: "1.1"
```
mikenairn: Is this used for anything?

maleck13:
I think we need to do more here to verify the upgrade.

We should create some queues IE walkthrough 1 and then run the upgrade and ensure these queues are still working afterwards

maleck13:
It would be good to understand the upgrade a little better. IE what is changed. As this may have an impact on our SoPs. For example it looks like Postgres is removed in 1.1.0

roles/enmasse/defaults/main.yml
philbrookes: I was just following their docs:
https://access.redhat.com/documentation/en-us/red_hat_amq/7.2/html-single/installing_and_managing_amq_online_on_openshift_container_platform/index#downloading-messaging

tremes:
I can try this out. I mean to do some w1 stuff on 1.3.0 and then to use this upgrade.

roles/enmasse/defaults/main.yml
philbrookes: it seems not, I've removed it now

maleck13:
@philbrookes can you take a look through or learn from the AMQ team what is different in this release to see what other stuff (IE sops etc)

matskiv:
btw: if postgresql is really gone in v1.1, we need to remove its backup job as well.

tremes:
Already talked to Phil so just for the record - I tried the upgrade from 1.3 with working W1 and the walkthrough works OK after upgrade. Broker and service broker images weren't updated because there's still `amq7/amq-online-1-broker-plugin:1.0` and `amq7/amq-online-1-service-broker:1.0`. 

Also the RHSSO service is pending in user's shared ns (not sure if I missed something).

maleck13:
@tremes not sure what you mean about the broker and service broker. Do you mean there a not new images for those?

david-martin:
@philbrookes after running the upgrade playbook it looks to have deployed a new enmasse-operator, so 1 extra pod.

```diff
diff enmasse_1.3.yaml enmasse_upgrade.yaml
2,7c2,8
< address-space-controller-7548bfb7f6-csp5c   1/1       Running   0          40m
< api-server-c6b49c444-ccfmn                  1/1       Running   0          40m
< keycloak-5c8c6b4b4f-fpn78                   1/1       Running   1          40m
< keycloak-controller-7b697bc49-fz74g         1/1       Running   0          40m
< postgresql-1-h2l5b                          1/1       Running   0          40m
< service-broker-76c9567bcc-5r555             1/1       Running   0          40m
---
> address-space-controller-865475ccfc-dcrf6   1/1       Running   0          48s
> api-server-5d44db9c99-s2kck                 1/1       Running   0          54s
> enmasse-operator-67cb99bb7d-t8m6f           1/1       Running   0          54s
> keycloak-5c8c6b4b4f-fpn78                   1/1       Running   1          3h
> keycloak-controller-7b697bc49-fz74g         1/1       Running   0          3h
> postgresql-1-h2l5b                          1/1       Running   0          3h
> service-broker-76c9567bcc-5r555             1/1       Running   0          3h
```

However, I was expecting there to be no postgresql after upgrade, and a console pod to be added.
Looking into this now.

tremes:
@maleck13 I meant that the images (for the brokers) weren't updated to the `1.1` version.

david-martin:
@tremes 

The amq-online-1-broker-plugin was updated in an env var value of the address-space-controller deployment after running this upgrade
```
          - name: BROKER_PLUGIN_IMAGE
            value: registry.redhat.io/amq7/amq-online-1-broker-plugin:1.1
```

Is there somewhere else you have seen the older image being used?

roles/backup/tasks/upgrade.yml
```diff
@@ -0,0 +1,3 @@
+---
+- name: "Remove enmasse postgres backup cronjob"
+  shell: oc delete cronjob {{ enmasse_postgres_cronjob_name }} -n {{ backup_namespace }}
```
matskiv: I think this will fail if we will need to re-run the upgrade. We need to ignore "NotFound" error.

roles/enmasse/tasks/upgrade.yml
```diff
@@ -0,0 +1,70 @@
...
+    namespace: "{{ enmasse_namespace }}"
+    display_name: "AMQ Online"
+
+  #hack to work around oc new-app failing when a ns begins with openshift- the ns will begin with openshift- in OSD so that it is hidden from the end user
+- name: Adjust project check role when using openshift prefix
+  template:
+    src: "project_task.yml"
+    dest: /tmp/enmasse-{{ enmasse_version }}/templates/ansible/roles/project/tasks/main.yml
+    force: yes
+  when: ns_prefix == "openshift-"
```
matskiv: IMHO: this should be done for any prefix. Not that we use other prefixes than "openshift-" at the moment, but it's a possibility.

roles/enmasse/defaults/main.yml
philbrookes: I switched to the github repo

roles/enmasse/tasks/upgrade.yml
philbrookes: changed that

roles/enmasse/defaults/main.yml
```diff
@@ -6,9 +6,12 @@ enmasse_api_server: true
...
-enmasse_authentication_services: ["standard"]
+enmasse_authentication_services: []
 enmasse_clean_artifacts: true
 enmasse_enable_monitoring: false
 enmasse_backup_postgres_secret: 'enmasse-postgres-secret'
 enmasse_postgres_cronjob_name: 'enmasse-postgres-backup'
 enmasse_pv_cronjob_name: 'enmasse-pv-backup'
+enmasse_upgrade_directory: 'amq-online-install-1.1'
+enmasse_upgrade_zip_url: 'https://access.redhat.com/node/4098741/423/0/13779961'
+enmasse_upgrade_zip_name: 'amq-online-install-1.1.zip'
```
mikenairn: I don't think these variables are being used are they?

roles/enmasse/tasks/upgrade.yml
```diff
@@ -0,0 +1,357 @@
...
+
+- name: Ensure tmp dir exists
+  file:
+    path: /tmp/enmasse-{{ enmasse_version }}
+    state: directory
+
+- set_fact:
+    enmasse_playbook_location: enmasse-"{{ enmasse_version }}"/templates/ansible/playbooks/openshift/deploy_all.yml
+    enmasse_inventory_path: /tmp/enmasse-{{ enmasse_version }}/templates/ansible/inventory/
+    enmasse_authentication_services: []
```
mikenairn: Why do you need to use set_fact here for enmasse_authentication_services?

roles/enmasse/tasks/upgrade.yml
```diff
@@ -0,0 +1,357 @@
...
+    force: true
+
+- name: Ensure tmp dir exists
+  file:
+    path: /tmp/enmasse-{{ enmasse_version }}
+    state: directory
+
+- set_fact:
+    enmasse_playbook_location: enmasse-"{{ enmasse_version }}"/templates/ansible/playbooks/openshift/deploy_all.yml
+    enmasse_inventory_path: /tmp/enmasse-{{ enmasse_version }}/templates/ansible/inventory/
```
mikenairn: From here up could be moved into it's own task "download_enmasse_version" and shared between install and upgrade tasks.

roles/enmasse/tasks/upgrade.yml
```diff
@@ -0,0 +1,357 @@
...
+    enmasse_authentication_services: []
+
+- name: Generate EnMasse inventory hosts file
+  template:
+    src: enmasse_hosts.j2
+    dest: "{{ enmasse_inventory_path }}/hosts"
```
mikenairn: Also shared with install

roles/enmasse/tasks/upgrade.yml
```diff
@@ -0,0 +1,357 @@
...
+- name: Generate custom manifest
+  template:
+    src: "manifest.j2"
+    dest: /tmp/generated_manifest
+
+- name: Delete secret if it exists
+  shell: oc delete secret manifest -n {{ eval_webapp_namespace }}
+
+- name: Create manifest secret
+  shell: oc create secret generic manifest --from-file=/tmp/generated_manifest --type=application/json -n {{ eval_webapp_namespace }}
```
mikenairn: Spoke to @maleck13 about this, we should move all this stuff out of the individual component upgrade scripts, and instead have the final part of the main upgrade script be to update this manifest secret.

philbrookes: Cool, that's a great idea

roles/enmasse/defaults/main.yml
philbrookes: maybe not, I'll check and delete if I can.

roles/enmasse/tasks/upgrade.yml
philbrookes: yeah, good idea, I'll do this.

roles/enmasse/tasks/upgrade.yml
philbrookes: and this.

mikenairn:
Trying this again now

PR #594, opened 2019-05-07
create fuse image pull secret

pb82:
Create the image pull secret for registry.redhat.io in the fuse namespace. This assumes that it exists in the openshift namespace.

Had to use a go-template-file because a jsonpath does not support keys that start with a dot.

Verification steps:
1. Run the installer using this branch
1. Check that there exists a `syndesis-pull-secret` in the fuse namespace
1. Make sure that the fuse imagestream is using the registry.redhat.io images


pb82:
Superseded by https://github.com/integr8ly/installation/pull/594

PR #596, opened 2019-05-08
fix: Move datasync default value

wtrocki:
## Motivation

Provide a default value for datasync templates

### DIFF
--- a/playbooks/install_services.yml
+++ b/playbooks/install_services.yml
@@ -177,4 +177,4 @@
       include_role:
         name: datasync
       tags: ['datasync']
-      when: datasync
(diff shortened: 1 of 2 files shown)


StevenTobin:
Would it not be cleaner to uncomment the var in the manifest so that datasync matches the other services in install_services.yml?

As is if someone is using the manifest to specify what to install datasync seems to be different.

wtrocki:
@StevenTobin WDYT about moving datasync to `common.yml` to follow the same approach as `user_rhsso`. The target here is to disable datasync for managed infrastructure

PR #597, opened 2019-05-08
Fuse pull secret msb

pb82:
Creates the syndesis pull secret in the msb and fuse namespaces.

Supersedes https://github.com/integr8ly/installation/pull/594

Verification steps:
1. Run the installer from this branch
1. Make sure the a syndesis-pull-secret is present in the fuse and msb namespaces
1. Make sure that the shared fuse deployment works and that it uses registry.redhat.io imagestreams


david-martin:
👀

PR #604, opened 2019-05-08
added removing enmasse crds to enmasse uninstall task

pawelpaszki:
## Additional Information
https://issues.jboss.org/browse/INTLY-2017

## Verification Steps
1. Install v1.4 of integreatly
2. Uninstall using this PR
3. Install release-1.3.0 (it should not fail on trying to provision enmasse)

### DIFF
--- a/roles/enmasse/tasks/uninstall.yml
+++ b/roles/enmasse/tasks/uninstall.yml
@@ -40,6 +40,6 @@
   failed_when: output.stderr != '' and 'not found' not in output.stderr and 'The system is ensuring all content is removed from this namespace.' not in output.stderr
   changed_when: output.rc == 0
 
-- name: "Remove remaining CRDs from {{ enmasse_namespace | default('enmasse') }}"
+- name: "Remove remaining CRDs"
   shell: oc delete crd -l app=enmasse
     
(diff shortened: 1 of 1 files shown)


mikenairn:
Would changing the uninstall task to call the uninstall playbook from the upstream ansible repo make more sense here https://github.com/jboss-container-images/amq-online-images/blob/amq-online-1-dev/templates/ansible/playbooks/openshift/uninstall.yml. Seems to deal with everything we want to delete including crds.

pawelpaszki:
@mikenairn I have simplified removing CRDs. Would you mind taking another look?

roles/enmasse/tasks/uninstall.yml
```diff
@@ -39,3 +39,7 @@
   register: output
   failed_when: output.stderr != '' and 'not found' not in output.stderr and 'The system is ensuring all content is removed from this namespace.' not in output.stderr
   changed_when: output.rc == 0
+
+- name: "Remove remaining CRDs from {{ enmasse_namespace | default('enmasse') }}"
```
mikenairn: You could remove this "from {{ enmasse_namespace | default('enmasse') }}" since you aren't removing anything from that namespace

mikenairn:
This change looks fine to me since it's just doing what the upstream uninstall playbook https://github.com/jboss-container-images/amq-online-images/blob/amq-online-1-dev/templates/ansible/playbooks/openshift/uninstall.yml#L11

roles/enmasse/tasks/uninstall.yml
pawelpaszki: done

PR #605, opened 2019-05-09
Intly 2007 is upgrade

odra:
## Additional Information

Adds a role and a library to re-sync all images used in a speicfic namespace.

Sample usage:

```yaml
- hosts: local
  tasks:
    - include_role:
        name: images
      vars:
        images_source_namespace: ns_to_retrieve-Image_names
```

## Verification Steps

Running the following command will resync images in the fuse ns:
```
ansible-playbook -i inventotries/host roles/images/test/test.yml
```

It will need to be used by pther upgrade tasks


odra:
role name is a bit weird but didn't know what to call it, maybe image-sync?

roles/images/tasks/main.yml
```diff
@@ -0,0 +1,19 @@
+---
+- import_tasks: prereq-check.yml
+
+- name: Get all pods from {{ images_source_namespace }}
+  shell: "oc get pods -o json -n {{ images_source_namespace }}"
```
odra: is basically retrieve all pods and get all images from them

roles/images/tasks/main.yml
```diff
@@ -0,0 +1,19 @@
...
+  register: pod_list_cmd
+
+- name: Get parsed pod list
+  set_fact:
+    pod_list: "{{ pod_list_cmd.stdout|from_json }}"
+
+- name: Filter all image names
+  pods_images:
```
odra: created a tiny module for that so it wouldn't 50 lines of yaml file

roles/images/tasks/main.yml
```diff
@@ -0,0 +1,19 @@
...
+    data: "{{ pod_list['items'] }}"
+  register: image_list
+
+- name: Import images
+  shell: "oc import-image {{ item.image }} --all=true --confirm -n {{ images_namespace }}"
```
matskiv: btw: not all image streams are in "openshift" NS - 3scale has image streams in it's own NS.
And not all image stream used by one product are in same namespace (Fuse uses mix of openshift and fuse namespaced image streams)

b) we should be aware that some image stream tags are used in multiple deployments, across multiple namespaces. E.g. postgres:10 is used by codeready

maleck13: I wonder would it be better to look at each of the imagestreams in the namepaces we managed and import the latest for each of those rather than using the pods to make the decision?

matskiv: @maleck13 I actually discussed this with @odra on chat, and suggested using ImageChange trigger from deployment config spec, as it provides us with all necessary information. 
Example:
```
spec:
  ...
  triggers:
    - imageChangeParams:
        ...
        from:
          kind: ImageStreamTag
          name: 'amp-apicast:latest'
          namespace: 3scale
      type: ImageChange
```

maleck13: Does that work ok? That is referencing an ImageStreamTag which can reference another ImageStream Tag.

What would be the drawback to running through each imagestream in the namespace and running an import image against it. Is the concern the OpenShift namespace?

matskiv: It does work with image stream that reference another tag, I tested it with some image stream on my cluster that was outdated (e.g. java:latest point to java:8, I run "oc import-image java:latest" and java:8 was updated.

odra: not all projects use deployment configs, such as enmasse, wich will require some checking in the role

matskiv: yes, but the things that don't use deployment configs, don't use image streams either.
We will have to automate their updates in different way (more info in [this SOP draft](https://docs.google.com/document/d/1C2Cjhz2K2rL0U1_mRZMKtRb6-Bc74fmKth_V0t2_9XU/edit?usp=sharing)).

odra:
I've updated the role to read deployment configobjects instead of pods suggested.
Should this role also patch images? Where do we get which images to be patched?

maleck13:
Do you mean moving to the latest patch release? I think that is outside of the scope. With this change we are moving the latest for a particular tag but not actually moving versions.

odra:
ok, another question: can I add this role for every namespace in the upgrade playbook?

maleck13:
It should be limited to the RHMI namespaces but for those yes

odra:
Ok, I`ve added the role in the upgrade playbook

mikenairn:
Ran the test playbook against a fresh install of master, and it worked as expected, images were imported, but nothing was restarted since there were no image changes.

```
# ansible-playbook -i inventories/hosts roles/images/test/import.yml
...
TASK [images : Sync ImageStreams] ********************************************************************************************************************************************************************************************************************************************************
changed: [127.0.0.1] => (item={u'image': u'jboss-amq-63', u'namespace': u'fuse'})                                                                                                                                                                                                         
changed: [127.0.0.1] => (item={u'image': u'postgres_exporter', u'namespace': u'openshift'})                                                                                                                                                                                               
changed: [127.0.0.1] => (item={u'image': u'fuse-ignite-meta', u'namespace': u'openshift'})                                                                                                                                                                                                
changed: [127.0.0.1] => (item={u'image': u'fuse-online-operator', u'namespace': u'fuse'})                                                                                                                                                                                                 
changed: [127.0.0.1] => (item={u'image': u'fuse-ignite-server', u'namespace': u'openshift'})                                                                                                                                                                                              
changed: [127.0.0.1] => (item={u'image': u'fuse-ignite-ui', u'namespace': u'openshift'})
...
# oc get pods -n fuse
NAME                          READY     STATUS      RESTARTS   AGE
syndesis-db-1-mbxtt           2/2       Running     0          29m
syndesis-meta-1-2l7bp         1/1       Running     0          29m
syndesis-oauthproxy-2-bh7r6   1/1       Running     0          26m
syndesis-operator-1-j5hzm     1/1       Running     0          29m
syndesis-prometheus-1-sz5lf   1/1       Running     0          29m
syndesis-server-1-qw2j5       1/1       Running     1          29m
syndesis-ui-1-j9fjs           1/1       Running     0          29m
todo-1-build                  0/1       Completed   0          29m
todo-1-xv6hl                  1/1       Running     0          27m
```

Will now test an upgrade from 1.3

mikenairn:
Failed during an upgrade 1.3 -> 1.4

```

odra:
@mikenairn could you please try again?

mikenairn:
Retested on a 1.3 -> 14 upgrade, worked as expected.

maleck13:
Looks like we are good to merge. Let's also get this cherry picked so it is part of the 1.4 upgrade.

PR #606, opened 2019-05-09
Added the missing cluster roles for Prometheus added in the Applicati…

davidkirwan:
…on Monitoring Operator 0.0.8

## Additional Information

- JIRA: https://issues.jboss.org/browse/INTLY-2009

## Verification Steps

1. Create Openshift workshop on RHPDS
2. Install Integreatly from this branch `davidkirwan:prometheus_cluster_roles`
3. Check the prometheus pod comes up correctly


davidkirwan:
This might require an upgrade task also, I'm not entirely sure how roles get handled, is it a once off thing when we run the installer?

aliok:
I tested this change with https://github.com/integr8ly/installation/pull/542 and fixed the problem of crashing Prometheus.

PR #607, opened 2019-05-09
add postgres version to secret

pb82:
Add the postgres version to the secret, defaulting to 9.6 and only using 10 for 3scale.

### DIFF
--- a/roles/backup/tasks/_create_postgres_secret.yml
+++ b/roles/backup/tasks/_create_postgres_secret.yml
@@ -9,6 +9,7 @@
     host: '{{ secret_postgres_host }}'
     superuser: '{{ secret_postgres_superuser | default("false") }}'
     password: '{{ secret_postgres_password }}'
+    version: '{{ secret_postgres_version | default("9.6") }}'
 
 - name: Create Postgres secret {{ secret_name }}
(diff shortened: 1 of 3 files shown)


roles/backup/tasks/_create_postgres_secret.yml
```diff
@@ -9,6 +9,7 @@
     host: '{{ secret_postgres_host }}'
     superuser: '{{ secret_postgres_superuser | default("false") }}'
     password: '{{ secret_postgres_password }}'
+    version: '{{ secret_postgres_version | default("9.6") }}'
```
matskiv: Maybe this a dumb question, but why aren't we defaulting to empty string?

pb82: That would work too because the script already picks 9.6 when the result is empty. This just doubly-ensures that we always use 9.6 unless otherwise specified.

odra: I believe installer will set  a different value only for 3scale.. everything else will be kept as it is to use the default version

matskiv: You are right guys, this will work perfectly fine.
And I was right as well, It is a dumb question 😂

matskiv:
Looks good. I'll test this a minute.

PR #608, opened 2019-05-09
add postgres version to secret

pb82:
Cherry pick of https://github.com/integr8ly/installation/pull/607


maleck13:
Have we captured the need for an upgrade here? Looks like we need to add that param to the secret as part of an upgrade

pb82:
@maleck13 Yes, i'm working on that now: https://issues.jboss.org/browse/INTLY-2034

PR #609, opened 2019-05-09
Intly 1750 3scale 2.4 to 2.5 upgrade

philbrookes:
## Verification
- Install from 1.3
- Login to 3scale as admin
- Create a new API
- Create some new Plans on that API
- Run upgrade playbook
- Log back into 3scale as admin
- see version 2.5.0 in lower left corner
- see changes preserved through upgrade


maleck13:
WalKthough 2 here would also be a good test. Ensuring that an API against a running application continued to work after an upgrade.

pawelpaszki:
I am going to verify it on my PDS cluster

pawelpaszki:
I went through the WT2 and all of the data created before the upgrade was available and fully functioning after the upgrade. The correct version of 3scale was displayed as well.

PR #612, opened 2019-05-10
ON HOLD - DO NOT MERGE - INTLY-2029 Remove Enmasse Postgres backup job & alert

david-martin:
## Additional Information
https://issues.jboss.org/browse/INTLY-2029

## Verification Steps

1. Install as normal, specifying a var override of `backup_restore_install=true`
2. Verify the Cronjob for `enmasse-postgres-backup` was not created in the integreatly-backups namespace
3. Verify the Cronjob alert for `enmasse-postgres-backup` is not showing up in Prometheus

Yes,
covered in #593


david-martin:
on hold as this may not be the right thing to do.
Postgres should be used for production installs, but the default install that happens at present uses a h2 in memory db for keycloak. 
In AMQ Online 1.0, a postgresql instance would have been deployed.
Sticking with postgresql makes sense for a few reasons:

* easier to do the upgrade (not chaning data source)
* the recommended approach for production environments
* easier/possible to backup

The way in which keycloak is deployed for amq 1.1 has changed to be an AuthenticationService CR https://enmasse.io/documentation/0.28.0/kubernetes/#con-authentication-services-messaging

PR #616, opened 2019-05-10
INTLY-2067 - Add Postgres instance & AuthenticationService to EnMasse

david-martin:
## Additional Information
https://issues.jboss.org/browse/INTLY-2067

## Verification Steps

* Install as normal
* Verify enmasse installed OK
* Verify keycloak in enmasse namespace is using postgresql for persistenct

See #593

### DIFF
--- a/roles/enmasse/tasks/install.yml
+++ b/roles/enmasse/tasks/install.yml
@@ -46,6 +46,32 @@
     chdir: "../"
   when: enmasse_resources_exist.stderr == "No resources found."
 
+- name: "Provision Postgres for EnMasse AuthenticationService"
+  shell: oc new-app --template postgresql-persistent -p POSTGRESQL_DATABASE=amqonline -p POSTGRESQL_VERSION=9.6 -n {{ enmasse_namespace }}
+
+- name: Generate AuthenticationService
+  template:
+    src: authenticationservice.yml.j2
+    dest: "/tmp/authenticationservice.yml"
+
+- name: Generate keycloak admin password
+  set_fact:
+    keycloak_admin_password: "{{ lookup('password', '/dev/null length=15 chars=ascii_letters') }}"
+
+- name: Create secret with the keycloak credentials
+  shell: oc create secret generic -n {{ enmasse_namespace }} keycloak-credentials --from-literal=admin.username=admin --from-literal=admin.password={{ keycloak_admin_password }}
+  register: secret_exists
+  failed_when: secret_exists.stderr != '' and 'AlreadyExists' not in secret_exists.stderr
+
+- name: Label secret
+  shell: oc label secret keycloak-credentials -n {{ enmasse_namespace }} app=enmasse
+  when: secret_exists.rc == 0
+
+- name: "Create EnMasse AuthenticationService"
+  shell: oc create -f /tmp/authenticationservice.yml
+  register: authetnicationservice_exists
+  failed_when: authetnicationservice_exists.stderr != '' and 'AlreadyExists' not in authetnicationservice_exists.stderr
+
 - name: "Verify EnMasse deployment succeeded"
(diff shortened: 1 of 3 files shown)


david-martin:
cc @philbrookes

roles/enmasse/tasks/install.yml
```diff
@@ -46,6 +46,32 @@
...
+    src: authenticationservice.yml.j2
+    dest: "/tmp/authenticationservice.yml"
+
+- name: Generate keycloak admin password
+  set_fact:
+    keycloak_admin_password: "{{ lookup('password', '/dev/null length=15 chars=ascii_letters') }}"
+
+- name: Create secret with the keycloak credentials
+  shell: oc create secret generic -n {{ enmasse_namespace }} keycloak-credentials --from-literal=admin.username=admin --from-literal=admin.password={{ keycloak_admin_password }}
+  register: secret_exists
```
maleck13: if it already exists would we want to override the value for the password as it is likely to be different

roles/enmasse/templates/authenticationservice.yml.j2
```diff
@@ -0,0 +1,34 @@
...
+  standard:
+    certificateSecret:
+      name: standard-authservice-cert
+    credentialsSecret:
+      name: keycloak-credentials
+    datasource:
+      type: postgresql
+      host: postgresql
+      port: 5432
+      database: amqonline
```
maleck13: does this match the previous name? Wondering about backup and upgrade

david-martin: Yes, this matches the previous name

roles/enmasse/tasks/install.yml
david-martin: This secret gets used by the keycloak instance that enmasse creates, so I don't think it matters as long as there's a secret with the right name & label before the AuthenticationService is created below.
If the Authentication already exists (e.g. re-run of installer), we probably shouldn't change this secret as it would already have been used to create the admin use in keycloak.

roles/enmasse/tasks/install.yml
```diff
@@ -46,6 +46,32 @@
     chdir: "../"
   when: enmasse_resources_exist.stderr == "No resources found."
 
+- name: "Provision Postgres for EnMasse AuthenticationService"
+  shell: oc new-app --template postgresql-persistent -p POSTGRESQL_DATABASE=amqonline -p POSTGRESQL_VERSION=9.6 -n {{ enmasse_namespace }}
```
matskiv: This fails on a re-run:
```
"stderr": "    error: secrets "postgresql" already exists
    error: services "postgresql" already exists
    error: persistentvolumeclaims "postgresql" already exists
    error: deploymentconfigs.apps.openshift.io "postgresql" already exists"
```

maleck13:
installing on my cluster.

matskiv:
FYI: once I skipped failing step as reported above, Enmasse installed successfully.
And I verified that it is using PostgreSQL for persistence.

maleck13:
Installed succesfully

roles/enmasse/tasks/install.yml
david-martin: Updated to check for postgres deploymentconfig first.

maleck13:
Will test rerun

roles/enmasse/tasks/install.yml
```diff
@@ -46,6 +46,37 @@
     chdir: "../"
   when: enmasse_resources_exist.stderr == "No resources found."
 
+- name: Check if postgresql already exists
+  shell: oc get deploymentconfig postgresql -n {{ enmasse_namespace }}
+  register: postgresql_exists
```
matskiv: Error check is needed here^
```
failed_when: postgresql_exists.stderr != '' and 'not found' not in postgresql_exists.stderr
```

PR #618, opened 2019-05-13
dont move cronjobs

pb82:
The backups in 1.3 are already installed in `openshift-integreatly-backups`. The tasks that move them over from the default namespace have to be removed.

All that's left to do is to add the postgres version to the 3scale secret and patch all cronjobs to use the new version.


odra:
:eyes:

PR #623, opened 2019-05-14
Intly 2007 is upgrade (#605)

odra:
cherry pick of #605


odra:
@maleck13 I saw there is a "v1.4_upgrade" branch, is this the right one?

maleck13:
Not sure why we have that branch. The 1.4 branch is the correct one

PR #627, opened 2019-05-15
[WIP] different secret names on osd and pds

pb82:
The secrets containing the credentials for registry.redhat.io have different names on PDS vs. OSD. This adds a check if we're on OSD and otherwise uses the PDS name.

Verification steps:
1. Run the installer from this branch on PDS
1. Check the secrets in the shared fuse namespace. There should be one named 'syndesis-pull-secret'
1. Start WT1a
1. The per-user instance of fuse should deploy successfully

### DIFF
--- a/roles/fuse_pull_secret/tasks/main.yml
+++ b/roles/fuse_pull_secret/tasks/main.yml
@@ -9,11 +9,11 @@
   failed_when: pick_secret_result.stderr != '' and 'NotFound' not in pick_secret_result.stderr
 
 - set_fact:
-  pull_secret_name: "{{ osd_pull_secret_name }}"
+    pull_secret_name: "{{ osd_pull_secret_name }}"
   when: pick_secret_result.rc == 0
 
 - set_fact:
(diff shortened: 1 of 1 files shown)


roles/fuse_pull_secret/tasks/main.yml
```diff
@@ -1,10 +1,27 @@
...
+    osd_pull_secret_name: registry-redhat-io-dockercfg
+    pds_pull_secret_name: imagestreamsecret
+
+- name: pick the secret depending on the platform
+  shell: "oc get secret {{ osd_pull_secret_name }} -n openshift"
+  register: pick_secret_result
+  failed_when: pick_secret_result.stderr != '' and 'NotFound' not in pick_secret_result.stderr
+
+- set_fact:
+  pull_secret_name: "{{ osd_pull_secret_name }}"
```
odra: I think needs an identitation (same as bellow)

odra:
verified in a pds cluster.

maleck13:
do we need a cherry pick here?

pb82:
@maleck13 yes, doing that now

PR #629, opened 2019-05-15
Remove apicurito template during fuse uninstall

mikenairn:
## Additional Information
* Remove apicurito template from openshift namespace during fuse uninstall.

  The apicurito template is added during the installation of fuse, but it isn't removed when we uninstall. If 
  you run the install against master or 1.4 it will use an apicurito APP_VERSION of 1.3, but the v1.3 
  release expects 1.2 and so fails to deploy if you have previously installed from master or 1.4 and didn't 
  manually remove the template.

* Adds a missing fuse template see https://issues.jboss.org/browse/INTLY-2092

## Verification Steps
* Install from this PR
* Verify apicurito is running
* uninstall
* Install from v1.3
* Verify apicurito is running

Note: This should be rolled back into 1.4 and this commit https://github.com/integr8ly/installation/pull/629/commits/94a605f241a264fc758fc6e173359f584146437c into 1.3 so you can uninstall correctly


odra:
gonna verify in my cluster

PR #633, opened 2019-05-16
remove rhsso_namespace var from manifest

StevenTobin:
## Additional Information
https://issues.jboss.org/browse/INTLY-1790

Tested against managed inventory install and uninstall
Tested against hosts inventory install and uninstall

## Verification Steps
1. Install integreatly with the managed inventory
2. Uninstall integreatly with the managed inventory
3. Previously SSO would not be deleted


matskiv:
Testing it now.

PR #635, opened 2019-05-16
Remove apicurito template from openshift namespace during fuse uninst…

mikenairn:
Backport of https://github.com/integr8ly/installation/commit/5cc516633561a2a6161f45f3abbdfce78825ca25


mikenairn:
This may or not be required, but i tend to do uninstalls from the same branch/tag that i installed from.

PR #641, opened 2019-05-17
Update PULL_REQUEST_TEMPLATE.md

maleck13:
We need to tighten up on ensuring upgrades are covered and Uninstallation is covered


odra:
I think I would add a new "idempotence" (I hope I spelled it right lol)  item for install, uninstall and upgrade  playbooks

PR #645, opened 2019-05-17
remove hardcoded image tags from the upgrade playbook 

JameelB:
## Additional Information
Remove any hardcoded image tags in the upgrade playbook. These tags are available in the manifest file so instead of hardcoding them, use the manifest variables for each product instead.

The image tag used by Fuse in their images is the `<major>.<minor>` version of their Fuse online operator release tag. The Fuse upgrade playbook should extract the image tag to be used during the upgrade from the release tag. 

## Verification Steps
1. Run the installer from the `v1.3` branch
2. Run the upgrade playbook from this branch
    - [x] Ensure upgrade is successful
    - [x] Ensure that the versions of each product upgraded are according to the version in the manifest file

- [ ] Tested Upgrade

### DIFF
--- a/playbooks/upgrades/upgrade.yaml
+++ b/playbooks/upgrades/upgrade.yaml
@@ -3,12 +3,11 @@
   gather_facts: no
   tasks:
     - set_fact:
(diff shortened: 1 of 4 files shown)


odra:
:eyes:

PR #647, opened 2019-05-17
added basic ns cleanup

jessesarn:
## Additional Information
[INTLY-2096](https://issues.jboss.org/browse/INTLY-2096)

## Verification Steps
1. Run installer using managed hosts template
2. Verify openshift-integreatly-backups namespace is created
3. Run uninstall playbook using managed hosts template

No

- [2] Tested Installation
- [2] Tested Uninstallation
- [-] Tested or Created follow on for Upgrade

### DIFF
--- a/playbooks/uninstall.yml
+++ b/playbooks/uninstall.yml
@@ -12,6 +12,14 @@
         name: delete-user-namespaces
       when: keep_namespaces is undefined or not keep_namespaces | bool
     -
+      name: Uninstall integreatly backups
+      include_role:
+        name: backup
+        tasks_from: uninstall
+      tags: ['backup']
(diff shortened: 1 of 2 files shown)


sedroche:
:eyes:

sedroche:
@jessesarn Does this need to go into v1.4?

jessesarn:
@sedroche oops, sorry, somehow I missed your question in my inbox.  I isn't a big deal to have in 1.4.  It wouldn't hurt if it was included, but certainly won't kill us if we don't have it.  We just need to remember to clean up the namespace if we do a full uninstall on an OSD cluster (which should be extremely rare outside of our testing on the integreatly OSD cluster).

PR #652, opened 2019-05-20
upgrade AMQ Online to 1.1

philbrookes:
## Verification
- install from v1.3
- run walkthrough 1A as admin an eval user
- upgrade
- run walkthrough 1A as admin and the same eval user


mikenairn:
I ran through the upgrade from 1.3 -> 1.4 using this branch and verified as described. Everything appeared to be working fine, except i did have a few issues getting the same walkthrough (1A) to complete as the same user after the upgrade. It worked fine when using a different user, but 2 attempts as the original user failed.

philbrookes:
retried on Mike's cluster with one of the users he had used pre-upgrade, and it worked OK for me, so we're assuming it was a user error. Going to merge this now.

roles/enmasse/tasks/install.yml
```diff
@@ -75,7 +45,7 @@
 
 - name: "Create EnMasse AuthenticationService"
   shell: oc create -f /tmp/authenticationservice.yml
-  register: authetnicationservice_exists
+  register: authenticationservice_exists
   failed_when: authetnicationservice_exists.stderr != '' and 'AlreadyExists' not in authetnicationservice_exists.stderr
```
mikenairn: typo wasn't fixed in enough places, added update here https://github.com/integr8ly/installation/pull/655/commits/216a163426aa212009276fb860d9cd03754b7bf3

PR #653, opened 2019-05-20
Reset the upgrade playbook for 1.4 to 1.5 upgrade

david-martin:
## Additional Information

* Remove all upgrade steps that are specific to the 1.3 to 1.4 upgrade
* Leave the imagestreams update as this is useful for every upgrade 

## Verification Steps

* No verification steps. Code review is sufficient as ugprade testing will happen as future changes land 

* N/A

### DIFF
--- a/playbooks/upgrades/upgrade.yaml
+++ b/playbooks/upgrades/upgrade.yaml
@@ -3,12 +3,9 @@
   gather_facts: no
   tasks:
     - set_fact:
-        upgrade_msb_image: quay.io/integreatly/managed-service-broker:{{ msbroker_release_tag }}
         upgrade_webapp_image: quay.io/integreatly/tutorial-web-app:{{ webapp_version }}
         upgrade_webapp_operator_image: quay.io/integreatly/tutorial-web-app-operator:{{ webapp_operator_release_tag }}
-        upgrade_sso_operator_image: quay.io/integreatly/keycloak-operator:{{ rhsso_operator_release_tag }}
-        upgrade_backup_container_image: quay.io/integreatly/backup-container:{{ backup_version }}
-
+        
     - name: Check current version
       shell: "oc get secret manifest -n {{ eval_webapp_namespace }} -o json | jq \".data.generated_manifest\" -r | base64 -d | jq \".version\" -r"
       register: version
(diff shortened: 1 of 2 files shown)


maleck13:
This makes sense to me. @mikenairn any concerns?

mikenairn:
@maleck13 No concerns, but wondering do we need to reset all the individual upgrade roles as well? i.e https://github.com/integr8ly/installation/blob/master/roles/3scale/tasks/upgrade.yml I assume the idea is that these things should be reset once a new minor release is created?

maleck13:
Yes they should be reset also. Which makes this approach a bit more combersome

david-martin:
@mikenairn How about having specific upgrade tasks in the roles?
e.g. `roles/fuse/tasks/upgrade_7.2.1_7.3.0.yml`?

If a component can be upgraded in a pretty generic way most releases, there could be a simple images updater task e.g. `roles/webapp/tasks/upgrade_images.yml`

mikenairn:
This is kind of related https://github.com/integr8ly/installation/pull/657. Those changes should be kept in place in the main upgrade task.

david-martin:
I've rebased and reset again, pulling in the upgrade prereq task from v1.4.

The following tasks are still included:
* webapp version update (exact version to be decided in a different PR)
* image stream updates
* update of version number in manifest

The install of the mobile security service is also included (was already added on master).

@pb82 Would you mind reviewing?

david-martin:
@mikenairn Are you OK with upgrade tasks following a convention of `roles/fuse/tasks/upgrade_7.2.1_7.3.0.yml`,
and the upgrade playbook being the orchestrator of the various upgrade tasks?

This would keep the actual upgrade logic out of the upgrade playbook and make resetting it much easier

mikenairn:
@david-martin Is that not what we are currently doing? Only difference is you want to make an upgrade task per upgrade in the components role, rather than resetting a generic upgrade task ever time? This seems reasonable to me anyway.

david-martin:
Great! Would you mind reviewing/approving?

playbooks/upgrades/upgrade.yaml
```diff
@@ -2,46 +2,15 @@
 - hosts: localhost
   gather_facts: no
   tasks:
+    - include_role:
+        name: prerequisites
+        tasks_from: upgrade
+      vars:
+        from_versions:
+          - "release-1.4.1"
```
mikenairn: Do we expect all patch releases to be applied before bumping a minor version? Maybe that's not what this means, but saw it here and was wondering.

david-martin: This means the playbook should only continue if the cluster is currently at version `release-1.4.1`

For getting to 1.4.1, the previous release will do that upgrade.

PR #655, opened 2019-05-20
Intly 2115 update integeratly version after upgrade v1.4

mikenairn:
## Additional Information

* Ensures the last thing that happens during an upgrade is to call the customisation role. This will update the versions of everything in the manifest stored in the webapp secret including the version of integreatly itself.
* Make backup upgrade stuff optional based on the install variable (backup_restore_install)
* Add restapi route deletion to AMQ upgrade

## Verification
* Install 1.3
* Dump secrets 
  * `oc get secret manifest -n webapp -o=jsonpath="{$.data.generated_manifest}" | base64 -d > 1.3-manifest.yaml`
  * `oc get secret inventory -n webapp -o=jsonpath="{$.data.generated_inventory}" | base64 -d > 1.3-inventory.yaml`
* Run the upgrade v1.4 upgrade from this branch
* Dump secrets
  * `oc get secret manifest -n webapp -o=jsonpath="{$.data.generated_manifest}" | base64 -d > 1.4-manifest.yaml`
  * `oc get secret inventory -n webapp -o=jsonpath="{$.data.generated_inventory}" | base64 -d > 1.4-inventory.yaml`
* Compare and make sure inventory and manifest are now as expected
  * `diff 1.3-inventory.yaml 1.4-inventory.yaml` -> Should be no differences
  * `diff 1.3-inventory.yaml 1.4-inventory.yaml ` -> Should be version changes, enmasse route update and some webapp config.

PR #657, opened 2019-05-21
Intly 2115 update integreatly version after upgrade

mikenairn:
## Additional Information
* Rolling in some changes made on v1.4 that should also be on master from https://github.com/integr8ly/installation/pull/655

## Verification Steps
* Ensure install succeeds from this branch

- [ ] Tested Installation
- [ ] Tested Uninstallation

### DIFF
--- a/playbooks/upgrades/upgrade.yaml
+++ b/playbooks/upgrades/upgrade.yaml
@@ -119,3 +119,6 @@
       include_role:
         name: code-ready
         tasks_from: upgrade
+
+#Update product version (should always be last)
+- import_playbook: "../generate-customisation-inventory.yml"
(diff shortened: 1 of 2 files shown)


playbooks/upgrades/upgrade.yaml
```diff
@@ -108,3 +108,6 @@
     - include_role:
         name: fuse_managed
         tasks_from: upgrade
+
+#Update product version (should always be last)
+- import_playbook: "../generate-customisation-inventory.yml"
```
mikenairn: This should always be executed last after every upgrade, so should be kept here all the time.

philbrookes:
:eye:

PR #658, opened 2019-05-22
split out enmasse role for re-use in upgrades

philbrookes:
## Additional Information
when writing upgrade playbooks for enmasse it is useful to re-use the download and execute of the ansible playbooks for it. This makes that easier for future upgrades.

It also corrects a spelling mistake.


sedroche:
:eyes:

PR #665, opened 2019-05-24
INTLY-2116 - Add status checks after the upgrade

matskiv:
## Additional Information
JIRA: https://issues.jboss.org/browse/INTLY-2116

New tasks have been added to check that individual components are running correctly.
These tasks are invoked during the upgrade + few extra checks landed in upgrade playbook.

P.S: I have to admit, there is fair amount of copy pasted code, so there might be some typos, please review carefully. I will review the code once more myself, but I just wanted to get push this ASAP.

## Verification Steps
Install Integreatly 1.3 release, with `-e ns_prefix='openshift-' -e backup_restore_install='true'` parameters
Check that cluster is in working state (Prometheus -> Alerts ?)
Using this PR, run the upgrade playbook, also with `-e ns_prefix='openshift-' -e backup_restore_install='true'` parameters
Check that cluster is in working state (Prometheus -> Alerts ?)

### DIFF
--- a/playbooks/upgrades/upgrade.yaml
+++ b/playbooks/upgrades/upgrade.yaml
@@ -40,12 +40,11 @@
   gather_facts: no
   tasks:
     - set_fact:
-        upgrade_msb_image: quay.io/integreatly/managed-service-broker:v0.0.6
-        upgrade_webapp_image: quay.io/integreatly/tutorial-web-app:2.10.1
-        upgrade_webapp_operator_image: quay.io/integreatly/tutorial-web-app-operator:v0.0.15
-        upgrade_sso_operator_image: quay.io/integreatly/keycloak-operator:v1.3.6
-        upgrade_backup_container_image: quay.io/integreatly/backup-container:1.0.2
-        upgrade_fuse_image_tag: 1.6
+        upgrade_msb_image: quay.io/integreatly/managed-service-broker:{{ msbroker_release_tag }}
+        upgrade_webapp_image: quay.io/integreatly/tutorial-web-app:{{ webapp_version }}
+        upgrade_webapp_operator_image: quay.io/integreatly/tutorial-web-app-operator:{{ webapp_operator_release_tag }}
+        upgrade_sso_operator_image: quay.io/integreatly/keycloak-operator:{{ rhsso_operator_release_tag }}
+        upgrade_backup_container_image: quay.io/integreatly/backup-container:{{ backup_version }}
 
     - name: Check current version
       shell: "oc get secret manifest -n {{ eval_webapp_namespace }} -o json | jq \".data.generated_manifest\" -r | base64 -d | jq \".version\" -r"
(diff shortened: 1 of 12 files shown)


playbooks/upgrades/upgrade.yaml
```diff
@@ -91,42 +111,69 @@
       shell: oc patch dc tutorial-web-app -n {{ eval_webapp_namespace }} --type json -p '[{"op":"replace", "path":"/spec/template/spec/containers/0/image", "value":"{{ upgrade_webapp_image }}"}]'
       register: upgrade_webapp
       failed_when: upgrade_webapp.stderr != '' and 'not patched' not in upgrade_webapp.stderr
-
-    # SSO Alerts
-    - name: delete alerts
-      shell: "oc delete prometheusrules application-monitoring -n {{ eval_rhsso_namespace }}"
-
-    - name: recreate alerts
-      shell: "oc patch keycloak rhsso --type=json --patch='[{\"op\": \"add\", \"path\": \"/status/monitoringResourcesCreated\", \"value\": false}]' -n {{ eval_rhsso_namespace }}"
```
matskiv: This piece was not removed, I just moved it up to group with other SSO changes.

JameelB:
:eyes:

playbooks/upgrades/upgrade.yaml
```diff
@@ -81,6 +95,12 @@
         name: msbroker
         tasks_from: upgrade
 
+    - include_role:
+        name: msbroker
+        tasks_from: check_readiness
+      vars:
+        msb_image_tag: "{{ upgrade_msb_image.split(':')[1] }}"
```
JameelB: You can use the vars in the manifest file for the image tags. `{{ msbroker_release_tag }}`

playbooks/upgrades/upgrade.yaml
```diff
@@ -65,6 +65,20 @@
     - name: add CreateOnly to openshift keycloak realm
       shell: "oc patch keycloakrealm openshift -n {{ eval_rhsso_namespace }} --type json -p '[{\"op\": \"add\", \"path\": \"/spec/createOnly\", \"value\": true}]'"
 
+    # SSO Alerts
```
odra: can you add a failed_when here? I think failing if the error is not  a "not found error" is enough

roles/webapp/tasks/check_readiness.yml
```diff
@@ -0,0 +1,38 @@
...
+      when: webapp_image_tag is defined
+
+    - name: "Verify webapp is running"
+      shell: oc get pods --selector='app=tutorial-web-app' -o jsonpath='{.items[*].status.containerStatuses[0].ready}' -n {{ webapp_namespace }}
+      register: result
+      until: "'true' in result.stdout"
+      retries: 50
+      delay: 10
+      changed_when: False
+      when: webapp_image_tag is not defined
```
JameelB: Would there be a case where the webapp image tag isn't defined?

roles/webapp/tasks/check_readiness.yml
```diff
@@ -0,0 +1,38 @@
...
+  name: "Check Tutorial WebApp readiness"
+  block:
+    - name: "Verify webapp operator is running"
+      shell: oc get pods --selector='name=tutorial-web-app-operator' -o jsonpath='{.items[?(@.status.containerStatuses[0].ready==true)].spec.containers[0].image}' -n {{ webapp_namespace }} | sed -e 's/.*://'
+      register: result
+      until: result.stdout == webapp_operator_tag
+      retries: 50
+      delay: 10
+      changed_when: False
+      when: webapp_operator_tag is defined
```
JameelB: Do we need this? Is there a case where the webapp operator tag isn't defined?

roles/msbroker/tasks/check_readiness.yml
```diff
@@ -0,0 +1,12 @@
...
+  name: "Check MSB readiness"
+  block:
+    - name: "Verify MSB is running"
+      shell: oc get pods --selector='app=managed-service-broker' -o jsonpath='{.items[?(@.status.containerStatuses[0].ready==true)].spec.containers[0].image}' -n {{ msbroker_namespace }} | sed -e 's/.*://'
+      register: result
+      until: result.stdout == msb_image_tag
+      retries: 50
+      delay: 10
+      changed_when: False
+      when: msb_image_tag is defined
```
JameelB: Same with webapp operator and webapp image tags. Would there be a case where this isn't defined?

roles/webapp/tasks/check_readiness.yml
matskiv: I actually implemented these checks to be called from the installation playbooks as well, but did not push it for 1.4. I will add a PR for master that will include that.
In case of the installation, we don't have tag information available, so it would only check the status.
Should I remove it front here or leave it as is (the same as I will add to master), wdyt?

playbooks/upgrades/upgrade.yaml
matskiv: Good catch, will do.

roles/webapp/tasks/check_readiness.yml
JameelB: @matskiv If that's the case, i'd say leave it as it is :+1:

matskiv:
Re-tested after last commit, works fine.
I will squash and merge if there are no objections. :)

PR #669, opened 2019-05-30
update to latest grafana operator and add oauth proxy

pb82:
installs the latest grafana operator including an oauth proxy.

This PR can be used for verification but the tags need to be changed before merging (still points to `latest`  images and a branch of the application monitoring operator).

Installs the application monitoring operator with those changes: https://github.com/integr8ly/application-monitoring-operator/pull/43

And the grafana operator with those changes: https://github.com/integr8ly/grafana-operator/pull/12

### DIFF
--- a/inventories/group_vars/all/manifest.yaml
+++ b/inventories/group_vars/all/manifest.yaml
@@ -94,8 +94,8 @@ datasync_template_tag: '0.5.0'
 
 monitoring_label_name: 'monitoring-key'
 monitoring_label_value: 'middleware'
-middleware_monitoring_operator_release_tag: '0.0.9'
-middleware_monitoring_operator_resources: 'https://raw.githubusercontent.com/integr8ly/application-monitoring-operator/{{ middleware_monitoring_operator_release_tag }}/deploy'
+middleware_monitoring_operator_release_tag: 'latest'
+middleware_monitoring_operator_resources: 'https://raw.githubusercontent.com/integr8ly/application-monitoring-operator/grafana-latest/deploy'
 
 msbroker_release_tag: 'v0.0.6'
 msbroker_template: 'https://raw.githubusercontent.com/integr8ly/managed-service-broker/{{ msbroker_release_tag }}/templates/broker.template.yaml'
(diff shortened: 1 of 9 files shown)


inventories/group_vars/all/manifest.yaml
```diff
@@ -94,8 +94,8 @@ datasync_template_tag: '0.5.0'
 
 monitoring_label_name: 'monitoring-key'
 monitoring_label_value: 'middleware'
-middleware_monitoring_operator_release_tag: '0.0.9'
-middleware_monitoring_operator_resources: 'https://raw.githubusercontent.com/integr8ly/application-monitoring-operator/{{ middleware_monitoring_operator_release_tag }}/deploy'
+middleware_monitoring_operator_release_tag: 'latest'
```
mikenairn: Any reason we are using latest here?

inventories/group_vars/all/manifest.yaml
```diff
@@ -94,8 +94,8 @@ datasync_template_tag: '0.5.0'
...
+middleware_monitoring_operator_resources: 'https://raw.githubusercontent.com/integr8ly/application-monitoring-operator/grafana-latest/deploy'
```
mikenairn: This should be left as it was and the "middleware_monitoring_operator_release_tag" variable updated to reflect whatever tag you want to use.

inventories/group_vars/all/manifest.yaml
mikenairn: I see now you called this out in the description, can ignore.

pb82: this is only temporary for verification. Once all the other PRs are merged and releases created, i will update this to the proper version.

inventories/group_vars/all/manifest.yaml
pb82: Same as above, only temporary for verification.

davidkirwan:
@pb82 I've installed rhmi from your branch `latest-grafana` on `oauthproxy-201e`

In the `middleware-monitoring` project, grafana has the oauthproxy enabled :)

PR #677, opened 2019-06-05
fix the playbooks to work with the ansible "implicit localhost"

nmasse-itix:
## Additional Information

On a system where multiple python installation co-exists (this is the case by default on MacOS when Ansible is installed with brew), the integr8ly install playbooks fail because of the "jsonpointer" package not being found. 

The README explains that the user needs to setup the `ansible_python_interpreter` accordingly but :
1. it is not so helpful for the newcomers
2. the playbook fails because the Ansible's "implicit localhost" is overriden by a `127.0.0.1` entry in the sample inventory file

For more details, see https://docs.ansible.com/ansible/latest/inventory/implicit_localhost.html

This PR is about removing the `127.0.0.1` entry from the sample inventory files and updating the playbooks to use the implicit localhost entry. 

As a result, the playbooks now correctly pick up the python interpreter and the error message about the "jsonpointer" module not being found is gone.

## Verification Steps

- Clone this repo on MacOS with Ansible installed via brew.sh
- Run the installation script
- Run the un-installation script

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -76,12 +76,6 @@ Prior to running the playbooks the master hostname and associated SSH username *
 ----
 ~/installation/inventories/hosts
 
-[local:vars]
-ansible_connection=local
-
-[local]
-127.0.0.1
-
 [OSEv3:children]
 master
 
@@ -108,10 +102,6 @@ Following an example of the expected output.
 [source,shell]
 ----
 $ ansible -m ping all
-127.0.0.1 | SUCCESS => {
-    "changed": false,
-    "ping": "pong"
-}
 master.example.openshiftworkshop.com | SUCCESS => {
     "changed": false,
     "ping": "pong"
--- a/inventories/hosts.template
+++ b/inventories/hosts.template
@@ -1,11 +1,3 @@
-[local:vars]
-ansible_connection=local
-
-run_master_tasks=true
-
-[local]
-127.0.0.1
-
 [OSEv3:children]
 master
 
--- a/inventories/hosts.template
+++ b/inventories/hosts.template
@@ -5,4 +5,4 @@ master
 ansible_user=ec2-user
 
 [master]
-127.0.0.1
+master.evals.example.com
--- a/inventories/managed.template
+++ b/inventories/managed.template
@@ -1,7 +1,4 @@
-[local]
-127.0.0.1
-
-[local:vars]
+[all:vars]
 ansible_connection=local
 
 # backup_restore_install decides whether to run and install the backup and restore features
--- a/playbooks/install.yml
+++ b/playbooks/install.yml
@@ -1,5 +1,5 @@
 ---
-- hosts: local
+- hosts: localhost
   tasks:
     - block:
       - debug: 
--- a/playbooks/openshift.yml
+++ b/playbooks/openshift.yml
@@ -1,6 +1,6 @@
 ---
 
-- hosts: local
+- hosts: localhost
   gather_facts: no
   tasks:
     - include_role:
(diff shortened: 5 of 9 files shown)


matskiv:
@nmasse-itix Hi, thank you for the contribution!
Unfortunately, implicit `localhost` doesn't work for me. It tries to connect via SSH and fails. Do you know why this happens? Log:
```
PLAYBOOK: install.yml ******************************************************************************************************************
21 plays in ./playbooks/install.yml

PLAY [localhost] ***********************************************************************************************************************

TASK [Gathering Facts] *****************************************************************************************************************
task path: /Users/omatskiv/projects/installation/playbooks/install.yml:2
<127.0.0.1> ESTABLISH SSH CONNECTION FOR USER: ec2-user
<127.0.0.1> SSH: EXEC ssh -C -o ControlMaster=auto -o ControlPersist=60s -o StrictHostKeyChecking=no -o KbdInteractiveAuthentication=no -o PreferredAuthentications=gssapi-with-mic,gssapi-keyex,hostbased,publickey -o PasswordAuthentication=no -o 'User="ec2-user"' -o ConnectTimeout=10 -o ControlPath=/Users/omatskiv/.ansible/cp/8c8d0f635c 127.0.0.1 '/bin/sh -c '"'"'echo ~ec2-user && sleep 0'"'"''
<127.0.0.1> (255, b'', b'ssh: connect to host 127.0.0.1 port 22: Connection refused\r\n')
fatal: [127.0.0.1]: UNREACHABLE! => {
    "changed": false,
    "msg": "Failed to connect to the host via ssh: ssh: connect to host 127.0.0.1 port 22: Connection refused",
    "unreachable": true
}
```

nmasse-itix:
hi @matskiv! 

According to the README (section 2), I updated the inventory to point the `master` section to my actual master:

```ini
[OSEv3:children]
master

[OSEv3:vars]
ansible_user=evals

[master]
master.evals.example.com
```

Maybe updating the sample inventory to replace:

```ini
[master]
127.0.0.1
```

with: 

```ini
[master]
master.evals.example.com
```

would make it clearer ?

matskiv:
I probably found the problem - when master group contains `127.0.0.1` it doesn't work. As soon as I removed `127.0.0.1` from `[master]` (and ran with `-e run_master_tasks=false`) it worked.

matskiv:
@nmasse-itix Yes, you are right, updating master would also work. I didn't do it because I needed an install without master related tasks anyway. Let me check our use cases, whether this won't cause any problems.

BTW: adding this into original implementation could also work:
```
[local:vars]
ansible_python_interpreter="{{ansible_playbook_python}}"
```

matskiv:
These three places would need to be updated as well:
https://github.com/integr8ly/installation/blob/388eace7603040bf03215fa10bf43a579efdc1f5/roles/images/test/import.yml#L1
https://github.com/integr8ly/installation/blob/c3bfdb5cfc832567632e4384531e8fea92af764b/test/inventories/hosts#L1-L5
https://github.com/integr8ly/installation/blob/d0c30612ea4740e5b62768671ff40e236961e89c/roles/customisation/templates/inventory.j2#L1

nmasse-itix:
do you also want me to update the sample inventory :
https://github.com/integr8ly/installation/blob/b8966d010d7ec41a8428c5c2669d78cdf973034f/inventories/hosts.template#L15-L16

with:
```ini
[master]
127.0.0.1 ansible_connection=local
```

so that your use-case still works out-of-the-box ?

or:

```ini
[master]
master.evals.example.com
```

to make it clearer that the master hostname needs to be changed ?

matskiv:
@nmasse-itix Setting sample inventory to "master.evals.example.com" is better idea.
Thanks again for your contribution. I will will update our tooling(it uses sample inventory) and merge it tomorrow.

nmasse-itix:
Thanks @matskiv ! 👍

PR #679, opened 2019-06-07
INTLY-2308: rename wt2 3scale route to a generic name to support reuse

evanshortiss:
## Additional Information
Removes the walkthrough 2 reference from the 3scale user route names.

This PR relies on the walkthroughs also being updated in [this PR](https://github.com/integr8ly/tutorial-web-app-walkthroughs/pull/79) first. So once that's complete the webapp-operator can be updated with the new walkthroughs tag and this can point to the new operator.

## Verification Steps
1. Run this new installer version / playbook
2. Verify route format is `https://wt-{username}-3scale.apps.{cluster-domain}`

Not clear on the meaning of this.

- [x] Tested Installation
- [x] Tested Uninstallation

### DIFF
--- a/inventories/group_vars/all/manifest.yaml
+++ b/inventories/group_vars/all/manifest.yaml
@@ -77,7 +77,7 @@ gitea_operator_resources: 'https://raw.githubusercontent.com/integr8ly/gitea-ope
 webapp: true
 #below vars are not currently used but reflect the current state. In the future they will be used when we source resources from outside of the installer
 webapp_version: '2.10.4'
-webapp_operator_release_tag: 'v0.0.20'
+webapp_operator_release_tag: 'v0.0.19'
 webapp_operator_resources: 'https://raw.githubusercontent.com/integr8ly/tutorial-web-app-operator/{{webapp_operator_release_tag}}/deploy'
 
 #controls whether apicurito is installed or not
--- a/inventories/group_vars/all/manifest.yaml
+++ b/inventories/group_vars/all/manifest.yaml
@@ -77,7 +77,7 @@ gitea_operator_resources: 'https://raw.githubusercontent.com/integr8ly/gitea-ope
 webapp: true
 #below vars are not currently used but reflect the current state. In the future they will be used when we source resources from outside of the installer
 webapp_version: '2.10.4'
-webapp_operator_release_tag: 'v0.0.19'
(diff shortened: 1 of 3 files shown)


inventories/group_vars/all/manifest.yaml
```diff
@@ -77,7 +77,7 @@ gitea_operator_resources: 'https://raw.githubusercontent.com/integr8ly/gitea-ope
 webapp: true
 #below vars are not currently used but reflect the current state. In the future they will be used when we source resources from outside of the installer
 webapp_version: '2.10.4'
-webapp_operator_release_tag: 'v0.0.19'
+webapp_operator_release_tag: 'v0.0.20'
```
evanshortiss: Making an assumption here. Can remove this change or target a different tag if necessary.

maleck13: An automatic PR is created when a new version of the webapp operator is created so we can remove this

evanshortiss: Very nice. Will undo this so 👍

finp:
is this in scope for 1.5?

evanshortiss:
@finp yep

finp:
@maleck13  

> The changes look fine but prob need to wait for the operator PR to close and a new tag created to be able to test it

@maleck13  did PR/tag/testing happen?

finp:
@david-martin This is the PR I spoke about during stand up. As I understand it, out of scope for 1.5, but might be merged to master?

david-martin:
@pmccarthy With 3scale 2.6, has the creation of the wt routes for evals users been removed?
If not, it probably should.
In that case, this PR is not needed, and only a change in the webapp is required to specify whatever route it wants to use (as 3scale creates the route when you click 'Update and test in Staging Environment'

pmccarthy:
@david-martin Yes, I think we should be able to remove the tasks that pre-create those evals routes but wasn't sure if it's something we wanted to have as part of the v1.5 release.

david-martin:
Closing this based on above comments.
Although the `wt2-` routes are still created during install for evals users, they aren't required.
They'll be removed in https://issues.jboss.org/browse/INTLY-3063 for 1.6.

@finp you can start using the `wt-` prefix hosts already in the webapp with no installer change needed.

PR #681, opened 2019-06-11
changed default value of self signed certs to false

pawelpaszki:
## Additional Information
changed default value of self signed certs to false

## Verification Steps
1. Run the installation and confirm that it passes and it is executed with self signed certs set to false

No

- [ ] Tested Installation
- [ ] Tested Uninstallation


odra:
:eyes:

PR #682, opened 2019-06-11
INTLY-2450 Update manifest.yaml

maleck13:
## Additional Information

This version var is only used in the component manifest. When run we should see the version populated in the manfiest secret in the webapp ns

- [ ] Tested Installation
- [ ] Tested Uninstallation


maleck13:
I will recreate against master then cherry-pick to 1.4

PR #686, opened 2019-06-14
set default 3scale admin to customer-admin

pawelpaszki:
## Additional Information
https://issues.jboss.org/browse/INTLY-2392

## Verification Steps
1. Run the fresh installation and confirm that the value of default admin in `system-seed` is `customer-admin` and that there is no `admin` user present in the list of users in `3scale` dashboard. 

2. Run the installation again and confirm that nothing has changed (e.g. no `admin` user and the default admin value in the secret)

- [x] Tested Installation
- [x] Tested Uninstallation

### DIFF
--- a/roles/3scale/defaults/main.yml
+++ b/roles/3scale/defaults/main.yml
@@ -30,10 +30,8 @@ rhsso_seed_users_name_format: evals%02d
 rhsso_seed_users_count: "{{ eval_seed_users_count }}"
 rhsso_seed_users_password: Password1
 
-threescale_cluster_admin_email: admin@example.com
-threescale_cluster_admin_username: "admin"
+threescale_cluster_admin_username: customer-admin
 threescale_cluster_admin_old_username: admin
-threescale_cluster_admin_new_username: "{{ threescale_cluster_admin_username }}"
 
 #Template Params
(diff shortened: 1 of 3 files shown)


jasonmadigan:
@maleck13 @david-martin is this something we also wanted to include in 1.4.1?

maleck13:
This one is probably  fine given the test should be very simple and easy, but we need to be careful with the scope each thing we add increases testing time.

odra:
givint it  a try on pds

odra:
@pawelpaszki I do see an user "admin@example.com" at the end of the user list

trepel:
Based on the JIRA metadata this is set for 1.4.2. I guess that it means we should push to v1.4 no sooner than 1.4.1 GA is out. Correct?

maleck13:
ahh ok I just saw Leo approved.

maleck13:
> Based on the JIRA metadata this is set for 1.4.2. I guess that it means we should push to v1.4 no sooner than 1.4.1 GA is out. Correct?

I have moved this to 1.4.1

maleck13:
I will merge and cherry-pick

maleck13:
cherry pick https://github.com/integr8ly/installation/pull/700

PR #688, opened 2019-06-14
INTLY-2287 Verification of post installation cluster

ciaranRoche:
## Additional Information

JIRA - https://issues.jboss.org/browse/INTLY-2287

*What*
Implement a script that when executed against a given openshift 3.11 cluster will output the readiness state of any integreatly installation. This script should interrogate our monitoring stack (prometheus) for firing alerts, and if any exist the script should fail as the cluster is not in a ready state.

## Verification Steps

1. Run `ansible-playbook -i inventories/hosts playbooks/verify_cluster.yml`, it should complete successfully
2. Scale some pods back to zero
3. Run `ansible-playbook -i inventories/hosts playbooks/verify_cluster.yml` again, it should fail, printing the Alert in the message. *note* it can take some time for the Alerts to update

- [ ] Tested Installation
- [ ] Tested Uninstallation

### DIFF
--- a/playbooks/verify_cluster.yml
+++ b/playbooks/verify_cluster.yml
@@ -1,0 +1,38 @@
+---
+- hosts: localhost
+  gather_facts: no
+  tasks:
+    - name: Fetch cluster ready state.
+      shell: oc -n {{ eval_middleware_monitoring_namespace }} exec -i prometheus-application-monitoring-0 -c prometheus -- /bin/sh -c "curl http://localhost:9090/api/v1/query?query=ALERTS"
+      register: alerts_response
+      changed_when: False
+
+    - name: Parse Alerts from JSON
+      set_fact:
+        alerts: "{{alerts_response.stdout|from_json}}"
+
+    - name: Get DeadMansSwitch State
+      set_fact:
+        DeadMansSwitch_state: "{{alerts | json_query('data.result[?metric.alertname == `DeadMansSwitch`].metric.alertstate')}}"
+
+    - name: Verify DeadMansSwitch state
+      fail:
+        msg: "DeadMansSwitch not firing"
+      when: "'firing' not in DeadMansSwitch_state"
+
+    - name:  Filter DeadMansSwitch
+      set_fact:
+        filter_deadmans_switch: "{{ alerts | json_query('data.result[?metric.alertname != `DeadMansSwitch`]') }}"
+        
+    - name: Filter Prometheus Alerts
+      set_fact:
+        filter_prometheus: "{{ filter_deadmans_switch | json_query('[?metric.prometheus != `openshift-monitoring/k8s`]') }}"
+
+    - name: Filter Firing Alerts Only
+      set_fact:
+        firing_alerts: "{{ filter_prometheus | json_query('[?metric.alertstate == `firing`]') }}"
+        
+    - name: Verify cluster ready state
+      fail:
+        msg: Cluster not ready "{{firing_alerts}}"
+      when: firing_alerts | length > 0
--- a/inventories/group_vars/all/common.yml
+++ b/inventories/group_vars/all/common.yml
@@ -18,6 +18,7 @@ eval_nexus_namespace: "{{ns_prefix | default('')}}nexus"
 eval_managed_fuse_namespace: "{{ns_prefix | default('')}}fuse"
 eval_enmasse_namespace: "{{ ns_prefix | default('')}}enmasse"
 eval_mobile_security_service_namespace: "{{ ns_prefix | default('')}}mobile-security-service"
+eval_middleware_monitoring_namespace: "{{ ns_prefix | default('')}}middleware-monitoring"
 
 eval_user_rhsso_namespace:  "{{ns_prefix | default('')}}user-sso"
 


playbooks/verify_cluster.yml
```diff
@@ -0,0 +1,28 @@
+---
+- hosts: localhost
+  gather_facts: no
+  tasks:
+    - name: Fetch cluster ready state.
+      shell: oc -n middleware-monitoring exec -i prometheus-application-monitoring-0 -c prometheus -- /bin/sh -c "curl http://localhost:9090/api/v1/query?query=ALERTS" | jq .data.result
+      register: alerts_response
```
trepel: Add `changed_when: false` here. This command does not change anything on the host. It is a recommended practice and it is also good for idempotency.

playbooks/verify_cluster.yml
```diff
@@ -0,0 +1,28 @@
...
+    - name: Parse Alerts from JSON
+      set_fact:
+        alerts: "{{alerts_response.stdout|from_json}}"
+
+    - name: Filter Prometheus Alerts
+      set_fact:
+        filter_prometheus: "{{ alerts | json_query('[?metric.prometheus != `openshift-monitoring/k8s`]') }}"
+        
+    - name: Filter DeadMansSwitch
+      set_fact:
```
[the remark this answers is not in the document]
ciaranRoche: Yeah, good idea :+1

playbooks/verify_cluster.yml
```diff
@@ -0,0 +1,38 @@
+---
+- hosts: localhost
+  gather_facts: no
+  tasks:
+    - name: Fetch cluster ready state.
+      shell: oc -n middleware-monitoring exec -i prometheus-application-monitoring-0 -c prometheus -- /bin/sh -c "curl http://localhost:9090/api/v1/query?query=ALERTS"
```
trepel: A namespace prefix (search for `ns_prefix` in this repo) is not taken into account here. For PoC/OSD clusters the namespace prefix is `openshift-` (there is no namespace prefix for PDS).

Not sure how to solve this really, but since the prereq for this playbook is that you are already logged in to the cluster via `oc` (I assume this is ok, but not 100% sure), I think some one-liner could be used in this playbook to see what the namespace prefix actually is:

`oc projects | grep middleware-monitoring | awk '{/ISO/gsub("middleware-monitoring", ""); print $1}'`

mikenairn: You should be able to do something like https://github.com/integr8ly/installation/blob/37116c878344ae7ae9aec71a9d594dca4695d31d/inventories/group_vars/all/common.yml#L11

ciaranRoche: @mikenairn who will call and when will this playbook be called? 
If it is part of the install or upgrade the `ns_prefix` will be known and passed as a param anyway. Where as if we want this to be called ad-hoc by a user they will need to pass the `ns_prefix` as a param with the playbook. If its the latter we will need to capture it in some documentation.

mikenairn: For the moment i think we will leave it as a standalone playbook and not include it in the normal install/upgrade run. We will add a tower job that can be run at any time against any cluster to run this verify play.

All the common group_vars will be available to any playbook being executed in this repo, so you don't need to worry about passing it in, if the "ns_prefix" variable is overridden in an inventory, it will be made available when running this script.

playbooks/verify_cluster.yml
```diff
@@ -0,0 +1,38 @@
...
+        msg: "DeadMansSwitch not firing"
+      when: "'firing' not in DeadMansSwitch_state"
+
+    - name:  Filter DeadMansSwitch
+      set_fact:
+        filter_deadmans_switch: "{{ alerts | json_query('data.result[?metric.alertname != `DeadMansSwitch`]') }}"
+
+    - name: Filter Prometheus Alerts
+      set_fact:
+        filter_prometheus: "{{ filter_deadmans_switch | json_query('data.result[?metric.prometheus != `openshift-monitoring/k8s`]') }}"
```
mikenairn: "filter_deadmans_switch" is an array, but you seem to be expecting a dict with "data.result" like the alerts variable. You need to remove "data.result" here. 
`filter_prometheus: "{{ filter_deadmans_switch | json_query('[?metric.prometheus != `openshift-monitoring/k8s`]') }}"`

playbooks/verify_cluster.yml
```diff
@@ -0,0 +1,38 @@
...
+
+    - name: Filter Firing Alerts Only
+      set_fact:
+        firing_alerts: "{{ filter_prometheus | json_query('data.result[?metric.alertstate == `firing`]') }}"
```
mikenairn: Same here, remove data.result

playbooks/verify_cluster.yml
```diff
@@ -0,0 +1,41 @@
...
+    - name: Filter Prometheus Alerts
+      set_fact:
+        filter_prometheus: "{{ filter_deadmans_switch | json_query('[?metric.prometheus != `openshift-monitoring/k8s`]') }}"
+
+    - name: Filter Firing Alerts Only
+      set_fact:
+        firing_alerts: "{{ filter_prometheus | json_query('[?metric.alertstate == `firing`]') }}"
+
+    - debug: 
+        msg: "{{firing_alerts}}"
```
mikenairn: Can we remove this?

ciaranRoche: Slipped through on the rebase, do that now.

PR #690, opened 2019-06-17
Bump to AMQ Online 1.1.1

maleck13:
… added

## Verification Steps

Run a fresh install. 
Ensure that postgres and keycloak are present
ensure keycloak is configured with the posgres details
run through WT 1
use the following gist with CRs and ensure they are created via CLI 
https://gist.githubusercontent.com/darahayes/ab446986a9d664ef96bce559dfe57b0c/raw/f6c9c7f597099a298c025e05c653cf5e31aed5f0/enmasse-resources.yml

Upgrade is handled in a separate PR

- [x] Tested Installation
- [x] Tested Uninstallation

### DIFF
--- a/roles/enmasse/tasks/install.yml
+++ b/roles/enmasse/tasks/install.yml
@@ -19,38 +19,6 @@
 
 - include_tasks: "_install_enmasse_from_artifact.yml"
 
-- name: Check if postgresql already exists
-  shell: oc get deploymentconfig postgresql -n {{ enmasse_namespace }}
-  register: postgresql_exists
-  failed_when: postgresql_exists.stderr != '' and 'not found' not in postgresql_exists.stderr
-
-- name: "Provision Postgres for EnMasse AuthenticationService"
-  shell: oc new-app --template postgresql-persistent -p POSTGRESQL_DATABASE=amqonline -p POSTGRESQL_VERSION=9.6 -n {{ enmasse_namespace }}
-  when: postgresql_exists.stderr != '' and 'not found' in postgresql_exists.stderr
-
-- name: Generate AuthenticationService
-  template:
-    src: authenticationservice.yml.j2
-    dest: "/tmp/authenticationservice.yml"
-
-- name: Generate keycloak admin password
-  set_fact:
-    keycloak_admin_password: "{{ lookup('password', '/dev/null length=15 chars=ascii_letters') }}"
-
-- name: Create secret with the keycloak credentials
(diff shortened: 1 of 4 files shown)


roles/enmasse/tasks/install.yml
```diff
@@ -19,37 +19,37 @@
...
-
-- name: Generate AuthenticationService
-  template:
-    src: authenticationservice.yml.j2
-    dest: "/tmp/authenticationservice.yml"
-
-- name: Generate keycloak admin password
-  set_fact:
-    keycloak_admin_password: "{{ lookup('password', '/dev/null length=15 chars=ascii_letters') }}"
+#- name: Check if postgresql already exists
```
david-martin: All of this block can be removed, similar to https://github.com/integr8ly/installation/pull/691/commits/35f4b983ab7bdb380c1f36a92713cc22534cfb84

roles/enmasse/tasks/install.yml
```diff
@@ -19,37 +19,37 @@
...
-  shell: oc create -f /tmp/authenticationservice.yml
-  register: authenticationservice_exists
-  failed_when: authenticationservice_exists.stderr != '' and 'AlreadyExists' not in authenticationservice_exists.stderr
+#- name: "Create EnMasse AuthenticationService"
+#  shell: oc create -f /tmp/authenticationservice.yml
+#  register: authenticationservice_exists
+#  failed_when: authenticationservice_exists.stderr != '' and 'AlreadyExists' not in authenticationservice_exists.stderr
 
 - name: "Verify EnMasse deployment succeeded"
   shell: sleep 5; oc get pods --namespace {{ enmasse_namespace }}  |  grep  "deploy"
```
david-martin: This check is not resilient enough. 
It will pass too quickly.

Regardless of the check though, a better approach would be to call out to the `./check_readiness.yml` task
e.g.
```
  - include_tasks: ./check_readiness.yml
```

I've a separate commit that updates the readiness checks here a88c848d9f8e16e0d4d26a9f862867d134e99ca6 which we could add to this PR to master, or PR seperately

maleck13: ok so you are suggesting I take the changes in that commit and add them to this PR?

david-martin: Yes. The commit has changed since then (some tweaks after testing) https://github.com/integr8ly/installation/pull/691/commits/578a6adf01ed971a95966038660998790626c01c

david-martin: On further inspection, the check_readiness files are not on master for any of the product installs.
I'll create a follow up issue for adding them to master & using them after 'install' of each product (not just for upgrades)

PR #699, opened 2019-06-19
Fix Grafana proxy RBAC issue

aliok:
Was having issues with running Grafana proxy in application monitoring namespace. (FYI: application monitoring is using the middleware monitoring role). The name of the cluster role binding did already exist.

Applied the approach in other similar places.

cc @pb82


aliok:
@pb82 mind having a look at this one?

aliok:
@odra can you merge this?

PR #710, opened 2019-06-24
fix: use sync templates 0.6.1

darahayes:
## Verification Steps

1. Deploy Data Sync Showcase from the service catalog
	* The application should be running successfully

@wtrocki it is low priority but maybe we can work together to verify this later. Once we have a testing cluster for it.
 
- [x] Tested Installation
- [ ] Tested Uninstallation


wtrocki:
@darahayes Definitely. We might look with someone that has an integrately cluster and we can just install this role separately.

PR #713, opened 2019-06-25
INTLY-2503 Restart keycloak after upgrade to ensure reconnection to k…

david-martin:
## Additional Information
Waits for upgrade to finish, then trigger a restart of keycloak.
A ready check is then done on the keycloak pod until it's ready

## Verification Steps

1. Install release-1.4.0
2. Run the upgrade playbook
3. Verify the keycloak pod is running and has no exceptions in logs related to postgresql

### DIFF
--- a/roles/enmasse/tasks/upgrade.yml
+++ b/roles/enmasse/tasks/upgrade.yml
@@ -9,3 +9,6 @@
   failed_when: restapi_delete_cmd.stderr != '' and 'not found' not in restapi_delete_cmd.stderr
 
 - include_tasks: ./check_readiness.yml
+
+- name: Trigger restart of keycloak/sso to reconnect to postgresql
+  shell: oc -n {{ enmasse_namespace }} scale deployment/keycloak --replicas=0 && oc -n {{ enmasse_namespace }} rollout status deployment/keycloak


roles/enmasse/tasks/upgrade.yml
```diff
@@ -9,3 +9,14 @@
   failed_when: restapi_delete_cmd.stderr != '' and 'not found' not in restapi_delete_cmd.stderr
 
 - include_tasks: ./check_readiness.yml
+
+- name: Trigger restart of keycloak/sso to reconnect to postgresql
+  shell: oc scale deployment keycloak --replicas=0 -n {{ enmasse_namespace }} && oc scale deployment keycloak --replicas=1 -n {{ enmasse_namespace }}
```
maleck13: I don't think we need to scale it up as the keycloak controller will do that

david-martin: The keycloak controller doesn't exist in 1.1.1.
That being said, you are right that something scales it back up (enmasse-operator?)

maleck13: yes probably

david-martin:
@maleck13 This is ready for verification.
I've tried it on a cluster & keycloak seems to be connecting OK after restart

maleck13:
will cut rc4

PR #723, opened 2019-06-28
replace base64 -d with base64 --decode

StevenTobin:
## Additional Information
To allow an installation to be run from MacOS base64 -d should be base64 --decode

## Verification Steps

* git checkout this branch
* Run a successful installation

- [x] Tested Installation
- [x] Tested Uninstallation

### DIFF
--- a/roles/3scale/tasks/admin_token.yml
+++ b/roles/3scale/tasks/admin_token.yml
@@ -1,6 +1,6 @@
 ---
 - name: Retrieve 3Scale admin auth token
-  shell: oc get secret/system-seed -n {{ threescale_namespace }} -o template --template=\{\{.data.ADMIN_ACCESS_TOKEN\}\} | base64 -d
+  shell: oc get secret/system-seed -n {{ threescale_namespace }} -o template --template=\{\{.data.ADMIN_ACCESS_TOKEN\}\} | base64 --decode
   register: admin_auth_config
 
 - set_fact:
--- a/roles/prerequisites/tasks/upgrade.yml
+++ b/roles/prerequisites/tasks/upgrade.yml
@@ -4,7 +4,7 @@
   when: not (run_master_tasks | default(true) | bool)
 
 - name: Get target version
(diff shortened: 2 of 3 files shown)


StevenTobin:
@Nanyte25 could you try this to validate that it fixes the issues you were seeing?

roles/3scale/tasks/admin_token.yml
```diff
@@ -1,6 +1,6 @@
 ---
 - name: Retrieve 3Scale admin auth token
-  shell: oc get secret/system-seed -n {{ threescale_namespace }} -o template --template=\{\{.data.ADMIN_ACCESS_TOKEN\}\} | base64 -d
```
jessesarn: Might be better to do the base64decode right in the template.   i.e.
```
oc get secret/system-seed -n 3scale -o template --template='{{.data.ADMIN_ACCESS_TOKEN | base64decode}}'
```

roles/prerequisites/tasks/upgrade.yml
```diff
@@ -4,7 +4,7 @@
   when: not (run_master_tasks | default(true) | bool)
 
 - name: Get target version
-  shell: "oc get secret manifest -n {{ eval_webapp_namespace }} -o json | jq \".data.generated_manifest\" -r | base64 -d | jq \".version\" -r"
```
jessesarn: Similar here:
```
oc get secret manifest -n webapp -o template='{{.data.generated_manifest | base64decode}}' | jq ".version" -r
```

scripts/get_admin_secret.sh
```diff
@@ -1,3 +1,3 @@
 #!/usr/bin/env bash
 
-oc get secret admin-user-credentials --template='{{index .data "password"}}' -n sso | base64 -d
```
jessesarn: 1. looks like this doesn't take ns_prefix into account
2. can remove call to external base64 binary with:
```
oc get secret admin-user-credentials --template='{{index .data "password" | base64decode}}' -n sso
```

PR #725, opened 2019-07-01
INTLY-2597/INTLY-2351 Specify prometheus retention of 45d & storage o…

david-martin:
…f 10Gi

## Additional Information

* https://issues.jboss.org/browse/INTLY-2597
* https://issues.jboss.org/browse/INTLY-2351
* Depends on https://github.com/integr8ly/application-monitoring-operator/pull/57
* https://github.com/integr8ly/application-monitoring-operator/pull/56

## Verification Steps

1. Install as normal
2. Verify the prometheus entrypoint inclucdes an arg --storage.tsdb.retention=15d
3. Verify the prometheus pod is mounting a PVC of size 10Gi
4. Verify metrics are retained after a restart of promtheus (can delete the prometheus pod and wait for a new one to start)

* Yes, should be covered by existing upgrade task for the monitoring stack https://github.com/integr8ly/installation/blob/master/playbooks/upgrades/upgrade.yaml#L68-L86

### DIFF
--- a/inventories/group_vars/all/manifest.yaml
+++ b/inventories/group_vars/all/manifest.yaml
@@ -98,7 +98,7 @@ mss_version: '0.1.0'
 
 monitoring_label_name: 'monitoring-key'
 monitoring_label_value: 'middleware'
-middleware_monitoring_operator_release_tag: '0.0.16'
+middleware_monitoring_operator_release_tag: '0.0.17'
 middleware_monitoring_operator_resources: 'https://raw.githubusercontent.com/integr8ly/application-monitoring-operator/{{ middleware_monitoring_operator_release_tag }}/deploy'
 
 msbroker_release_tag: 'v0.0.6'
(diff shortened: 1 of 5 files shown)


david-martin:
@pb82 Can you confirm the upgrade steps already in  https://github.com/integr8ly/installation/blob/master/playbooks/upgrades/upgrade.yaml#L68-L86 are sufficient to pick up on this change too?

david-martin:
I've verified this against a pds cluster. I can try the existing upgrade to see if it adds the retention & storage after.

pb82:
@david-martin The additional fields also have to be added here: https://github.com/integr8ly/installation/blob/master/roles/middleware_monitoring/templates/application_monitoring_cr_upgrade.yml.j2

The reason this exists is that for the upgrade we don't know the name of the monitoring CR (could be middleware-monitoring or example-application-monitoring). So this resource takes the name that is retrieved in a previous step as a template param.

PR #729, opened 2019-07-04
ups installation

odra:
## Additional Information
Adds ups operator as part of the installation (defaults to not install it).

### Insttallation

```
ansible-playbook \
-i inventories/hosts \
-e 'ups=true' \
playbooks/install.yml
```

The above command will result in adding ups in a `unifiedpush` namespace, you can get the web console url, `oc get route -n unifiedpush`, and try signing in (might require incognito mode due to keycloak cookie issue).

Check if the service monitor cr was properly created: `oc get servicemonitor -n unifiedpush`

#### Backups

```
-i inventories/hosts \
-e 'ups=true' \
playbooks/install_backups.yml
```

You can check if the required cronjob was successfully created bu running: `oc get cj -n unifiedpush`

You can also check if the backup definiton was added: `oc get unifiedpushservers/unifiedpush -o yaml -n unifiedpush`

Check if ups was added in the webapp manifest secret: `oc get secret/manifest -o jsonpath='{.data.generated_manifest}' -n webapp | base64 -d`

### Tests

There are a couple of tests playbooks in `roles/ups/test`, those were created for an easier development workflow - this role does not require ssh access so it should work as long as you are logged in an openshift instance.

Sample usage:

```
ansible-playbook -i inventories/hosts roles/ups/test/install.yml
```
### Uninstallation

Just the usual integreatly uninstall playbook, which shoudl remove the unifiedpush namespace:

```
ansible-playbook -i inventories/hosts -e 'eval_self_signed_certs=true' -e 'ups=true' roles/ups/test/uninstall.yml
```

### Upgrade

* Uninstall integreatly
* Install integreatly latest release: `release-1.4.1`
* Swtich to this branch
* Run the upgrade playbook using the ups tag: `ansible-playbook -i inventories/hosts -e 'eval_self_signed_certs=true' -e 'ups=true' playbooks/upgrades/upgrade.yaml` (note: add the following flag to enable backups: `-e 'backup_restore_install=true'`

There should be a `unifiedpush` namespace with a running instance of ups now.

Running the upgrade playbook after installing the latest release (1.4.1) is part of this task's verification - more details in the "Upgrade" section.

- [ ] Tested Installation
- [ ] Tested Uninstallation

### DIFF
--- a/inventories/group_vars/all/common.yml
+++ b/inventories/group_vars/all/common.yml
@@ -18,6 +18,8 @@ eval_nexus_namespace: "{{ns_prefix | default('')}}nexus"
 eval_managed_fuse_namespace: "{{ns_prefix | default('')}}fuse"
 eval_enmasse_namespace: "{{ ns_prefix | default('')}}enmasse"
 eval_mobile_security_service_namespace: "{{ ns_prefix | default('')}}mobile-security-service"
+eval_ups_namespace: "{{ ns_prefix | default('')}}unifiedpush"
+eval_mdc_namespace: "{{ ns_prefix | default('')}}mobile-developer-console"
 eval_middleware_monitoring_namespace: "{{ ns_prefix | default('')}}middleware-monitoring"
 
 eval_user_rhsso_namespace:  "{{ns_prefix | default('')}}user-sso"
--- a/inventories/group_vars/all/manifest.yaml
+++ b/inventories/group_vars/all/manifest.yaml
@@ -96,6 +96,16 @@ mobile_security_service_operator_resources: 'https://raw.githubusercontent.com/a
 mobile_security_service_operator_image: 'quay.io/aerogear/mobile-security-service-operator:{{ mobile_security_service_operator_release_tag }}'
 mss_version: '0.1.0'
 
+#unifiedpush
+ups: false
+ups_operator_release_tag: 'master'
+ups_operator_resources: 'https://raw.githubusercontent.com/aerogear/unifiedpush-operator/{{ups_operator_release_tag}}/deploy'
+ups_operator_image: 'quay.io/aerogear/unifiedpush-operator:{{ ups_operator_release_tag }}'
+ups_release_tag: '2.2.1.Final'
+ups_image: 'docker.io/aerogear/unifiedpush-wildfly-plain:{{ ups_release_tag }}'
+ups_proxy_release_tag: 'v1.1.0'
+ups_proxy_image: 'docker.io/openshift/oauth-proxy:{{ ups_proxy_release_tag }}'
+
 monitoring_label_name: 'monitoring-key'
 monitoring_label_value: 'middleware'
 middleware_monitoring_operator_release_tag: '0.0.17'
--- a/playbooks/install_backups.yml
+++ b/playbooks/install_backups.yml
@@ -37,6 +37,13 @@
       tags: ['user_rhsso']
       when: user_rhsso | default(true) | bool
     - include_role:
+        name: ups
+        tasks_from: backup.yml
+      vars:
+        ups_patch_backup: true
+      tags: ['ups']
+      when: ups | default(true) | bool
+    - include_role:
         name: backup
         tasks_from: monitoring.yml
       when: backup_restore_monitoring | default(true) | bool
--- a/roles/customisation/tasks/main.yaml
+++ b/roles/customisation/tasks/main.yaml
@@ -234,6 +234,7 @@
     component_manifests: "{{component_manifests}} + {{enmasse_manifest}}"
   when: enmasse
 
+#mobile security service(mss)
 - name: Get Mobile Security Service rest route
   shell: oc get route/route -o template --template \{\{.spec.host\}\} -n {{ eval_mobile_security_service_namespace | default('mobile-security-service') }}
   register: mss_route_cmd
@@ -254,6 +255,24 @@
 - set_fact:
     component_manifests: "{{component_manifests}} + {{ mss_manifest }}"
   when: mobile_security_service
+
+#unified push server(ups)
+- name: Unified Push Server
+  block:
+    - name: Get Unified Push Server route
+      shell: oc get route/unifiedpush-unifiedpush-proxy  -o template --template \{\{.spec.host\}\} -n {{ eval_ups_namespace | default('unifiedpush') }}
+      register: ups_route_cmd
+    - set_fact:
+        ups_route: "https://{{ups_route_cmd.stdout}}"
+    - name: Set Unified Push Server component
+      set_fact:
+        ups_manifest:
+          - name: Unified Push Server
+            version: "{{ ups_release_tag }}"
+            host: "{{ ups_route }}"
+    - set_fact:
+        component_manifests: "{{ component_manifests }} + {{ ups_manifest }}"
+  when: ups
 
 - name: Retrieve launcher sso hostvars
   shell: "oc get route/{{launcher_sso_prefix}} -o jsonpath='{.spec.host}' -n {{ launcher_namespace }}"
--- a/roles/ups/tasks/uninstall.yml
+++ b/roles/ups/tasks/uninstall.yml
@@ -1,0 +1,45 @@
+---
+
+- name: Delete ups server
+  shell: "oc delete unifiedpushserver/{{ ups_server_name }} -n {{ ups_namespace }}"
+  register: output
+  failed_when: output.stderr != '' and 'not found' not in output.stderr and 'The system is ensuring all content is removed from this namespace.' not in output.stderr and "the server doesn't have a resource type" not in output.stderr
+  changed_when: output.rc == 0
+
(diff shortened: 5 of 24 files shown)


inventories/group_vars/all/manifest.yaml
```diff
@@ -96,6 +96,16 @@ mobile_security_service_operator_resources: 'https://raw.githubusercontent.com/a
 mobile_security_service_operator_image: 'quay.io/aerogear/mobile-security-service-operator:{{ mobile_security_service_operator_release_tag }}'
 mss_version: '0.1.0'
 
+#unifiedpush
+ups: false
+ups_operator_release_tag: 'master'
```
odra: there is currently no release for the operator now, will update it in another pr once the work there is done.

playbooks/install_backups.yml
```diff
@@ -20,6 +20,7 @@
       include_role:
         name: mobile_security_service
         tasks_from: backup.yml
+      when: mobile_security_service | default(true) | bool
```
laurafitzgerald: Was chatting to @StevenTobin Im missed this in my pr but he has added it in another pr. :)

odra: rebased it

roles/ups/tasks/uninstall.yml
```diff
@@ -0,0 +1,45 @@
+---
```
laurafitzgerald: @odra this looks great. I didn't take all these steps in the uninstall for mss. I just removed the namespace. Should I improve the mss one based on this?

roles/ups/templates/operator.yml.j2
```diff
@@ -0,0 +1,36 @@
...
+            - name: POD_NAME
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.name
+            - name: OPERATOR_NAME
+              value: "unifiedpush-operator"
+            - name: UPS_IMAGE_STREAM_INITIAL_IMAGE
+              value: "{{ ups_image }}"
+            - name: OAUTH_PROXY_IMAGE_STREAM_INITIAL_IMAGE
+              value: "{{ ups_proxy_image }}"
```
laurafitzgerald: I think there will probably be a new env var going in here based on the work done by Jozef in this JIRA https://issues.jboss.org/browse/AEROGEAR-9024

laurafitzgerald: There's an example here: https://github.com/integr8ly/installation/blob/master/roles/mobile_security_service/templates/operator.yml.j2#L28-L29

roles/ups/templates/unifiedpushserver.yml.j2
```diff
@@ -0,0 +1,17 @@
+apiVersion: push.aerogear.org/v1alpha1
+kind: UnifiedPushServer
+metadata:
+  name: {{ ups_server_name }}
+{% if ups_backup %}
```
laurafitzgerald: nice :)

laurafitzgerald: We used a separate cr for the backups in other services. Not sure whether we want to do that here as well?

roles/ups/test/install_and_patch_backups.yml
```diff
@@ -0,0 +1,21 @@
+- hosts: localhost
```
laurafitzgerald: Again I didn't add any tests for the MSS install work that I did. I'll create a follow on JIRA which could be addressed later?

odra: Not really required (feel free to create a jira for it) - I think just one or two roles use it, I add it so I don't have to run the installer from the bastion server all the time

roles/ups/templates/unifiedpushserver.yml.j2
odra: I could be wrong but I think this is the way the ups operator works (it creates  a cornjob in the same namespace) - keycloak operator does the same I think

ping @grdryn

roles/ups/test/install_and_patch_backups.yml
laurafitzgerald: ok. if it's not required I will leave it for now. :)

laurafitzgerald:
@odra great work! Just a couple of comments inline. I don't have a pds right now but requesting one in order to verify this.

roles/ups/templates/unifiedpushserver.yml.j2
grdryn: Yes, the unifiedpush-operator does the same as what the keycloak-operator does (as @odra described it).

laurafitzgerald: interesting. It's not relevant to this pr anyway so will resolve this comment.

roles/ups/templates/operator.yml.j2
odra: are they all going to use the mdc namespace?

odra: Ok I`ve added a var there

inventories/group_vars/all/common.yml
```diff
@@ -18,6 +18,8 @@ eval_nexus_namespace: "{{ns_prefix | default('')}}nexus"
 eval_managed_fuse_namespace: "{{ns_prefix | default('')}}fuse"
 eval_enmasse_namespace: "{{ ns_prefix | default('')}}enmasse"
 eval_mobile_security_service_namespace: "{{ ns_prefix | default('')}}mobile-security-service"
+eval_ups_namespace: "{{ ns_prefix | default('')}}unifiedpush"
+eval_mdc_namespace: "{{ ns_prefix | default('')}}mdc"
```
odra: I added this to be used by `APP_NAMESPACES` (it would be added anyway when we add the mdc operator in the pour).

inventories/group_vars/all/manifest.yaml
```diff
@@ -96,6 +96,16 @@ mobile_security_service_operator_resources: 'https://raw.githubusercontent.com/a
...
+ups_operator_resources: 'https://raw.githubusercontent.com/aerogear/unifiedpush-operator/{{ups_operator_release_tag}}/deploy'
+ups_operator_image: 'quay.io/aerogear/unifiedpush-operator:{{ ups_operator_release_tag }}'
+ups_release_tag: '2.2.1.Final'
+ups_image: 'docker.io/aerogear/unifiedpush-wildfly-plain:{{ ups_release_tag }}'
+ups_proxy_release_tag: 'v1.1.0'
+ups_proxy_image: 'docker.io/openshift/oauth-proxy:{{ ups_proxy_release_tag }}'
```
odra: those version tags are used by the operator + should be used by delorean as well.

roles/customisation/tasks/main.yaml
```diff
@@ -255,6 +256,24 @@
...
+    - name: Get Unified Push Server route
+      shell: oc get route/unifiedpush-unifiedpush-proxy  -o template --template \{\{.spec.host\}\} -n {{ eval_ups_namespace | default('unifiedpush') }}
+      register: ups_route_cmd
+    - set_fact:
+        ups_route: "https://{{ups_route_cmd.stdout}}"
+    - name: Set Unified Push Server component
+      set_fact:
+        ups_manifest:
+          - name: Unified Push Server
+            version: "{{ ups_release_tag }}"
```
odra: Using `ups_release_tag` here as it is our source of ups version, can change it in the future.

laurafitzgerald:
Environment ready. Verifying this change now.

inventories/group_vars/all/common.yml
laurafitzgerald: The namespace will be mobile-developer-console, that's what I've used for MSS here: https://github.com/integr8ly/installation/blob/master/roles/mobile_security_service/defaults/main.yml#L24 But maybe we can move that to this file instead for both?

roles/customisation/tasks/main.yaml
laurafitzgerald: I did something similar for MSS. I think it's fine. We will need to manually update it in the group vars. Do other products have it so it updates automatically?

roles/ups/tasks/uninstall.yml
odra: Yeah this might be a good idea, I did this becaue of all the crs and finalizers, there might be a situation where the operator gets deleted before the crs so it ends up with lingering crs with no operator left to remove those finalizers

inventories/group_vars/all/common.yml
odra: ok let me change that var as well so both use the eval one

roles/customisation/tasks/main.yaml
odra: yeah, I don't think delorean watchers for  mobile releases atm but it will use this file as the single source of truth once it does

laurafitzgerald: That's neat. It would be cool to hook into that in the future.

PR #724, opened 2019-07-01
Adding alert for elastic search pods

davidkirwan:
## Additional Information
JIRA: https://issues.jboss.org/browse/INTLY-2190

## Verification Steps
1. Install from my fork/branch ``davidkirwan:es_pod_alert``.
2. Installer should run to completion.
3. In prometheus dashboard in the ``middleware-monitoring`` namespace, two new alerts should go green: ``ESNotReady`` and ``ESPodCount``.
4. In the ``openshift-logging`` namespace, scale the Elastic Search instance down to ``0``.
5. Both alerts should go ``pending`` within a short period of time and then ``firing`` after 5 minutes.
6. Scale the Elastic Search instance back to 1 (RHPDS)
7. Both alerts should go back green once more.
8. Uninstaller should run to completion.

## Is an upgrade task required and are there additional steps needed to test this?

- [x] Tested Installation
- [x] Tested Uninstallation
- [ ] Tested or Created follow on for Upgrade

PR #707, opened 2019-06-24
Changing Prometheus rules labels from exported_pod to pod

davidkirwan:
## Additional Information
JIRA: https://issues.jboss.org/browse/INTLY-2408

Seems to be an issue related to the label `exported_pods` vs `pods`

## Verification Steps
Once RHMI is installed from my fork/branch `davidkirwan:prometheus_alerts`
1. Go to the `middleware-monitoring` namespace.
2. Edit the `application-monitoring-operator` image, and modify the tag to 0.0.20 (doesn't exist yet).
3. After a short period, will see the following alerts go critical in Prometheus:
-  KubePodImagePullBackOff
-  KubePodNotReady
4. The name of the pod will show amongst the labels in the alert.
5. The alert will also show in Alertmanager.

Screenshot from 2019-06-24 15-18-51
Screenshot from 2019-06-24 15-19-16

## Is an upgrade task required and are there additional steps needed to test this?

- [x] Tested Installation
- [x] Tested Uninstallation
- [ ] Tested or Created follow on for Upgrade

PR #734, opened 2019-07-05
INTLY-2190 Apply ksm alerts to ensure we have all the latest rules after upgrade

david-martin:
## Additional Information
Follow up from https://github.com/integr8ly/installation/pull/724 (upgrade for changes in https://issues.jboss.org/browse/INTLY-2190)

This PR changes the way the upgrade of the ksm-alerts PrometheusRule is done.

*Before*

The CR was exported, modified using sed, then reimported.

*After*

The CR is applied to the cluster (`oc apply`) to ensure it has the latest changes.

*Rationale*

This way of upgrading of individual rule changes is safe and less error prone.
The CR should not be modified in the cluster outside of the install or upgrade. We can't allow for any changes outside of the install & upgrade.
Therefore, the CR can simply be updated in-situ.
The prometheus-operator will reconcile the latest alerts and get any changes we've made.

## Verification Steps

1. Install from tag `release-1.4.1`
2. Checkout this branch
2. Run the upgrade playbook
3. Verify the alert changes in https://github.com/integr8ly/installation/pull/707 & https://github.com/integr8ly/installation/pull/724 have been applied and are showing up correctly in Prometheus UI > Alerts tab. This may take a few minutes to update


davidffrench:
:eyes:

PR #738, opened 2019-07-08
Intly 2120 user sso upgrade

maleck13:
## Additional Information
Add the user level sso to existing clusters. Also modifies user permissions on the user sso ns so that the customer-admin can view the ns. 

## Verification Steps
- Run an install from 1.4.1
- no user sso should be installed in the user-sso ns
- Run an upgrade from this branch ```ansible-playbook -i inventories/hosts.template playbooks/upgrade/upgrade.yaml ```
- login as the `customer-admin``` user you should see the user-sso ns

- [x] Tested Installation
- [x] Tested Uninstallation

### DIFF
--- a/playbooks/upgrades/upgrade.yaml
+++ b/playbooks/upgrades/upgrade.yaml
@@ -52,6 +52,7 @@
       loop_control:
         loop_var: images_source_namespace
 
+
     # Update Mobile Security Service - TODO at next version - change to upgrade - current version will run installation
     - name: Install/Upgrade Mobile Security Service as part of upgrade
       include_role:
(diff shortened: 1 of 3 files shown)


dimitraz:
Verification steps:

- [ ]
- [ ]

playbooks/upgrades/upgrade.yaml
```diff
@@ -52,6 +52,9 @@
       loop_control:
         loop_var: images_source_namespace
 
+    - name: Install user level sso
+      import_playbook: install_user_rhsso.yml
```
dimitraz: This has to be moved to the end since nested plays are not supported

dimitraz:
Getting the following error trying to run the upgrade playbook: 

```
"hostvars['EVAL_VARS']" is undefined
```

https://github.com/integr8ly/installation/blob/0502b670200556ef6e579706f84da2885a900138/playbooks/user.yaml#L39

david-martin:
ON HOLD until verified on OSD

maleck13:
Will merge now OSD test will still happen

PR #748, opened 2019-07-16
mdc support

odra:
## Additional Information

Adds MDC into the installer.

### Installation
Run the install playbook using the mdc var:

```
ansible-playbook \
-i inventories/hosts \
-e 'mdc=true' \
playbooks/install.yml 
```
Retrieve mdc url:

```
echo https://`oc get route/mdc-mdc-proxy -o jsonpath='{.spec.host}'  -n mobile-developer-console`
```

Login into mdc and create a client.

### Uninstallation

Run the uninstall playbook

```
ansible-playbook \
-i inventories/hosts \
-e 'mdc=true' \
playbooks/uninstall.yml
```

SSH into the master node and and check if he mdc cors data was removed from the master config file:

```
vi /etc/origin/master/master-config.yaml
```

Search for "cors", the mdc url should not be there.

- [ ] Tested Installation
- [ ] Tested Uninstallation


wei-lee:
👀

playbooks/install_services.yml
```diff
@@ -185,6 +185,14 @@
       tags: ['ups']
       when: ups | default(true) | bool
 
+    - name: Install mobile developer console (mdc)
+      include_role:
+        name: mdc
+      vars:
+        openshift_master_url: "{{ hostvars['EVAL_VARS']['openshift_master_url'] }}"
+      tags: ['mdc']
+      when: mdc | default(true) | bool
```
wei-lee: this playbook as well as the `install_mdc` playbook are both included in the `install` playbook. Do we actually need this part here?

wei-lee: actually, never mind. looks like the `install_mdc` one is only running the `cors` role. any reason we keep them in separate playbooks?

odra: Because mdc route needs to be available so we can retrieve it, the same pattern is used in the webapp role (install_webapp playbook serves the same purpose)

roles/customisation/tasks/main.yaml
```diff
@@ -274,6 +274,24 @@
...
+        mdc_route: "https://{{mdc_route_cmd.stdout}}"
+    - name: Set Mobile Developer Console component
+      set_fact:
+        mdc_manifest:
+          - name: Mobile Developer Console
+            version: "{{ mdc_release_tag }}"
+            host: "{{ mdc_route }}"
+    - set_fact:
+        component_manifests: "{{ component_manifests }} + {{ mdc_manifest }}"
+  when: mdc
```
wei-lee: Originally my idea is to have the MDC installation playbook to generate a separate manifest file to only include information about the mobile-specific services (like user-sso, push, mss) etc. But given that we already have a manifest file that containers everything, then I don't see a problem of using the same generated manifest file. 

However, to get it work with mdc itself, there are a few small updates will be required.  
1. Can we add a `mobile:true` attribute to the specific services (user-sso, push and mss) so that mdc can filter the services using this attribute?
2. The secret should be created in the mdc namespace as well, and mounted to the pod in a path, and set that path as an env var for the pod. See https://github.com/aerogear/mobile-developer-console/blob/master/server/server.js.
3. We also need to update the server.js file of MDC to filter the services using the new attribute and follow the real format of the manifest file. I will take care of this change.

odra: :+1: will add those

playbooks/install_services.yml
wei-lee: ah, yes, understood. Thanks for the explanation. I forgot about that the `cors` update has to run on the master host.

odra:
@wei-lee I applied some changes, including the ones we talked about, mind reviewing it again?

roles/customisation/tasks/main.yaml
```diff
@@ -324,6 +345,21 @@
...
+      register: manifest_secret
+      failed_when: manifest_secret.stderr != '' and 'NotFound' not in manifest_secret.stderr
+
+    - name: Delete secret if it exists
+      shell: oc delete secret manifest -n {{ mdc_namespace }}
+      when: manifest_secret.stderr == ""
+
+    - name: Create manifest secret
+      shell: oc create secret generic manifest --from-file=/tmp/generated_manifest --type=application/json -n {{ mdc_namespace }}
+  when: mdc
```
wei-lee: we also need to mount the secret to the mdc pod. where is the best place to do it? here or in the operator?

odra: The opearator should do it, and the installer should specify the secret name into its cr

odra: I can add it into the cr right now so there won't be any changes in the installer

odra: example:
```
apiVersion: mdc.aerogear.org/v1alpha1
kind: MobileDeveloperConsole
metadata:
  name: {{ mdc_name }}
spec:
  oAuthClientId: {{ mdc_oauth_client_id }}
  oAuthClientSecret: {{ mdc_oauth_client_secret }}
  manifestName: {{ mdc_manifest_secret_name }}
```

wei-lee: ok, sounds good. This then just need to patch the CR.

PR #761, opened 2019-07-22
INTLY -2359 add blackboxtargets

StevenTobin:
## Verification Steps

1. Install from branch
2. Open prometheus route and search for `probe_` metrics
3. Ensure there are metrics for each service specified

1. Install 1.4.1
2. checkout this branch and run upgrade/yml
3. Open prometheus route and search for `probe_` metrics
4. Ensure there are metrics for each service specified

- [x] Tested Installation
- [x] Tested Uninstallation

### DIFF
--- a/roles/middleware_monitoring_config/tasks/get_blackbox_routes.yml
+++ b/roles/middleware_monitoring_config/tasks/get_blackbox_routes.yml
@@ -1,0 +1,54 @@
+---
+- name: get solution explorer routes for blackbox targets
+  block:
+  - name: get solution explorer route
+    shell: oc get route tutorial-web-app -o template --template \{\{.spec.host\}\} -n {{ eval_webapp_namespace }}
+    register: webapp_tutorial_web_app_route
+  when: webapp | bool
+
+- name: get launcher routes for blackbox targets
+  block:
+  - name: get launcher route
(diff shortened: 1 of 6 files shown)


roles/middleware_monitoring_config/tasks/get_blackbox_routes.yml
```diff
@@ -0,0 +1,54 @@
+---
+- name: get solution explorer routes for blackbox targets
+  block:
+  - name: get solution explorer route
+    shell: oc get route tutorial-web-app -o template --template \{\{.spec.host\}\} -n {{ eval_webapp_namespace }}
```
davidffrench: @StevenTobin Are the route names themselves stored in the installation repo anywhere? `tutorial-web-app`

StevenTobin: Most aren't, `tutorial-web-app` specifically has a var defined in `roles/webapp/defaults/main.yml` but every other route doesn't.

There are vars like `launcher_sso_prefix` which could be used but they don't specifically refer to the route name just happen to have the same value, and they still don't cover each route.

davidffrench: ok, thanks @StevenTobin . Since these route names are defined outside of the installer, there isn't much we can do here then.

PR #766, opened 2019-07-24
add cluster_type and update webapp

pb82:
Adds the cluster_type env var to the webapp

Verification steps:

1. install from this branch
2. once installed, make sure that:
2.1. webapp operator is version 0.0.22
2.2. webapp is version 2.12.2
3. open the `config.js` endpoint of the webapp (`<webapp route>/config.js`) and make sure there is a `clusterType` in the result and it has a value).

### DIFF
--- a/inventories/group_vars/all/manifest.yaml
+++ b/inventories/group_vars/all/manifest.yaml
@@ -1,5 +1,7 @@
 ---
 integreatly_version: master
+cluster_type: rhpds
+
 # threescale version is controlled by the operator. Changing the operator version will change the 3scale version
 
 # controls whether threescale is installed or not.
@@ -73,8 +75,8 @@ gitea_operator_resources: 'https://raw.githubusercontent.com/integr8ly/gitea-ope
 #controls whether webapp is installed or not
 webapp: true
 #below vars are not currently used but reflect the current state. In the future they will be used when we source resources from outside of the installer
-webapp_version: '2.10.5'
-webapp_operator_release_tag: 'v0.0.21'
+webapp_version: '2.12.2'
+webapp_operator_release_tag: 'v0.0.22'
 webapp_operator_resources: 'https://raw.githubusercontent.com/integr8ly/tutorial-web-app-operator/{{webapp_operator_release_tag}}/deploy'
 
 #controls whether apicurito is installed or not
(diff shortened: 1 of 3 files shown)


inventories/group_vars/all/manifest.yaml
```diff
@@ -1,5 +1,7 @@
 ---
 integreatly_version: master
+cluster_type: rhpds
```
pb82: @StevenTobin @aidenkeating mentioned that you already use a variable with the cluster type in the ansible tower job. Does it make sense to add it to the manifest like this?

StevenTobin: The cluster_type var is only used during the cluster provisioning jobs. Right now the only vars that are used are either `dev` or `poc`. These vars are passed in to tower long before the integreatly vars are sourced into an inventory so I don't see an issue with there being a cluster_type var here. Although it will need to be overwritten in the managed.template to be osd or not-rhpds or whatever it will be.

aidenkeating: Could we add a comment here with the current possible values that the web app will support?

pb82: @StevenTobin @aidenkeating added the override to managed.template and a comment with the possible values

pb82:
Verified on RHPDS:

```
$ curl https://tutorial-web-app-webapp.apps.pbraun-3867.openshiftworkshop.com/config.js --insecure
```

returns:

```json
window.OPENSHIFT_CONFIG = {
    clientId: 'tutorial-web-app',
    accessTokenUri: 'https://master.pbraun-3867.openshiftworkshop.com:443/oauth/token',
    authorizationUri: 'https://master.pbraun-3867.openshiftworkshop.com:443/oauth/authorize',
    redirectUri: 'https://tutorial-web-app-webapp.apps.pbraun-3867.openshiftworkshop.com/oauth/callback',
    scopes: ['user:full'],
    masterUri: 'https://master.pbraun-3867.openshiftworkshop.com:443',
    wssMasterUri: 'wss://master.pbraun-3867.openshiftworkshop.com:443',
    ssoLogoutUri: 'https://sso-sso.apps.pbraun-3867.openshiftworkshop.com/auth/realms/openshift/protocol/openid-connect/logout?redirect_uri=https://tutorial-web-app-webapp.apps.pbraun-3867.openshiftworkshop.com',
    threescaleWildcardDomain: '',
    integreatlyVersion: 'master',
    clusterType: 'rhpds'
  };
```

PR #769, opened 2019-07-25
INTLY-602 Move monitoring stack install earlier in process to allow f…

david-martin:
…or monitoring resources being created by product installs

## Additional Information
This is required for upcoming changes (no PR until GAs) for fuse & amq online where they both create monitoring resources during their install.
Without the monitoring stack being installed beforehand, the various monitoring CRDs will not exist e.g. GrafanaDashboard.

## Verification Steps

1. Install should complete as before. No changes to end state from before.
1. Uninstall should complete as before.

No. This is an install time ordering thing only.

- [x] Tested Installation
- [x] Tested Uninstallation


david-martin:
@pb82 rebased with fix from master

pb82:
@david-martin running the installer now

PR #770, opened 2019-07-25
using user_rhsso flag

odra:
## Verification Steps
 Runing the installation without the "user_rhsso" flag should not break the installation

- [ ] Tested Installation
- [ ] Tested Uninstallation


david-martin:
👀

PR #774, opened 2019-07-26
docs links for mdc

laurafitzgerald:
## Additional Information
https://issues.jboss.org/browse/INTLY-2775

## Verification Steps
**Pre-req**
- cluster on pds
- Set inventories file with the following

Get `master node` using `oc get nodes | grep master`

Add hosts file at `inventories/hosts`
```
[local:vars]
ansible_connection=local

[local]
127.0.0.1

[OSEv3:children]
master

[OSEv3:vars]
ansible_user=ec2-user

[master]
<master node>
```

**Verify Install** 

`ansible-playbook -i inventories/hosts playbooks/install.yml -e github_client_id=<github_client_id> -e github_client_secret=<github_client_secret> -e eval_self_signed_certs=true -e mdc=true`

Verify that the env vars are configured with the correct urls. 
The version of mdc which uses these env vars is latest. You can update the image tag to also verify that the docs links work inside the UI, but this would already have been verified as part of [this pr](https://github.com/aerogear/mobile-developer-console/pull/339)

I don't believe there is an upgrade task here as there would be no installations of mdc since the last release of integr8ly. @odra feel free to correct me here.

- [ ] Tested Installation
- [ ] Tested Uninstallation

### DIFF
--- a/inventories/group_vars/all/manifest.yaml
+++ b/inventories/group_vars/all/manifest.yaml
@@ -118,7 +118,7 @@ application_metrics: false
 #mobile developer console
 mdc: false
 mdc_operator_release_tag: 'master'
-mdc_operator_resources: 'https://raw.githubusercontent.com/aerogear/mobile-developer-console-operator/{{mdc_operator_release_tag}}/deploy'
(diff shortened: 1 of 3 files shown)


aliok:
Verified on Laura's cluster. I saw the env vars set properly.

PR #777, opened 2019-07-26
add monitoring for mdc

wei-lee:
Add ServiceMonitor, PrometheusRules resources for MDC.

This is being provisioned to this cluster: https://master.demo27-8353.openshiftworkshop.com/console/projects.

When it is finished (should be in about 30 minutes), you should be able to login to the cluster and check the Prometheus rules and metrics, and also there should be a dashboard for the mdc operator.

@grdryn @aliok @odra If any of you can take a look in about 30 minutes to verify the PR, that will be great!

- [ ] Tested Installation
- [ ] Tested Uninstallation

### DIFF
--- a/roles/mdc/tasks/monitoring.yml
+++ b/roles/mdc/tasks/monitoring.yml
@@ -7,18 +7,18 @@
 
 - name: Copy MDC monitor templates
   template:
-    src: "{{ item }}.yml.j2"
-    dest: "{{ mdc_template_dir }}/{{ item }}.yml"
+    src: "{{ item }}.yaml.j2"
+    dest: "{{ mdc_template_dir }}/{{ item }}.yaml"
   with_items: "{{ mdc_monitor_templates }}"
 
 - name: Create MDC monitor resources
-  shell: "oc apply -f {{ mdc_template_dir }}/{{ item }}.yml -n {{ mdc_namespace }}"
+  shell: "oc apply -f {{ mdc_template_dir }}/{{ item }}.yaml -n {{ mdc_namespace }}"
   with_items: "{{ mdc_monitor_templates }}"
   register: output
   failed_when: output.stderr != '' and 'already exists' not in output.stderr
 
 - name: Delete MDC monitor temporary files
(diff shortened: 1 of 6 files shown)


roles/mdc/tasks/monitoring.yml
```diff
@@ -0,0 +1,24 @@
+---
+- name: Create Service Monitor resource
+  shell: "oc apply -f {{ item }} -n {{ mdc_namespace }}"
+  with_items: "{{ mdc_monitor_resources }}"
+  register: output
+  failed_when: output.stderr != '' and 'already exists' not in output.stderr
```
grdryn: If `oc apply` would ever give an `already exists` error message, then I don't understand `oc apply`. I don't see any problem with including it like this here, just pointing out that it shouldn't ever happen. If I'm right, then I think both the `register` and `failed_when` lines could be removed from both here and everywhere else that `oc apply` is used.

wei-lee:
@odra @aliok the installation job failed again on my cluster 😒. Feel free to use your own clusters to verify this issue or use this cluster.

wei-lee:
This PR is now deployed to this cluster: https://master.mdsdemo-710a.openshiftworkshop.com.

You can see the alerts here: https://prometheus-route-middleware-monitoring.apps.mdsdemo-710a.openshiftworkshop.com/alerts. There is also the Grafana dashboard for the mdc operator here: https://grafana-route-middleware-monitoring.apps.mdsdemo-710a.openshiftworkshop.com/d/s5vWZgvWz/mdc-operator?refresh=10s&orgId=1.

roles/mdc/tasks/monitoring.yml
odra: apply wouldn't throw a create error as it would just update the resource if the local file differs from what is in the cluster

odra:
:eyes:

grdryn:
Thanks for checking this @odra. Since @wei-lee is away today, I'll merge this and rebase #778 on top of it

PR #778, opened 2019-07-26
Use release tags for RHMDS components

grdryn:
This depends on the following PRs (and the tags that will be created as a result of them):
https://github.com/aerogear/mobile-security-service-operator/pull/156
https://github.com/aerogear/mobile-developer-console-operator/pull/26

### DIFF
--- a/inventories/group_vars/all/manifest.yaml
+++ b/inventories/group_vars/all/manifest.yaml
@@ -95,7 +95,7 @@ datasync_template_tag: '0.6.1'
 
 #controls whether the mobile security service is installed or not
 mobile_security_service: false
-mobile_security_service_operator_release_tag: 'master'
+mobile_security_service_operator_release_tag: '0.3.0'
 mobile_security_service_operator_resources: 'https://raw.githubusercontent.com/aerogear/mobile-security-service-operator/{{mobile_security_service_operator_release_tag}}/deploy'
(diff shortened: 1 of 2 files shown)


odra:
needs a rebase due to latest changes in marter branch

inventories/group_vars/all/manifest.yaml
```diff
@@ -117,10 +117,10 @@ application_metrics: false
 
 #mobile developer console
 mdc: false
-mdc_operator_release_tag: 'master'
+mdc_operator_release_tag: '0.1.0'
```
grdryn: Everything in this PR deployed as expected, except the operator deployment for mdc still uses the master image `quay.io/aerogear/mobile-developer-console-operator:master`. I'll investigate that after lunch

odra: ah the operator.yaml.j2 file is harcoding the tag instead of using this var, can you change it?

roles/mdc/templates/operator.yaml.j2
```diff
@@ -15,7 +15,7 @@ spec:
       serviceAccountName: mobile-developer-console-operator
       containers:
         - name: mobile-developer-console-operator
-          image: quay.io/aerogear/mobile-developer-console-operator:master
+          image: quay.io/aerogear/mobile-developer-console-operator:{{ mdc_operator_release_tag }}
```
odra: can you use the operator image var instead of just the tag?

odra: `mdc_operator_image`

PR #779, opened 2019-07-26
general upgrade fixes

odra:
## Additional Information

Fixes mdc and msb upgrade issues

## Verification Steps
* Install integretly from 1.4.1 release;
* Run the following upgrade command:

```
ansible-playbook \
-i inventories/hosts \
-e 'user_rhsso=true' \
-e 'mdc=true' \
-e 'backup_restore_install=true' \
-e 'aws_s3_backup_secret_name=s3-credentials' \
playbooks/upgrades/upgrade.yaml  \
```

Note: remove `-e 'eval_self_signed_certs=true'`if you are not using a rhpds openshift workshop;

* Get the mdc route by running: `oc get route/mdc-mdc-proxy -n mobile-developer-console` and access its secure url;
* Create and delete a mobile client, both operations should work with no errors. 

- [ ] Tested Installation
- [ ] Tested Uninstallation

### DIFF
--- a/playbooks/upgrades/upgrade.yaml
+++ b/playbooks/upgrades/upgrade.yaml
@@ -135,3 +135,27 @@
 - import_playbook: install_user_rhsso.yml
 #Update product version (should always be last)
 - import_playbook: "../generate-customisation-inventory.yml"
+
+- hosts: master
+  gather_facts: no
+  tasks:
+    - include_role:
+        name: mdc
+        tasks_from: cors.yml
+      vars:
+        openshift_master_url: "{{ hostvars['EVAL_VARS']['openshift_master_url'] | replace('https://', '') }}"
+      when: mdc
+    - name: restart openshift master services
+      shell: "/usr/local/bin/master-restart {{ item }}"
+      become: yes
+      when: mdc
+      with_items:
+        - api
+        - controllers
+    - name: check if openshift master services are running
+      shell: oc get nodes
+      register: task_result
+      until: task_result.rc == 0
+      retries: 6
+      delay: 5
+      failed_when: task_result.rc != 0
--- a/roles/msbroker/tasks/apply_msbroker_template.yml
+++ b/roles/msbroker/tasks/apply_msbroker_template.yml
@@ -44,7 +44,7 @@
 
 # Used to pull images from registry.redhat.io
 - name: Expose vars
-  include_vars: "../roles/imagestream_pull_secret/defaults/main.yml"
+  include_vars: "{{role_path}}/../imagestream_pull_secret/defaults/main.yml"
 - include_role:
     name: imagestream_pull_secret
   vars:


playbooks/upgrades/upgrade.yaml
```diff
@@ -135,3 +135,27 @@
...
+- hosts: master
+  gather_facts: no
+  tasks:
+    - include_role:
+        name: mdc
+        tasks_from: cors.yml
+      vars:
+        openshift_master_url: "{{ hostvars['EVAL_VARS']['openshift_master_url'] | replace('https://', '') }}"
+      when: mdc
+    - name: restart openshift master services
```
davidffrench: @odra Will restarting the openshift master services cause downtime for the customer? is there anything to consider with this type of restart on a running production cluster?

odra: yes there is a tiny downtime but there is no other way around this, mdc won't work unless we restart those

odra:
what are the errors of those services? is it because the related crds are not there?

davidffrench:
@odra sorry, I didn't get an email with your response. It is just a 404 on all those endpoints. so for example, one of the endpoints is https://master.dffrench-6ca9.openshiftworkshop.com/apis/mobile-security-service.aerogear.org/v1alpha1/namespaces/mobile-developer-console/mobilesecurityserviceapps

Please feel free to look at my cluster. https://mdc-mdc-proxy-mobile-developer-console.apps.dffrench-6ca9.openshiftworkshop.com/mobileclient/aaaa

odra:
The errors make sense because neither push or mss is instaled, so the cluster has no crds for those projects (such as androidvariants) which is why a 404 error is shown.

The upgrade would need to be executed again adding those products (ups=true and mobile_security_service=true) for mdc to now show an error.

davidffrench:
@odra I was chatting to @maleck13 about this PR. I was asking him about the restart of the OpenShift master services, he advised that all upgrades are run through tower and the master restart would require ssh access so would need to be a manual step by the CS SRE team. 

More importantly, though, he advised that the upgrade playbook is for production clusters, RHMI is not upgraded anywhere else. Until it is decided that mobile is to be added to production clusters, we need to hold off on adding any mobile specific code to the upgrade playbook.

odra:
ok but all mobile products were added there - should we remove them?

odra:
we have also been adding mobile restoration sops.. should those still be added?

laurafitzgerald:
@davidffrench we've added flags to installation and upgrades for mobile. They will currently only run if the flags are passed.

davidffrench:
Sorry for the late update on this. Craig is talking to Wei and Joe about the inclusion of MDS in RHMI production clusters for the upcoming release. Please chat to Wei about what should be done moving forward.

PR #786, opened 2019-07-30
update monitoring operator and label additional scrape configs

pb82:
Upgrade the monitoring operator to add support for multiple label values and reconciling the additional scrape configs.

Verification steps:

1. Try an install and an upgrade from this branch
2. Make sure all the pods in the monitoring workspace are running afterwards
3. Make sure that:
3.1. application monitoring operator is version 0.0.21
3.2. there are no rbac errors in the prometheus instance
3.3. The secret `integreatly-additional-scrape-configs` should have a label `monitoring-key: middleware`

### DIFF
--- a/roles/middleware_monitoring/tasks/upgrade/trigger.yml
+++ b/roles/middleware_monitoring/tasks/upgrade/trigger.yml
@@ -12,6 +12,14 @@
   changed_when: aom_role_apply_cmd.rc == 0
   with_items: "{{ monitoring_resources }}"
 
+- name: Upgrade the prometheus operator roles
+  include: ./apply_resource_from_template.yml
+  with_items:
+    - "prometheus_operator_cluster_role.yml"
+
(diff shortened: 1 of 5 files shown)


roles/middleware_monitoring/tasks/upgrade/trigger.yml
```diff
@@ -12,6 +12,14 @@
   changed_when: aom_role_apply_cmd.rc == 0
   with_items: "{{ monitoring_resources }}"
 
+- name: Upgrade the prometheus operator roles
+  include: ./apply_resource_from_template.yml
+  with_items:
+    - "prometheus_cluster_role.yml"
```
StevenTobin: Shouldn't this be `prometheus_operator_cluster_role.yml`?

pb82: Yes it should, updating

PR #787, opened 2019-07-30
wait for rhsso to be ready if rhsso_provision_immediately is set to true

wei-lee:
https://issues.jboss.org/browse/AEROGEAR-9649

- [x] Tested Installation
- [ ] Tested Uninstallation

### DIFF
--- a/roles/rhsso/tasks/install_sso.yml
+++ b/roles/rhsso/tasks/install_sso.yml
@@ -25,3 +25,13 @@
   shell: oc create -f /tmp/keycloak.json -n {{ sso_namespace }}
   register: rhsso_keycloak
   failed_when: rhsso_keycloak.stderr != '' and 'AlreadyExists' not in rhsso_keycloak.stderr
+
+- name: "Wait for keycloak to be ready"
+  shell: "oc get keycloak/rhsso -o jsonpath='{.status.phase}' -n {{ sso_namespace }}"
+  register: result
+  until: result.stdout == 'reconcile'
+  retries: 50
+  delay: 10
+  failed_when: result.stderr
+  changed_when: False
+  when: rhsso_provision_immediately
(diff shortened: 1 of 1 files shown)


wei-lee:
The fix is installed in this cluster: https://master.mdsdemo-710a.openshiftworkshop.com/console

roles/rhsso/tasks/install_sso.yml
```diff
@@ -25,3 +25,13 @@
   shell: oc create -f /tmp/keycloak.json -n {{ sso_namespace }}
   register: rhsso_keycloak
   failed_when: rhsso_keycloak.stderr != '' and 'AlreadyExists' not in rhsso_keycloak.stderr
+
+- name: "Wait for keycloak to be ready"
+  shell: "oc get keycloak/rhsso -o jsonpath='{.status.phase}' -n {{ sso_namespace }}"
+  register: result
+  until: result.stdout == 'reconcile'
```
grdryn: Is this the final status phase? I just notice that there's also e.g. a 'complete' phase string here: https://github.com/integr8ly/keycloak-operator/blob/master/pkg/apis/aerogear/v1alpha1/types.go#L270..L290

roles/rhsso/tasks/install_sso.yml
```diff
@@ -25,3 +25,13 @@
...
+  retries: 50
+  delay: 10
+  failed_when: result.stderr
```
grdryn: Would it ever be useful to retry even after a message on stderr (it will fail eventually when it runs out of retries anyway)?

roles/rhsso/tasks/install_sso.yml
wei-lee: yes, I think so.

roles/rhsso/tasks/install_sso.yml
wei-lee: good point. will remove it.

wei-lee:
@odra mind approving this PR? thanks

PR #788, opened 2019-07-31
mobile playbooks

odra:
## Additional Information

Refactors mobile installation into its own platbook(s).

## Verification Steps

* run the install playbook
* check if any mobile specific product was installed
* run the install_mobile playbook
* check if all mobile products were properly installed
* run the install_mobile playbook and check if something went wrong (if mdc login stopped working for example)
* run the uninstall_mobile playbook and check if all mobile products were properly removed
* run the uninstall playbook

NOTE: a first time installation can also be achieved by running the usual install playbook adding the mobile flag: `-e mobile=true`

- [ ] Tested Installation
- [ ] Tested Uninstallation


odra:
ping @wei-lee

playbooks/install.yml
```diff
@@ -8,15 +8,13 @@
...
 - import_playbook: "./install_fuse_managed.yml"
 - import_playbook: "./install_services.yml"
-- import_playbook: "./install_mdc.yml"
 - import_playbook: "./install_webapp.yml"
 - import_playbook: "./install_middleware_monitoring_config.yml"
 - import_playbook: "./install_application_metrics.yml"
 - import_playbook: "./install_user_rhsso.yml"
 - import_playbook: "./generate-customisation-inventory.yml"
 - import_playbook: "./customise_web_console_install.yml"
 - import_playbook: "./apply_ocp_roles.yml"
```
wei-lee: does this mean that we now need to run 2 playbooks to have mobile services installed?

odra: if you are not using a integratly workshop yes, isn't the whole point of this  change to split the mobile stuff in a distinct playbook (the same way we do with backups)?

wei-lee: ok, so are we going to modify provision steps for the integreatly workshop on RHPDS to run both playbooks? where is that defined?

odra: it is another repository managed by the pds team, @pmccarthy may know more about it

wei-lee: ok, what if we include the mobile playbook as part of the default installation one? This way we don't need to modify the other repo, but the mobile playbook can still be ran on its own if required?

odra: I added it in the "install.yml" playbook

odra:
@wei-lee I have updated my pr, mind reviewing it again

wei-lee:
@odra just got time to try this and there are 2 problems I come across:

1. The first time when I run the playbook on an existing cluster, I got a permission error about trying to update an existing `/tmp/mdc.json` file. Looks like this file was created from a previous run and didn't clean up properly. After I manually removed it everything works like a dream, and it reduce my provision time to less than 5 minutes! (at least for just mobile services).
2. After provision, when I try to login to MDC, I still get a login error. I checked the deployment config and looks like the oauthproxy is using the wrong client secret. After I manually fix it in the deployment config, everything works fine. Not sure if this issue is already fixed on master.

Anyway, that's only 2 issues I found. Great work @odra.

odra:
was there an existing oauth proxy? playbook will skip the oauth bits if there is one already there

wei-lee:
@odra I think there wasn't an existing oauth proxy (as the mdc deployment was removed), however, there should still be an OAuthClient in place. However, I don't know why the oauthproxy is picking the wrong client secret if the oauthclient is already there.

wei-lee:
@odra going to merge this now, are you ok with that?

odra:
I was going to test it again to see if I got the authclient problem but we can merge it as it is and open a new jira if that shows up again

PR #794, opened 2019-08-01
INTLY-2763 add some delay and retries when getting user-sso route

tremes:
## Additional Information
This is just an attempt to fix the issue which is intermittent and probably just timing issue

## Verification Steps
Try to run the installation probably multiple times

No

- [ ] Tested Installation
- [ ] Tested Uninstallation


trepel:
I think the https://github.com/integr8ly/installation/pull/787 solves the same issue, although the solution is completely different. This PR has been merged, not sure if we need to merge this one too or just close this one.

tremes:
OK. Feel free to close it then.

wei-lee:
I am pretty sure this is solved in #787. So going to close this.

PR #796, opened 2019-08-01
use variable for secret name retrieved from the openshift namespace

JameelB:
## Additional Information
Replace hardcoded secret name `imagestreamsecret` with `{{ pull_secret_name }}` as this is overridden during OSD installation. A separate variable `{{ product_ns_pull_secret_name }}` should be used in the template to allow roles, like Fuse, to define this secret name that will be created in the product namespace.

**NOTE**: 
- The fuse operator requires this secret to be named `syndesis-pull-secret`

## Verification Steps
<!--

1. Run the installation
2. Ensure you can go through Walkthrough 1A successfully

Upgrade task has been updated and needs to be tested. Ensure that you can go through Walkthrough 1A Successfully after upgrade.

- [ ] Tested Installation
- [ ] Tested Uninstallation

### DIFF
--- a/roles/3scale/tasks/install.yml
+++ b/roles/3scale/tasks/install.yml
@@ -18,7 +18,7 @@
     namespace: "{{ threescale_namespace }}"
 
 - name: Link imagestream pull secret for images
-  shell: oc secrets link default {{ pull_secret_name }} --for=pull -n {{ threescale_namespace }}
+  shell: oc secrets link default {{ product_ns_pull_secret_name }} --for=pull -n {{ threescale_namespace }}
 
 - name: Check namespace for existing resources
   shell: oc get all -n {{ threescale_namespace }}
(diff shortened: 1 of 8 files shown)


StevenTobin:
So this references the `{{ product_ns_pull_secret_name }} ( imagestreamsecret )` as its name https://github.com/integr8ly/installation/blob/d222c9916e48f2a00947fb39b55b9f9d47fdb1ac/roles/imagestream_pull_secret/templates/imagestream-pull-secret.yml.j2#L6 but here https://github.com/integr8ly/installation/blob/d222c9916e48f2a00947fb39b55b9f9d47fdb1ac/roles/3scale/tasks/install.yml#L21 it uses `{{ pull_secret_name }} (registry-redhat-io-whatever) ` but no secret exists called pull_secret_name because product_ns_pull_secret_name is hardcoded. I think changing the 3scale task to use product_ns_pull_secret_name would work?

JameelB:
@StevenTobin Updated the PR according to the changes requested. Mind taking another look?

StevenTobin:
Hey, I'll take a run at this this afternoon

JameelB:
@StevenTobin Thanks for verification! :grinning:

PR #802, opened 2019-08-06
Fix backups for mobile

wei-lee:


wei-lee:
@odra mind taking a quick look?

wei-lee:
@odra quick review if that's ok?

odra:
sure let me run it

PR #804, opened 2019-08-06
fix(AEROGEAR-9652) : Fix mss grafana resources dashboard

camilamacedo86:
## Additional Information
Applying changes performed here: https://github.com/aerogear/mobile-security-service-operator/pull/165/files

## Motivation:
https://issues.jboss.org/browse/AEROGEAR-9652

PR #805, opened 2019-08-06
fix(AEROGEAR-9652) : Fix mss grafana resources dashboard

camilamacedo86:
master pr is: https://github.com/integr8ly/installation/pull/804

### DIFF
--- a/roles/mobile_security_service/templates/mss_grafana_dashboard.yml.j2
+++ b/roles/mobile_security_service/templates/mss_grafana_dashboard.yml.j2
@@ -630,14 +630,14 @@ spec:
           "steppedLine": false,
           "targets": [
             {
-              "expr": "process_virtual_memory_bytes{job='mobile-security-service-operator'}",
+              "expr": "kube_pod_container_resource_limits_memory_bytes{namespace='{{ mobile_security_service_namespace }}', container='application'}",
               "format": "time_series",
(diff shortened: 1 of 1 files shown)


roles/mobile_security_service/templates/mss_grafana_dashboard.yml.j2
```diff
@@ -703,112 +703,6 @@ spec:
...
-              "expr": "(kube_pod_container_resource_limits_cpu_cores{container='mobile-security-service-operator'})*1000",
-              "format": "time_series",
-              "interval": "",
-              "intervalFactor": 2,
-              "legendFormat": "Maximum Limit of Millicores",
-              "refId": "B",
-              "step": 2
-            },
-            {
-              "expr": "(((kube_pod_container_resource_limits_cpu_cores{container='mobile-security-service-operator'})*1000)/100)*90",
```
wei-lee: are the cpu metrics supposed to be removed?

camilamacedo86: Yes. Because they are wrong before. Note that is showing the CPU of the operator in the Service Dashboard. This metric exists already in the Oper Dash.
However, for the service, I could not find any metric which would allow us to do the same. So, if we would like to provide this graph we need check how to export for the Prometheus.

PR #806, opened 2019-08-06
removing mss var from install_mobile playbook

odra:
removing mobile_security_services var from install_mobile.yml playbook (mss was not being added in the generated manifest sceret because of this var)


odra:
ping @wei-lee

PR #808, opened 2019-08-06
feat(ups grafana dash) : Upgrade it with the latest version

camilamacedo86:
## Motivation

https://issues.jboss.org/browse/AEROGEAR-9692

## What

Applying the latest version of this file in UPS Oper repo.

PR #809, opened 2019-08-06
feat(ups grafana dash) : Upgrade it with the latest version

camilamacedo86:
Ref: https://github.com/integr8ly/installation/pull/808

### DIFF
--- a/roles/ups/templates/grafana_dashboard.yml.j2
+++ b/roles/ups/templates/grafana_dashboard.yml.j2
@@ -36,15 +36,15 @@ spec:
           ],
           "annotations": {
             "list": [
-              {
-                "builtIn": 1,
-                "datasource": "-- Grafana --",
-                "enable": true,
-                "hide": true,
-                "iconColor": "rgba(0, 211, 255, 1)",
-                "name": "Annotations & Alerts",
-                "type": "dashboard"
-              }
+            {
+              "builtIn": 1,
+              "datasource": "-- Grafana --",
+              "enable": true,
+              "hide": true,
+              "iconColor": "rgba(0, 211, 255, 1)",
+              "name": "Annotations & Alerts",
+              "type": "dashboard"
+            }
             ]
           },
           "description": "Application metrics",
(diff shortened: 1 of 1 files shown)


roles/ups/templates/grafana_dashboard.yml.j2
```diff
@@ -6,6 +6,9 @@ metadata:
     monitoring-key: middleware
     prometheus: application-monitoring
 spec:
+  selector:
+    matchLabels:
+      name: {{ ups_server_name }}
```
wei-lee: I am pretty sure this part is not required.

wei-lee:
@camilamacedo86 how did you create the dashboard? Did you deploy it anywhere? This PR is hard to review at the moment. Also the JIRA wasn't really brought in for the sprint. Do we really need this version of dashboard? What things are missing/broken in the current one?

roles/ups/templates/grafana_dashboard.yml.j2
camilamacedo86: Oks we can remove it so.

camilamacedo86:
HI @wei-lee, 

Note that it was created 22 days ago here in the installer and the latest changes performed in the UPS on it was 15 days ago.  So, it cannot be the same content.

Why do we need to update it?
- Will be very hard applying the next requests/fixes and changes if it is not the same version
- I believe that the version of 15 days ago has bug fixes and addressing the comments made by you folks in the review. So, is really valid we spend time comparing the both? Is not better we merge it and then check the result? Because as you said will hard validate it another way now. I think that it was added here before we finish the tasks regards the dashboards there.

wei-lee:
@camilamacedo86 but do you know exactly what this PR fixes? Or can I see it somewhere. Just because the version is 15 days old doesn't mean there are problems with it.

camilamacedo86:
Hi @wei-lee,

The changes was made in the PR: https://github.com/aerogear/unifiedpush-operator/commit/bd59a360a700a3d2cde9eb5c1b1f4daf9dabcac2#diff-31f10ce7bb67936fff50032371bb0bec. Note that here we addressed all comments, suggestions and requested made regards to it in the review of this graph. 

We changed many things will be hard we check all in the same way that will be hard we apply the other task do replace the % for an upside/down as any other one without keeping them both using the same versions.

roles/ups/templates/grafana_dashboard.yml.j2
```diff
@@ -36,15 +36,15 @@ spec:
...
-              }
+            {
+              "builtIn": 1,
+              "datasource": "-- Grafana --",
+              "enable": true,
+              "hide": true,
+              "iconColor": "rgba(0, 211, 255, 1)",
+              "name": "Annotations & Alerts",
+              "type": "dashboard"
+            }
```
wei-lee: datasource is not required. Let's not add them.

wei-lee:
@camilamacedo86 ok. can you address my last comment and then merge this PR. We can see how it works after provisioned to a cluster.

roles/ups/templates/grafana_dashboard.yml.j2
wei-lee: nvm. Wrong section.

camilamacedo86: 👍

PR #816, opened 2019-08-07
delete mdc first when uninstall mds

wei-lee:


wei-lee:
@odra mind taking a look?

PR #817, opened 2019-08-07
enable mobile services

wei-lee:


odra:
I am confused, isn't this a dejavu of https://github.com/integr8ly/installation/pull/812/files?

wei-lee:
@odra yes, it is, but for somereason it got revered....

PR #820, opened 2019-08-07
add mobile walkthrough if mdc is enabled

wei-lee:


wei-lee:
ping @odra

roles/webapp/tasks/provision-webapp.yml
```diff
@@ -18,6 +18,14 @@
   register: create_webapp_operator_resource_cmd
   failed_when: create_webapp_operator_resource_cmd.stderr != '' and 'AlreadyExists' not in create_webapp_operator_resource_cmd.stderr
 
+- set_fact:
+    walkthrough_locations: 
+      - "{{ webapp_default_walkthrough_location }}"
+
+- set_fact:
+    walkthrough_locations: "{{ walkthrough_locations }} + {{ webapp_mobile_walkthrough_location }}"
```
odra: I think this needs to be written as `"{{ walkthrough_locations }} + {{ [webapp_mobile_walkthrough_location] }}"` otherwise it will try to concatenate a list with a string

PR #823, opened 2019-08-08
only set the url for the mobile walkthrough

wei-lee:
only set the mobile walkthrough url in the webapp CR, because the webapp always add the core walkthrough by default. Set it again in the CR will cause errors in the webapp.


wei-lee:
ping @odra

wei-lee:
@odra mind approving this PR?

PR #822, opened 2019-08-08
split mobile playbook to 2 parts and run them at different time

wei-lee:

PR #815, opened 2019-08-07
AEROGEAR-9691 - Rename MobileClient CRD group

aliok:
JIRA: https://issues.jboss.org/browse/AEROGEAR-9691

Merging prereqs:
* https://github.com/aerogear/mobile-developer-console-operator/pull/40 is merged
* MDC operator image is created and tagged (as 0.1.8)

PR #825, opened 2019-08-08
Add RHMDS services to BlackboxTarget

grdryn:
## Additional Information

This is marked as a 'Draft' since it's blocked until a fix in #822 is merged.

- [ ] Tested Installation
- [ ] Tested Uninstallation

### DIFF
--- a/inventories/group_vars/all/manifest.yaml
+++ b/inventories/group_vars/all/manifest.yaml
@@ -117,10 +117,10 @@ application_metrics: false
 
 #mobile developer console
 mdc: true
-mdc_operator_release_tag: '0.1.7'
+mdc_operator_release_tag: '0.1.9'
 mdc_operator_resources: 'https://raw.githubusercontent.com/aerogear/mobile-developer-console-operator/{{ mdc_operator_release_tag }}/deploy'
 mdc_operator_image: 'quay.io/aerogear/mobile-developer-console-operator:{{ mdc_operator_release_tag }}'
-mdc_release_tag: '1.1.3'
+mdc_release_tag: '1.1.4'
 mdc_image: 'quay.io/aerogear/mobile-developer-console:{{ mdc_release_tag }}'
 mdc_proxy_release_tag: 'v1.1.0'
 mdc_proxy_image: 'docker.io/openshift/oauth-proxy:{{ mdc_proxy_release_tag }}'
--- a/roles/customisation/templates/inventory.j2
+++ b/roles/customisation/templates/inventory.j2
@@ -115,9 +115,3 @@ apicurito_namespace="{{eval_apicurito_namespace}}"
 apicurito_route="{{apicurito_secure_ui_route}}"
 
 {% endif %}
-
-{% if mobile_security_service %}
-mss_namespace="{{ eval_mobile_security_service_namespace }}"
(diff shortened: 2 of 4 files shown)


inventories/group_vars/all/manifest.yaml
```diff
@@ -117,10 +117,10 @@ application_metrics: false
 
 #mobile developer console
 mdc: true
-mdc_operator_release_tag: '0.1.7'
+mdc_operator_release_tag: '0.1.9'
 mdc_operator_resources: 'https://raw.githubusercontent.com/aerogear/mobile-developer-console-operator/{{ mdc_operator_release_tag }}/deploy'
 mdc_operator_image: 'quay.io/aerogear/mobile-developer-console-operator:{{ mdc_operator_release_tag }}'
-mdc_release_tag: '1.1.3'
+mdc_release_tag: '1.1.4'
```
grdryn: This now replaces #815

roles/customisation/templates/inventory.j2
```diff
@@ -115,9 +115,3 @@ apicurito_namespace="{{eval_apicurito_namespace}}"
 apicurito_route="{{apicurito_secure_ui_route}}"
 
 {% endif %}
-
-{% if mobile_security_service %}
-mss_namespace="{{ eval_mobile_security_service_namespace }}"
-mss_route="{{ mss_route }}"
```
grdryn: @odra I'm not sure if it's correct to remove this stanza, but the setting of this `mss_route` was removed from the task file in a different PR recently, causing the installation to fail. I'm not really sure what this role is for though, so I'm not sure if it's correct to remove this, or if I should put the missing tasks back. Thoughts?

roles/middleware_monitoring_config/templates/blackboxtargets.yml.j2
```diff
@@ -51,4 +51,22 @@ spec:
...
+{% endif %}
+{% if ups %}
+  # Unified Push Server targets
+  - service: ups-ui
+    url: https://{{ ups_ups_route.stdout }}/rest/auth/config/
+    module: http_2xx
+{% endif %}
+{% if mobile_security_service %}
+  # Mobile Security Service targets
+  - service: ups-ui
```
pb82: this should have a different name than `ups-ui`. Otherwise you'll get the multi series error

roles/customisation/templates/inventory.j2
odra: will it be the only service that is not going to store its route here?

grdryn: @odra Like I said, I don't really know enough about integreatly to know what this role is for. I don't see either of UPS or MDC routes stored here, so my guess is "no, it's not the only service without a route here", but it might be a bug that those other two aren't there. WDYT?

odra: technically all services should, but since the other mobile services arent`t so lets remove it and properly add them all in another pr

odra: my vm got deleted so I requested a new one.. won't be able to test it for 20 mins or so.

odra: ok vm ready, givint it a spin

PR #826, opened 2019-08-08
uninstall mobile services as part of the overall uninstall

wei-lee:


wei-lee:
ping @odra @psturc

odra:
giving it a try

PR #836, opened 2019-08-12
feat(mdc) : Add MDC Application Dash

camilamacedo86:
## Additional Information
https://issues.jboss.org/browse/AEROGEAR-9650

## What
Re-applying changes since it was missed in some rebase with the master in this branch.

### DIFF
--- a/roles/mdc/defaults/main.yml
+++ b/roles/mdc/defaults/main.yml
@@ -29,3 +29,4 @@ mdc_monitor_templates:
   - operator_grafana
   - operator_prometheus_rules
   - mdc_prometheus_rules
+  - mdc_grafana_dashboard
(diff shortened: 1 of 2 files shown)


wei-lee:
@camilamacedo86 I am not sure why it is show up but this PR is already landed in the feature branch. See https://github.com/integr8ly/installation/blob/add-mobile-services/roles/mdc/defaults/main.yml

wei-lee:
This PR is in the feature branch anyway, so I am going to close this.

camilamacedo86:
Hi @wei-lee but if it was fully there, so why it shows the differences between both?
Could you check the changes files?

grdryn:
@camilamacedo86 I believe that might be because your branch is stale (it's based on an old master, rather than the add-mobile-services branch). If you check out your branch locally and try to rebase it on top of the origin/add-mobile-services branch, you'll notice that you just get fast-forwarded to the tip of the origin/add-mobile-services branch, and no new commit is added on top, because the changes already exist on that branch.

I guess GitHub just doesn't display the diff very well when the commits are this far apart or something. I'm not sure though.

PR #837, opened 2019-08-12
disable mobile services by default for managed clusters

wei-lee:


wei-lee:
ping @grdryn @odra

odra:
there is a `datasync` var as well

odra:
let me try it in a cluster

odra:
The vars ddn't work as all components are still being installed but no related to this pr - those checks may need to be improved

odra:
ok I see there are no usage of those vars in the install_mobile_services playbook which is why it didn't work - could you please add them there?

wei-lee:
@odra right, yes. will do

wei-lee:
@odra mind taking another look?

odra:
giving it a try again

playbooks/install_mobile_services.yml
```diff
@@ -13,24 +13,28 @@
...
 
     - name: Install mobile developer console (mdc)
       include_role:
         name: mdc
       vars:
         openshift_master_url: "{{ hostvars['EVAL_VARS']['openshift_master_url'] | replace('https://', '') }}"
       tags: ['mdc']
+      when: mdc | default(true) | bool
 
     - name: User rhsso configuration
```
odra: this one needs a mdc check as well

wei-lee: done

wei-lee:
@david-martin do you mind reviewing this PR as well and see if I am doing the feature flags correctly? Thanks.

playbooks/install_mobile_manifest.yml
```diff
@@ -126,26 +129,31 @@
...
+        - run_master_tasks | default(true) | bool 
+        - mdc | default(true) | bool
     
     - name: restart openshift master services
       shell: "/usr/local/bin/master-restart {{ item }}"
       become: yes
       with_items:
         - api
         - controllers
+      when: mdc | default(true) | bool
```
david-martin: Should `run_master_tasks | default(true) | bool` also be satisfied here too?

PR #838, opened 2019-08-12
Webapp upgrade 1 5 0

pb82:
Webapp upgrade to 1.5.0. This requires adding new fields to the CR as well as forcing the operator to create a volume.

Verification steps:

1. Start from RHPDS 1.4.1
1. Run the upgrade from this branch: `ansible-playbook -i inventories/hosts playbooks/upgrades/upgrade.yaml -e eval_self_signed_certs=true -e backup_restore_install=false`
1. After the upgrade ensure that:
1.1. The version of the webapp operator should be v0.0.24
1.2. The version of the webapp should be 2.14.0
1.3. The webapp deployment should have environment vars with `CLUSTER_TYPE`, `INTEGREATLY_VERSION` and `DATABASE_LOCATION`
1.4. There should be a volume claim named `user-walkthroughs` in the namespace

### DIFF
--- a/playbooks/upgrades/upgrade.yaml
+++ b/playbooks/upgrades/upgrade.yaml
@@ -35,15 +35,10 @@
         sso_realm: "{{ rhsso_realm }}"
 
     # Solution Explorer
-    - name: Bump the Web App operator version to {{ webapp_operator_release_tag }}
-      shell: oc patch deployment tutorial-web-app-operator -n {{ eval_webapp_namespace }} --type json -p '[{"op":"replace", "path":"/spec/template/spec/containers/0/image", "value":"{{ upgrade_webapp_operator_image }}"}]'
-      register: upgrade_webapp_operator
-      failed_when: upgrade_webapp_operator.stderr != '' and 'not patched' not in upgrade_webapp_operator.stderr
-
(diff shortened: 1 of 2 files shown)


davidffrench:
👀

playbooks/upgrades/upgrade.yaml
```diff
@@ -35,15 +35,14 @@
         sso_realm: "{{ rhsso_realm }}"
 
     # Solution Explorer
-    - name: Bump the Web App operator version to {{ webapp_operator_release_tag }}
-      shell: oc patch deployment tutorial-web-app-operator -n {{ eval_webapp_namespace }} --type json -p '[{"op":"replace", "path":"/spec/template/spec/containers/0/image", "value":"{{ upgrade_webapp_operator_image }}"}]'
-      register: upgrade_webapp_operator
-      failed_when: upgrade_webapp_operator.stderr != '' and 'not patched' not in upgrade_webapp_operator.stderr
-
-    - name: Bump the Web App deployment version to {{ webapp_version }}
```
davidffrench: @pb82 Is this not needed anymore? I see it wasn't moved to the webapp upgrade file.

pb82: No, the operator will automatically install the new version

PR #841, opened 2019-08-12
Add missing file 

camilamacedo86:
## Additional Information
https://issues.jboss.org/browse/AEROGEAR-9736

## Verification Steps
After the installation using this branch in the mobile-security-service namespace, a service for the operator should be there. Also, the Grafana Dashboard called `MSS Operator` should be working.

### DIFF
--- a/roles/mobile_security_service/defaults/main.yml
+++ b/roles/mobile_security_service/defaults/main.yml
@@ -12,6 +12,7 @@ mobile_security_service_operator_resource_items:
 
 mobile_security_service_monitoring_resource_items:
   - "{{ mobile_security_service_operator_resources }}/monitor/service_monitor.yaml"
+  - "{{ mobile_security_service_operator_resources }}/monitor/operator_service.yaml"
   - "{{ mobile_security_service_operator_resources }}/monitor/mss_service_monitor.yaml"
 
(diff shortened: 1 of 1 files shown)


wei-lee:
I am pretty sure we deliberately didn't include the `operator_service.yaml` file, because it is giving errors: during some runs, this step will produce errors complaining the service already exists (or the service is created by a different owner).

So this PR doesn't fix the problem, and also it removes the `ServiceMonitor` setup which means even if the service is created, there won't be a servicemonitor setup so that Prometheus can scrape for metrics.

camilamacedo86:
HI @wei-lee,

Following the comments inline.

> I am pretty sure we deliberately didn't include the operator_service.yaml file, because it is giving errors: during some runs, this step will produce errors complaining the service already exists (or the service is created by a different owner).
> --

* I cannot see an ay error when the file which is missing is applied just the issue solved with the integr8ly/installation. 
* The file is in called in the make command here; https://github.com/aerogear/mobile-security-service-operator/blob/master/Makefile#L93. So, if it was delivered removed so why it was not removed from the project?

camilamacedo86:
Hi @wei-lee I am closing this one and moving with your suggestion.

PR #843, opened 2019-08-13
Add mobile services

wei-lee:
Add MDS to Integreatly. Services are enabled by default so that they will work in PDS and POC, and turned off for managed environments.

ping @maleck13 @davidffrench  @david-martin


davidffrench:
@wei-lee I think there may be something weird with the git history on this. Is this branch rebased from master?

wei-lee:
@davidffrench check you check again, it should be fixed now. Thanks

davidffrench:
thats looking better, thanks @wei-lee

playbooks/generate-customisation-inventory.yml
```diff
@@ -11,9 +11,10 @@
     - include_vars: ../roles/nexus/defaults/main.yml
     - include_vars: ../roles/rhsso/defaults/main.yml
     - include_vars: ../roles/webapp/defaults/main.yml
-    - include_vars: ../roles/mdc/defaults/main.yml
+    - include_vars: ../roles/mobile_security_service/defaults/main.yml
```
davidffrench: this is here twice, also on line 16.

playbooks/install.yml
```diff
@@ -26,9 +29,6 @@
       become: yes
       when: (run_master_tasks | default(true) | bool) and (create_cluster_admin == true)
 
-- import_playbook: ./install_backups.yml
```
davidffrench: why was this removed?

playbooks/install_mobile_manifest.yml
```diff
@@ -0,0 +1,164 @@
...
+        name: mdc
+        tasks_from: patch-manifest.yml
+      vars:
+        mdc_manifest_secret_name: manifest
+      tags: ['mdc', 'remote']
+      when: 
+        - run_master_tasks | default(true) | bool 
+        - mdc | default(true) | bool
+    
+    - name: restart openshift master services
```
davidffrench: I thought this would be a manual process done by CS-SRE. What have i missed? 

cc @maleck13

roles/webapp/defaults/main.yml
```diff
@@ -11,3 +11,4 @@ webapp_operator_resource_items:
   - "{{ webapp_operator_resources }}/sa.yaml"
   - "{{ webapp_operator_resources }}/crd.yaml"
   - "{{ webapp_operator_resources }}/operator.yaml"
+webapp_mobile_walkthrough_location: 'https://github.com/aerogear/mobile-walkthrough#master'
```
davidffrench: should this always be set to master or some release tag?

playbooks/install_mobile_manifest.yml
david-martin: @davidffrench That's correct, for install on OSD.
For pds, poc, this is fine.
There's a check on the `run_master_tasks` flag below

playbooks/generate-customisation-inventory.yml
wei-lee: yes, will remove the duplicated one

playbooks/install.yml
wei-lee: I think this is a merge issue, will add it back

roles/webapp/defaults/main.yml
wei-lee: yeah, we like to use master at least for now, there will be changes coming to the walkthrough for a while I think. Will change this to a tag probably in a future release.

wei-lee:
@davidffrench your comments are addressed now, mind taking another look?

PR #845, opened 2019-08-13
check mobile services feature flags before uninstall

wei-lee:
@odra this is the right one to check. thanks.


wei-lee:
ping @grdryn

PR #846, opened 2019-08-13
INTLY-2838 Move readiness check before users setup

david-martin:
## Additional Information
https://issues.jboss.org/browse/INTLY-2838

This change should ensure all 3scale pods are running before making calls to the API by re-using the more robust `check_readiness` task

## Verification Steps

1. Do an install
2. Verify the install succeeds 

No. This is an install time change only.

- [ ] Tested Installation
- [ ] Tested Uninstallation


david-martin:
I've verified an install in pds.
Would appreciate a second set of 👀

PR #849, opened 2019-08-13
run mdc manifest patch task from localhost

wei-lee:
@odra I noticed that on OSD, the urls for services are not setup correctly, because the manifest file is not mounted to the MDC pod because the master tasks are not running. However, I don't think the tasks to patch the mdc dc needs to run from master. It can be invoked from localhost.


wei-lee:
cc @psturc

PR #850, opened 2019-08-13
support tuning of resources for 3scale/enmasse

jessesarn:
## Additional Information
https://issues.jboss.org/browse/INTLY-2653 - new resource requests/limits for enmasse
https://issues.jboss.org/browse/INTLY-2654 - new resource requests/limits for 3scale

## Verification Steps
1. Dump current request/limit values for 3scale and enmasse with (may need to modify if using a different ns_prefix other than blank or openshift-):
```
$ for i in {openshift-,}{3scale,enmasse}; do echo $i; (echo -e "pod\tqos\tcontainer\tcpu\tmem" && oc get pod -o json -n $i | jq '.items[] | {name: .metadata.name, qos: .status.qosClass} as $p | .spec.containers[] | $p + {container: .name, cpu: (.resources.requests.cpu + "-" + .resources.limits.cpu), mem: (.resources.requests.memory + "-" + .resources.limits.memory ) } | [.[]] | @tsv' -r) | column -t; done
```
2. Run the resource_limits playbook directly to apply changes to 3scale/enmasse dcs/deploys:
```
ansible-playbook -i inventories/hosts playbooks/update_resources.yml -e prerequisite_checks=false
```
3. Confirm 3scale and enmasse are still functioning properly.
4. Re-run dump from step 1 to make sure the resource ranges have been updated.

Not yet (will add logic to call this playbook during install and upgrade after next release)

- [x] Tested Installation
- [n/a] Tested Uninstallation
- [n/a] Tested or Created follow on for Upgrade

### DIFF
--- a/roles/enmasse/defaults/main.yml
+++ b/roles/enmasse/defaults/main.yml
@@ -12,3 +12,31 @@ enmasse_enable_monitoring: false
 enmasse_backup_postgres_secret: 'enmasse-postgres-secret'
 enmasse_postgres_cronjob_name: 'enmasse-postgres-backup'
 enmasse_pv_cronjob_name: 'enmasse-pv-backup'
+
+enmasse_resources:
+  - name: postgresql
+    kind: dc
+    resources:
+      requests:
+        memory: 50Mi
+  - name: api-server
+    kind: deploy
+    resources:
+      requests:
+        memory: 256Mi
+  - name: enmasse-operator
+    kind: deploy
+    resources:
+      requests:
+        memory: 75M
+  # keycloak operator undoes any externally made changes
+  # - name: keycloak
+  #   kind: deploy
+  #   resources:
+  #     requests:
+  #       memory: 750M
+  - name: service-broker
+    kind: deploy
+    resources:
+      requests:
+        memory: 200M
--- a/roles/resource_limits/tasks/patch_resource.yml
+++ b/roles/resource_limits/tasks/patch_resource.yml
@@ -11,7 +11,7 @@
 - name: Update resource values for {{ resource_patch.kind }}/{{ resource_patch.name }}
   shell: |
     oc set resources {{ resource_patch.kind }} {{ resource_patch.name }} -n {{ ns }}
-      {%- for k,v in resource_patch.resources.iteritems() %}
+      {%- for k,v in resource_patch.resources.items() %}
       --{{ k }}={% for k2,v2 in v.items() %}{{ k2 }}={{ v2 }}{% if not loop.last %},{% endif %}{% endfor %}
       {%- endfor %}
   register: set_resources_cmd
--- a/roles/resource_limits/tasks/patch_resource.yml
+++ b/roles/resource_limits/tasks/patch_resource.yml
@@ -1,0 +1,23 @@
+---
+
+#TODO Add support for multiple containers with different resource values by pausing first and resuming after all patches have been made
+
+# - name: Export resource config for {{ resource_patch.kind }}/{{ resource_patch.name }}
+#   shell: |
+#     oc get {{ resource_patch.kind }} {{ resource_patch.name }} -o yaml \
+#       -n {{ ns }} > /tmp/old_limits_{{ resource_patch.kind }}_{{ resource_patch.name }}.yaml
+#   changed_when: false
+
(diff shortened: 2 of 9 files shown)


roles/resource_limits/tasks/patch_resource.yml
```diff
@@ -0,0 +1,23 @@
+---
+
+#TODO Add support for multiple containers with different resource values by pausing first and resuming after all patches have been made
```
ciaranRoche: Can we remove this TODO and maybe create a JIRA for it?

roles/resource_limits/tasks/patch_resource.yml
```diff
@@ -0,0 +1,23 @@
...
+
+# - name: Export resource config for {{ resource_patch.kind }}/{{ resource_patch.name }}
```
ciaranRoche: Are we ok to remove this.

roles/resource_limits/tasks/patch_resource.yml
```diff
@@ -0,0 +1,23 @@
...
+  shell: |
+    oc set resources {{ resource_patch.kind }} {{ resource_patch.name }} -n {{ ns }}
+      {%- for k,v in resource_patch.resources.items() %}
+      --{{ k }}={% for k2,v2 in v.items() %}{{ k2 }}={{ v2 }}{% if not loop.last %},{% endif %}{% endfor %}
+      {%- endfor %}
+  register: set_resources_cmd
+  changed_when: ('not changed' not in set_resources_cmd.stderr)
+  failed_when: set_resources_cmd.stderr != '' and 'not changed' not in set_resources_cmd.stderr
+
+# - name: Apply new resource config for {{ resource_patch.kind }}/{{ resource_patch.name }}
```
ciaranRoche: And this can we remove it?

roles/enmasse/defaults/main.yml
```diff
@@ -11,3 +11,31 @@ enmasse_clean_artifacts: true
...
+    kind: deploy
+    resources:
+      requests:
+        memory: 256Mi
+  - name: enmasse-operator
+    kind: deploy
+    resources:
+      requests:
+        memory: 75M
+  # keycloak operator undoes any externally made changes
```
ciaranRoche: Can we remove this and maybe track in a JIRA instead?

roles/resource_limits/tasks/patch_resource.yml
jessesarn: I don't even think its worth creating a JIRA for this unless we find we really need it.  At this point, there are only 2 RHMI managed product components that have a multi-container pod:
- system-app (3scale) - all containers use the same default values and we were leaving as-is anyway..  overriding will do the same because 'oc set resources' without a container parameter specified uses a match * anyway.
- syndesis-db (fuse) - it does have different values for each container but was out of scope until at least next quarter anyway.

roles/resource_limits/tasks/patch_resource.yml
jessesarn: Yeah.  I was originally going the export so that I could then update the file rather than the actual resource during development.  The only other reason I can think of to export the config would be to have an easy way to revert everything back to the way it was for repeated testing.   However, you can revert to the last working config using 'oc rollout undo' which can be done by another person, from any machine, etc.

You can undo all of the resource changes made to the configs for both 3scale and enmasse with 2 commands:
```
for i in api-server enmasse-operator postgresql service-broker; do oc rollout undo deploy $i; done
oc get dc -n openshift-3scale -o name | while read r; do oc rollout undo $r; done
```

roles/resource_limits/tasks/patch_resource.yml
jessesarn: This was only needed when the edits were being made to the dumped file during development.

roles/enmasse/defaults/main.yml
jessesarn: I think these commented lines should stay.  They are the overrides we want for enmasse which is ultimately what this PR addresses.  The lines will be getting uncommented as part of https://issues.jboss.org/browse/INTLY-2656 where the operator will be updated to support custom requests and limits instead of only a memory limit like it does now.  I'll add a note to that ticket to capture this as well.

ciaranRoche: Awesome sounds good.

PR #852, opened 2019-08-14
Ignore rollout cmd error if rollout already in progress

david-martin:
## Additional Information
https://issues.jboss.org/browse/INTLY-2871

## Verification Steps

1. Install should succeed without any errors i.e. shouldn't see this

```
00:51:41 TASK [3scale : Redeploy system-app] ********************************************
00:51:41 fatal: [localhost]: FAILED! => {"changed": false, "cmd": "oc rollout latest system-app -n 3scale", "delta": "0:00:00.215321", "end": "2019-08-13 22:51:41.625569", "failed_when_result": true, "msg": "non-zero return code", "rc": 1, "start": "2019-08-13 22:51:41.410248", "stderr": "error: #1 is already in progress (Running).", "stderr_lines": ["error: #1 is already in progress (Running)."], "stdout": "", "stdout_lines": []}
```

No

- [ ] Tested Installation
- [ ] Tested Uninstallation

### DIFF
--- a/roles/3scale/tasks/users.yml
+++ b/roles/3scale/tasks/users.yml
@@ -44,7 +44,7 @@
 - name: Redeploy system-app
   shell: oc rollout latest system-app -n {{ threescale_namespace }}
   register: result
-  failed_when: result.stderr != '' and "already in progress" not in result.stderr
+  failed_when: not result.stdout
   changed_when: False
 
 - name: Verify 3scale deployment succeeded
--- a/roles/3scale/tasks/users.yml
+++ b/roles/3scale/tasks/users.yml
@@ -44,7 +44,7 @@
 - name: Redeploy system-app
   shell: oc rollout latest system-app -n {{ threescale_namespace }}
   register: result
-  failed_when: not result.stdout
+  failed_when: result.stderr != '' and "already in progress" not in result.stderr
   changed_when: False
 
 - name: Verify 3scale deployment succeeded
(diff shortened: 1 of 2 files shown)


roles/3scale/tasks/users.yml
```diff
@@ -44,7 +44,7 @@
 - name: Redeploy system-app
   shell: oc rollout latest system-app -n {{ threescale_namespace }}
   register: result
-  failed_when: not result.stdout
+  failed_when: result.stderr != '' and "already in progress" not in result.stderr
   changed_when: False
```
trepel: I assume that this redeploy is done here for some reason (e.g. config change to be applied). If we just skip it due to another redeploy is still in progress, we might want to end up in an undesirable state (e.g. config change won't be applied).

Not sure, can something like that be an issue?

david-martin: @trepel Hmm, good point.
The redeploy here looks to be because of the patching of the secrets above.
For some reason the check_readiness task is passing despite rollout #1 being in progress.
I'll take another stab at this.

odra: I manually changed the config map and it did not trigger a redeploy, something else did

trepel:
@david-martin I think this can be closed in favor of https://github.com/integr8ly/installation/pull/860

david-martin:
@trepel agreed

PR #856, opened 2019-08-14
Webapp mdc

odra:
## Additional Information

MDC required changes for msn and webapp usage

### Mobile Enabled

* Run an install using this pr branch
* Login into the webapp and check if mdc url is listed in the right menu
* Run the uninstall playbook

### Mobile Disabled

* Run the install playbook adding the following flags; `-e ups=false -e mdc=false -e datasync=false -e mobile_security_service=false`
* Login into the webapp and check if mdc service is not listed in the right menu
* Check if any of the mobile services were created
* Ru the uninstall playbook using the same vars listed in the first bullet item

- [ ] Tested Installation
- [ ] Tested Uninstallation

### DIFF
--- a/playbooks/install_mobile_services.yml
+++ b/playbooks/install_mobile_services.yml
@@ -7,9 +7,9 @@
     - include_vars: ../roles/mobile_security_service/defaults/main.yml
     - include_vars: ../roles/ups/defaults/main.yml
     - include_vars: ../roles/mdc/defaults/main.yml
+    - include_vars: ../roles/datasync/defaults/main.yml
 
-    - 
-      name: Install DataSync Templates
+    - name: Install DataSync Templates
       include_role:
         name: datasync
       tags: ['datasync']
@@ -21,8 +21,7 @@
       tags: ['ups']
       when: ups | default(true) | bool
 
-    -
-      name: Install Mobile Security Service
+    - name: Install Mobile Security Service
       include_role:
         name: mobile_security_service
       tags: ['mobile_security_service']
--- a/playbooks/install_mobile_services.yml
+++ b/playbooks/install_mobile_services.yml
@@ -42,3 +42,14 @@
       when:
         - user_rhsso | default(true) | bool
         - mdc | default(true) | bool
+    
+    - name: Install managed services broker
+      include_role:
+        name: msbroker
+      vars:
+        route_suffix: "{{ eval_app_host }}"
+        launcher_namespace: "{{ eval_launcher_namespace }}"
+        che_namespace: "{{ eval_che_namespace }}"
+        apicurio_namespace: "{{ eval_apicurito_namespace }}"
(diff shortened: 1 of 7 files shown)


trepel:
Running on my PDS and PoC cluster

roles/3scale/tasks/users.yml
```diff
@@ -47,6 +47,15 @@
...
 
+- name: Wait for system-app rollout
+  shell: "oc rollout status dc/system-app -n {{ threescale_namespace }} --watch=false | grep 'rolled out'"
+  register: result
+  until: result.stdout
+  retries: 50
+  delay: 10
+  failed_when: result.stderr
+  changed_when: False
+
```
trepel: @odra I am not sure if this fixes INTLY-2871. In order to fix INTLY-2871 this piece of code needs to be added before the `Redeploy system-app` step. That is not the best place though, the best would be to place this piece of code just after the piece of code that causing the system-app redeploy (but I don't know where that is).

playbooks/install_mobile_services.yml
```diff
@@ -41,4 +41,15 @@
         name: rhsso-user
       when:
         - user_rhsso | default(true) | bool
-        - mdc | default(true) | bool
\ No newline at end of file
+        - mdc | default(true) | bool
+    
+    - name: Install managed services broker
```
wei-lee: I thought the default installer will install msb, no? Why do we need to install it again?

playbooks/install_services.yml
```diff
@@ -137,12 +137,21 @@
...
     - name: Expose vars
       include_vars: "../roles/apicurito/defaults/main.yml"
+    
+    - name: Set additional walkthrough paths
+      block:
+        - set_fact:
+            webapp_wt_list: []
+        - set_fact:
+            webapp_wt_list: "{{ webapp_wt_list + [mobile_walkthrough_location] }}"
+          when: mdc
```
wei-lee: is the indentation right here for the `when` statement?

wei-lee:
left a few comments, otherwise LGTM. I think @grdryn is trying it atm. Will let him approve the PR once it's verified.

playbooks/install_mobile_services.yml
odra: because it needs to re-apply the msb config for mdc

playbooks/install_services.yml
odra: yes, it just executes the mdc stuff in here but I encapsulated all additional wt tasks in a block

roles/3scale/tasks/users.yml
odra: it is added just after the rollout, this command checks fo the rollout status

odra: I got what you mean, I think the config map changes is triggering a new deployment so I moved it before the rollout command

PR #860, opened 2019-08-14
3scale rollout fix

odra:
## Additional Information

Waits for 3scale system-app deployment rollout instead of rolling it out twice

## Verification Steps

Run an usual installation

- [ ] Tested Installation
- [ ] Tested Uninstallation

### DIFF
--- a/roles/3scale/tasks/service_discovery.yml
+++ b/roles/3scale/tasks/service_discovery.yml
@@ -25,12 +25,28 @@
 
 - name: Redeploy system-app
   shell: oc rollout latest system-app -n {{ threescale_namespace }}
+  changed_when: False
+
+- name: Wait for system-app rollout
+  shell: "oc rollout status dc/system-app -n {{ threescale_namespace }} --watch=false | grep 'rolled out'"
   register: result
-  failed_when: not result.stdout
+  until: result.stdout
+  retries: 50
+  delay: 10
+  failed_when: result.stderr
   changed_when: False
 
 - name: Redeploy system-sidekiq
   shell: oc rollout latest system-sidekiq -n {{ threescale_namespace }}
   register: result
   failed_when: not result.stdout
+  changed_when: False
+
+- name: Wait for system-sidekiq rollout
+  shell: "oc rollout status dc/system-sidekiq -n {{ threescale_namespace }} --watch=false | grep 'rolled out'"
+  register: result
+  until: result.stdout
+  retries: 50
+  delay: 10
+  failed_when: result.stderr
   changed_when: False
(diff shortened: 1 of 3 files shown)


trepel:
this passed successfully against my PDS twice. Running against PoC atm.

trepel:
Looking at the code, I am not quite sure what is the difference between `oc rollout latest system-app` and `oc rollout status dc/system-app` but if it effectively does the same the code looks good to me.

odra:
Status just checks the rollout status of a deployment config.

The problem was that patching 3scale config map was already triggering a rollout of system app so we just need to check the rollout status until it is done instead of doing another deployment config rollout (which was causing the "rollout already running" error

roles/3scale/tasks/service_discovery.yml
```diff
@@ -23,14 +23,26 @@
   register: result
   failed_when: not result.stdout
 
-- name: Redeploy system-app
-  shell: oc rollout latest system-app -n {{ threescale_namespace }}
+- name: Wait for system-app rollout
+  shell: "oc rollout status dc/system-app -n {{ threescale_namespace }} --watch=false | grep 'rolled out'"
```
trepel: ok, so there is no explicit need to trigger a rollout here because `oc patch configmap system...` from previous step causes the system-app to be redeployed automatically. Correct @odra ?

odra: yes - so I just replaced it with a rollout status check

roles/3scale/tasks/users.yml
```diff
@@ -41,10 +41,13 @@
   shell: |
     oc patch secret/system-seed -n {{ threescale_namespace }} -p "{\"stringData\": {\"ADMIN_USER\": \"{{ rhsso_evals_admin_username }}\"}}"
 
-- name: Redeploy system-app
-  shell: oc rollout latest system-app -n {{ threescale_namespace }}
+- name: Wait for system-app rollout
+  shell: "oc rollout status dc/system-app -n {{ threescale_namespace }} --watch=false | grep 'rolled out'"
```
trepel: Are you sure that `oc patch secret/system-seed...` in the previous step triggers the system-app rollout so we can only wait for the rollout to finish here? Not sure if mere patching of a secret triggers rollouts.

roles/3scale/tasks/service_discovery.yml
trepel: cool, thanks for confirming

roles/3scale/tasks/users.yml
odra: I changed the sceret and it didn't tigger a deployment, trying it without it

pmccarthy:
Currently validating against OSD 👀

PR #861, opened 2019-08-15
[WIP] update mdc version

craicoverflow:
Updated MDC to 1.1.6

### DIFF
--- a/inventories/group_vars/all/manifest.yaml
+++ b/inventories/group_vars/all/manifest.yaml
@@ -117,10 +117,10 @@ application_metrics: false
 
 #mobile developer console
 mdc: true
-mdc_operator_release_tag: '0.1.10'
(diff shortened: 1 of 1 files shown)


inventories/group_vars/all/manifest.yaml
```diff
@@ -117,10 +117,10 @@ application_metrics: false
 
 #mobile developer console
 mdc: true
-mdc_operator_release_tag: '0.1.10'
+mdc_operator_release_tag: '0.2.0'
```
odra: the operator image is not available, is it still being built?

craicoverflow: Yeah I meant to set this as draft.

craicoverflow: It should be ready now.

PR #864, opened 2019-08-15
monitoring checks

odra:
## Verification Steps

Run an installation setting the mobile flags to false: `-e 'mdc=false' -e 'ups=false' -e 'mobile_service_security=false' -e 'datasync=false'`

- [ ] Tested Installation
- [ ] Tested Uninstallation


pmccarthy:
👀

PR #869, opened 2019-08-16
upgrade keycloak

philbrookes:
## Verification
- Install from tag  `release-1.4.1`
- checkout this branch
- run upgrade playbook
- see SSO running 7.3 and init container is tag `0.0.3`


davidkirwan:
Should the init container tag be 0.0.3 or 1.0.0 here @davidffrench ? Or do we now also need to update that image in a later PR ?

philbrookes:
it should be 0.0.3, as this has the updated plugin, without pulling in the changes from Oleg that require a new volume to be created.

PR #886, opened 2019-08-20
fix backup namespaces for alerts

pb82:
cherry pick of #885


pb82:
ping @odra @davidkirwan

PR #894, opened 2019-08-21
[WIP] 3scale update

fh-automatron:


mikenairn:
The "delorean/integration-tests" failures can be ignored, seems to be some issue with the tower job we run before the install, will need further investigation. I was able to complete an installation on PDS manually, so happy enough it at least completes the install, did no further verification though.

PR #900, opened 2019-08-21
3scale ga cherry pick

pmccarthy:
## Additional Information
Cherry-picking commits merged back to master for 3scale 2.6.0 GA

Also removing brew yaml files generated by Delorean as no longer needed with 3Scale going GA


mikenairn:
@pmccarthy This *should* i think be all you need to do here, but as i have said elsewhere we should never really have been in a position where we had to do this on a release branch.

davidffrench:
Thank you @pmccarthy

PR #911, opened 2019-08-23
support mgmt of horizontal scaling

jessesarn:
## Additional Information

[INTLY-2957](https://issues.jboss.org/browse/INTLY-2957) - Adding support for managing replicas

[INTLY-2822](https://issues.jboss.org/browse/INTLY-2822) - explains the default values used for number of replicas along with reasoning for defaulting resource_limits_horizontal_scaling to false.

Also added tags which allow for fine-grained control of which product(s) will be affected.

## Verification Steps
Cluster can be running Integreatly 1.4 or 1.5
1.Log into cluster with oc as user with cluster-admin
2. List the number of replicas for dcs/deploys for 3scale and enmasse:
```
for i in {,openshift-}{3scale,enmasse}; do oc get deploy,dc -n $i; done
NAME                                                    REVISION   DESIRED   CURRENT   TRIGGERED BY
deploymentconfig.apps.openshift.io/apicast-production   5          1         1         config
deploymentconfig.apps.openshift.io/apicast-staging      5          1         1         config
...
NAME                                             DESIRED   CURRENT   UP-TO-DATE   AVAILABLE   AGE
deployment.extensions/address-space-controller   1         1         1            1           2d
deployment.extensions/agent.xzvb0k2oq1           1         1         1            1           2d
deployment.extensions/api-server                 1         1         1            1           2d
...
```
Note the values in the DESIRED and CURRENT columns are all 1's.
3. Run the playbook:
```
ansible-playbook -i inventories/hosts playbooks/update_resources.yml -e prerequisite_checks=false -e resource_limits_horizontal_scaling=true
```
4. Run the same cmd from step 2 where you should now see 6 components in 3scale with a 2 in the DESIRED column.  The 6 should be the the following:
```
deploymentconfig.apps.openshift.io/apicast-production   6          2         2         config
deploymentconfig.apps.openshift.io/apicast-staging      6          2         2         config
deploymentconfig.apps.openshift.io/backend-listener     5          2         2         config
deploymentconfig.apps.openshift.io/backend-worker       4          2         2         config
deploymentconfig.apps.openshift.io/system-sidekiq       6          2         2         config
deploymentconfig.apps.openshift.io/zync                 4          2         2         config
```
Note: It may take some time for the CURRENT column to show a 2.  The script is intended to minimize the amount of time services might disrupted disruption of services so it only waits for 1 replica to be ready before continuing.

No

- [n/a] Tested Installation
- [n/a] Tested Uninstallation
- [n/a] Tested or Created follow on for Upgrade

### DIFF
--- a/playbooks/update_resources.yml
+++ b/playbooks/update_resources.yml
@@ -5,6 +5,8 @@
     - include_role:
         name: 3scale
         tasks_from: new_limits.yml
+      tags: ['3scale']
     - include_role:
         name: enmasse
         tasks_from: new_limits.yml
+      tags: ['enmasse']
--- a/roles/3scale/defaults/main.yml
+++ b/roles/3scale/defaults/main.yml
@@ -68,6 +68,7 @@ threescale_resources:
     requests:
       cpu: 5m
       memory: 32Mi
+  replicas: 2
 - name: backend-cron
   kind: dc
   resources:
--- a/roles/3scale/defaults/main.yml
+++ b/roles/3scale/defaults/main.yml
@@ -61,18 +61,13 @@ threescale_resources:
     requests:
       cpu: 50m
       memory: 50Mi
+  replicas: 2
 - name: apicast-staging
   kind: dc
   resources:
     requests:
       cpu: 5m
       memory: 32Mi
-- name: apicast-wildcard-router
-  kind: dc
-  resources:
-    requests:
-      cpu: 12m
-      memory: 16Mi
 - name: backend-cron
   kind: dc
   resources:
@@ -85,18 +80,22 @@ threescale_resources:
     requests:
       cpu: 50m
       memory: 300Mi
+  replicas: 2
 - name: backend-redis
   kind: dc
   resources:
     requests:
       cpu: 100m
       memory: 10Mi
+#    limits:
+#      memory: 0
 - name: backend-worker
   kind: dc
   resources:
     requests:
       cpu: 15m
       memory: 30Mi
+  replicas: 2
 - name: system-app
   kind: dc
   resources:
@@ -121,12 +120,15 @@ threescale_resources:
     requests:
       cpu: 15m
       memory: 30Mi
+#    limits:
+#      memory: 0
 - name: system-sidekiq
   kind: dc
   resources:
     requests:
       cpu: 10m
       memory: 50Mi
+  replicas: 2
 - name: system-sphinx
   kind: dc
   resources:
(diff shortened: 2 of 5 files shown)


pmccarthy:
👀

pmccarthy:
@jessesarn Worked as expected 👍  I would suggest importing this playbook from install.yml based on a flag. By default it wouldn't get called but we can set overrides in tower to ensure its set for POC/OSD environments.

These overrides are set here:
**POC**
https://github.com/integr8ly/ansible-tower-configuration/blob/master/playbooks/roles/integreatly/templates/extra_vars_poc_install.yml.j2

**OSD**
https://github.com/integr8ly/ansible-tower-configuration/blob/master/playbooks/roles/integreatly/files/osd_install_vars.yml

maleck13:
Is this a playbook we expect CSSRE to run in order to scale up 3scale, if so do we document what the outcome of running this playbook will be in the SoP for CSSRE? 
Also will new clusters have the 2 replica setup for 3scale after this change lands https://github.com/integr8ly/installation/pull/911/files#diff-0111b6195c539fe946e1eac4c3dd692aR83 ? 
If so this is something we should ensure CS SRE are aware of, we should also ensure our monitoring is up to date with these changes to expect two replicas, and that upgrades of existing clusters bring existing clusters into line with this change. Is there a significant resource hit from the additional replicas?

jessesarn:
> Looks good 
> 
> One other thing we'll need to make the CSSRE team aware of is the fact that if they run this on an existing customer OSD there will be service disruption due to pods restarting following dc updates e.g. system-app in 3scale

Yeah, I did talk about this briefly yesterday on a call w/ @brizrobbo.  They can do horizontal scaling w/o service disruption as long as they don't have replicas=0 somewhere (can't imagine why they would do that, but just being thorough with my response).  I was thinking of this very issue when I added the flags for horizontal vs. vertical scaling.

So to be extra safe, they can pass resource_limits_vertical_scaling=false on the cmd line which will prevent any changes to requests/limit.  Vertical scaling can also be done without disruption as long as they are only changing resources for pods that have either:
- use rolling strategy
- more than one replica (even ok with recreate strategy)


I even thought about including a prompt which would show them the changes that are about to be made which would also highlight changes that could be service disrupting, but it would require quite a bit of time to implement.  The danger though would be that the logic could easily classify changes incorrectly due to complex relationships.  Even if we try to put in safety nets, CS-SRE definitely need to be conscious of what they are doing.

PR #914, opened 2019-08-27
ups_message_broker var

odra:
## Additional Information

Disables ups amq integration as the default installation path.

This PR introduces a new ups var: `ups_message_broker` which defaults fo `false` (setting it to true will enable amq online usage in ups). 

## Verification Steps
* run the installation (should succeed)
* check if there are any amq addresses crs in ups namespace (should be none)

- [ ] Tested Installation
- [ ] Tested Uninstallation


odra:
ping @grdryn @wei-lee

PR #921, opened 2019-08-27
default to public url for webapp console

pb82:
Use `openshift_master_public_url` for the webapp console link if available (usually set in tower environments). Default to `openshift_master_url` if not available.

Verification steps:

1. Install on PoC or PDS using this branch
2. Make sure the console link in the webapp works
3. Install on OSD and make sure the link works there too

### DIFF
--- a/roles/webapp/tasks/provision-webapp.yml
+++ b/roles/webapp/tasks/provision-webapp.yml
@@ -5,7 +5,7 @@
   failed_when: false
 
 - name: Construct openshift host string
-  shell: echo {{ openshift_master_url }} | cut -d '/' -f3
+  shell: echo {{ openshift_master_public_url | default(openshift_master_url) }} | cut -d '/' -f3
   register: openshift_host
 
 - name: Get SSO route


david-martin:
@pb82 The default value works as expected.
I also ran an install with an overwrite of `-e openshift_master_public_url=a.b.c`

This resulted in the following browser console error when loading the Solution Explorer
image
It makes sense that dns wouldn't resolve for this,
but I'm wondering if the webapp should be using a different url fo API calls than what's used for the link to the OpenShift console?

pb82:
@david-martin for OS4 a separate API URL has been added: https://github.com/integr8ly/tutorial-web-app-operator/blob/master/deploy/template/tutorial-web-app.yml#L11

But for OS3 we seem to just rely on one variable OPENSHIFT_HOST, which is also used to construct the console URL: https://github.com/integr8ly/tutorial-web-app/blob/master/src/components/installedAppsView/InstalledAppsView.js#L120

It looks like no `api` subdomain is required on OS3.

PR #939, opened 2019-08-30
INTLY-3165 Do user sso upgrade/install before msbroker

david-martin:
## Additional Information
Fix for https://issues.jboss.org/browse/INTLY-3165

## Verification Steps
1. Install release-1.4.1
2. Run upgrade playbook
3. Verify it completes successfully
3. Verify user sso is installed

### DIFF
--- a/playbooks/upgrades/upgrade.yaml
+++ b/playbooks/upgrades/upgrade.yaml
@@ -24,6 +24,12 @@
         upgrade_webapp_operator_image: quay.io/integreatly/tutorial-web-app-operator:{{ webapp_operator_release_tag }}
         upgrade_fuse_image_tag: 1.7
 
+
+- import_playbook: install_user_rhsso.yml
+
+- hosts: localhost
+  gather_facts: yes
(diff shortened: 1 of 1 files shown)


tremes:
I got 
`fatal: [127.0.0.1]: FAILED! => {"msg": "The conditional check 'user_rhsso | default(true) | bool and backup_restore_install | bool' failed. The error was: error while evaluating conditional (user_rhsso | default(true) | bool and backup_restore_install | bool): 'backup_restore_install' is undefined\n\nThe error appears to have been in '/tmp/installation/playbooks/upgrades/install_user_rhsso.yml': line 19, column 7, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n    -\n      name: Setup backup for user rhsso\n      ^ here\n"}`

when trying to run the upgrade. Am I missing some vars?

pb82:
@tremes try adding `-e backup_restore_install=false` when running the upgrade playbook

PR #943, opened 2019-09-02
INTLY-3165 Do user sso upgrade/install before msbroker

pb82:
cherry pick of #939


david-martin:
@pb82 Given that the user sso instance is being added to all clusters in 1.5 (via upgrade) I don't think we need to fix it here.
It would be better to remove these upgrade steps instead on master

pb82:
Makes sense, closing this

PR #946, opened 2019-09-03
wait for web app operator to be ready before applying new CR

pb82:
Wait for the new web app operator to be ready before applying the new CR (otherwise the old operator will still handle it and deploy the old version).

Verification steps:

1. Install 1.4.1
2. Verify that the version of the webapp operator is v0.0.17 and the version of the webapp is v2.10.x.
3. Run the upgrade from this branch
4. Upgrade should succeed
5. The version of the webapp operator, web app and walkthroughs must match what is specified in the manifest.
6. Make sure that under `Storage` there exists a PV called `user-walkthroughs`.

### DIFF
--- a/roles/webapp/tasks/upgrade.yaml
+++ b/roles/webapp/tasks/upgrade.yaml
@@ -12,6 +12,12 @@
   register: upgrade_webapp_operator
   failed_when: upgrade_webapp_operator.stderr != '' and 'not patched' not in upgrade_webapp_operator.stderr
 
+- name: Wait for the new web app operator to be ready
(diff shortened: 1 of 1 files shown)


davidffrench:
👀

davidffrench:
@pb82 I have verified using the steps in the description. the versions in the manifest match for the operator and webapp. I am only seeing 2 walkthroughs after upgrade, is this correct or should I be seeing the mobile walkthrough as well?

pb82:
@davidffrench yes that's correct because on the master branch we install v1.7.2 of the WT repository and there are only two walkthroughs: https://github.com/integr8ly/tutorial-web-app-walkthroughs/tree/v1.7.2/walkthroughs

I guess this will be backported at some point.

PR #947, opened 2019-09-03
Reset upgrade playbook for 1.6

david-martin:
## Additional Information
Reset the upgrade playbook to remove 1.5 specific upgrade bits.

## Verification Steps
n/a


playbooks/upgrades/upgrade.yaml
```diff
@@ -116,19 +81,16 @@
     - include_role:
         name: middleware_monitoring
         tasks_from: upgrade/grafana
-      when: target_version.stdout == "release-1.4.1"
```
david-martin: @pb82 I'm removing the extra version checks here and relying on the prerequisite at the start.
Does it make sense to include these monitoring tasks as a starting point for the next upgrade?

davidffrench:
@david-martin Do we want to wait for 1.5 to be released before this is merged?

playbooks/upgrades/upgrade.yaml
```diff
@@ -156,8 +117,5 @@
       vars:
         fuse_image_tag: "{{ upgrade_fuse_image_tag }}"
 
-- import_playbook: install_user_rhsso.yml
-- import_playbook: 3scale_upgrade_2.5_to_2.6.yml
```
david-martin: These 2 are specific to upgrade to 1.5

playbooks/upgrades/upgrade.yaml
```diff
@@ -73,41 +73,6 @@
...
-    - name: Install application metrics
-      include_role:
-        name: application_monitoring
-      when: application_metrics
-
-    - name: Upgrade code ready workspaces 1.0 to 1.2
-      include_role:
-        name: code-ready
-        tasks_from: upgrade_1.0_to_1.2.yml
-
```
david-martin: All of above are specific to upgrade to 1.5.

@wei-lee is there a generic `upgrade` task for any of the mobile components at this time that could stay here on master?

roles/3scale/tasks/upgrade.yml
```diff
@@ -1,226 +0,0 @@
-#upgrade
```
david-martin: Removing 3scale 2.5 specific upgrade file to avoid confusion.

david-martin:
> @david-martin Do we want to wait for 1.5 to be released before this is merged?

I think it's safe to do this as soon as the release branch (v1.5 in this case) is created.
Once the release has it's first RC, any upgrade fixes are only relevant to that release branch, not master.

Does that sound reasonable?

docs/release.md
```diff
@@ -10,6 +10,21 @@
     
     ``` ./scripts/release.sh -b v2.7 -r release-2.7.0-rc1```
 
+
+### Resetting the upgrade playbook on master
```
david-martin: @maleck13 @mikenairn I'd appreciate your review of this doc section.
It's kind of what has been done for the previous 2 releases anyways.

davidffrench:
> > Do we want to wait for 1.5 to be released before this is merged?
> 
> @davidffrench I think it's safe to do this as soon as the release branch (v1.5 in this case) is created.
> Once the release has it's first RC, any upgrade fixes are only relevant to that release branch, not master.
> 
> Does that sound reasonable?

Ya, I think that sounds reasonable. Right now we don't test the cherry picked PR much as this is done off master. We will need to make sure both are tested correctly with this change which would also make sense anyway.

david-martin:
@maleck13 @mikenairn Would appreciate your thoughts on tagged section above

docs/release.md
maleck13: looks reasonable to me

mikenairn: Sounds good to me also

PR #951, opened 2019-09-04
turn on mds by default

wei-lee:


wei-lee:
@davidffrench this is the cherry-pick one.

PR #955, opened 2019-09-05
Adding 3scale, AMQ and SSO alerts and scaling documents

obrienrobert:
## Additional Information
https://issues.jboss.org/browse/INTLY-3040

- Creation of memory usage alerts for the 3scale, SSO and AMQ namespaces
- Creation of high CPU usage alert for the 3scale namespace
- Adding links in the relevant alerts to the AMQ and SSO scaling documents

## Verification Steps

1. Install `Integreatly` using this feature branch

2. Set-up [SMTP](https://github.com/fheng/integreatly-help/blob/master/sops/POC_provisioning.asciidoc#44-smtp-configuration ) on your cluster, ensuring to set your email address as the recipient

3. Change the default value (90%) of each alert by navigating to the `openshift-middleware-monitoring namespace -> resources -> other resources -> Prometheus Rule -> ksm-alerts` to a low value to trigger each alerts e.g. 0%. Also change the `for` duration to a smaller duration e.g. from `15m` to `1m` - this is the length of time it the CPU/memory must be over the specified value for before an alert is triggered

4. Ensure that you receive an email for each of the alerts that are triggered, and that each alert has a link to the correct scaling plan document

This task is not related to any upgrade process.

### DIFF
--- a/roles/middleware_monitoring_config/templates/kube_state_metrics_alerts.yml.j2
+++ b/roles/middleware_monitoring_config/templates/kube_state_metrics_alerts.yml.j2
@@ -148,3 +148,40 @@ spec:
         for: 5m
         labels:
           severity: critical
+      - alert: ThreeScalePodHighMemory
+        annotations:
+          message: The {{ '{{' }} $labels.container {{ '}}' }} pod has been using {{ '{{' }} printf "%.0f" $value {{ '}}' }}% of available memory for longer than 15 minutes.
+          scaling_plan: https://github.com/integr8ly/middleware-load-testing/blob/master/sops/3scale-scaling.md
+        expr: |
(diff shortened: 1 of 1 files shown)


david-martin:
> * Set-up [SMTP](https://github.com/fheng/integreatly-help/blob/master/sops/POC_provisioning.asciidoc#44-smtp-configuration) on your cluster, ensuring to set your email address as the recipient

@obrienrobert Although this will verify the alert is triggered, it's probably overkill.
To make things easier in future, it would be enough to check in Prometheus that the alert is triggering rather than having to configure alertmanager as well with a sendgrid or similar api key

obrienrobert:
@david-martin these were just the steps that I used when testing, but you are correct in that it is probably more work that necessary to validate - thanks for the feedback!

PR #957, opened 2019-09-05
INTLY-3299 Change keycloak backup job name to include namespace as su…

david-martin:
…ffix

## Additional Information
https://issues.jboss.org/browse/INTLY-3299

## Verification Steps
1. Install from release-1.4.1 tag.
2. Verify the sso cronjob for the cluster sso instance *doesn't* have the namespace suffix (`daily-at-midnight`). `oc get cronjob -n sso`
3. Run the upgrade
4. Verify the sso cronjob for the cluster sso instance *does* have the namespace suffix (`daily-at-midnight-sso`). `oc get cronjob -n sso`
5. Verify there is an alert called `CronJobExists_sso_daily-at-midnight-sso` in prometheus, and it is *not* triggering


david-martin:
I've followed the verification steps myself to sanity check.
Would appreciate another set of eyes to ensure I haven't missed something.

PR #963, opened 2019-09-06
use assetPublicUrl which works for console and oauth redirects

pb82:
Switch from `masterPublicUrl` to `assetPublicUrl`. This one can be used for OAuth redirects as well as for console access.

Docs: https://docs.openshift.com/container-platform/3.5/install_config/master_node_configuration.html#master-config-oath-authentication-config

The values of the vars are for example:

```
On OSD:

assetPublicURL: https://console.rhmds-qe1.openshift.com/console/ 
masterPublicURL: https://api.rhmds-qe1.openshift.com

On PDS:

assetPublicURL: https://master.intlymon-95de.open.redhat.com/console/
masterPublicURL: https://master.intlymon-95de.open.redhat.com:443
```

This *should* fix the console llinks as well as the login issues on OSD. Needs to be tested on both.

### DIFF
--- a/roles/openshift/tasks/set_master_vars.yml
+++ b/roles/openshift/tasks/set_master_vars.yml
@@ -8,4 +8,4 @@
   - add_host:
       name: EVAL_VARS
       eval_app_host: "{{ (eval_master_config['content'] | b64decode | from_yaml)['routingConfig']['subdomain'] }}"
-      openshift_master_url: "{{ (eval_master_config['content'] | b64decode | from_yaml)['masterPublicURL'] }}"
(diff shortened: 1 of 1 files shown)


wei-lee:
the change seems fine, the only question I have is will this still works for PDS?

pb82:
@wei-lee we just tested it and unfortunately it doesnt seem to work on PDS. The only difference in the URL seems to be the port. This is on my PDS instance:

```
masterPublicURL: https://master.pbraun-edd4.open.redhat.com:443
assetPublicURL: https://master.pbraun-edd4.open.redhat.com/console/
```

We're currently testing this on OSD. If it works there then i think we can postprocess the URL for pds (remove the /console and append the port).

PR #964, opened 2019-09-06
use assetPublicUrl which works for console and oauth redirects

pb82:
cp of https://github.com/integr8ly/installation/pull/963

### DIFF
--- a/roles/openshift/tasks/set_master_vars.yml
+++ b/roles/openshift/tasks/set_master_vars.yml
@@ -8,4 +8,4 @@
   - add_host:
       name: EVAL_VARS
       eval_app_host: "{{ (eval_master_config['content'] | b64decode | from_yaml)['routingConfig']['subdomain'] }}"
-      openshift_master_url: "{{ (eval_master_config['content'] | b64decode | from_yaml)['masterPublicURL'] }}"
+      openshift_master_url: "{{ (eval_master_config['content'] | b64decode | from_yaml)['oauthConfig']['assetPublicURL'] }}"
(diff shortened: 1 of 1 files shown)


roles/openshift/tasks/set_master_vars.yml
```diff
@@ -8,4 +8,4 @@
   - add_host:
       name: EVAL_VARS
       eval_app_host: "{{ (eval_master_config['content'] | b64decode | from_yaml)['routingConfig']['subdomain'] }}"
-      openshift_master_url: "{{ (eval_master_config['content'] | b64decode | from_yaml)['masterPublicURL'] }}"
\ No newline at end of file
+      openshift_master_url: "{{ (eval_master_config['content'] | b64decode | from_yaml)['oauthConfig']['assetPublicURL'] }}"
```
david-martin: @pb82 what is this value likely to be in pds, poc & osd clusters?

pb82:
```
On OSD:

assetPublicURL: https://console.rhmds-qe1.openshift.com/console/ 
masterPublicURL: https://api.rhmds-qe1.openshift.com

On PDS:

assetPublicURL: https://master.intlymon-95de.open.redhat.com/console/
masterPublicURL: https://master.intlymon-95de.open.redhat.com:443
```

I don't have an example for PoC yet, but would be good to get one.

david-martin: Will the `/console` path on the end be a problem for this var?

pb82: I think not because the docs say that it should be used for redirect urls: https://docs.openshift.com/container-platform/3.5/install_config/master_node_configuration.html#master-config-oath-authentication-config

pb82: actually, it might be a problem. Testing it now with the `/console` part stripped away

PR #965, opened 2019-09-06
Removal of 3scale wildcard imagestream for 2.6 upgrade

pmccarthy:
**Summary**
Following further investigations into 3Scale wildcard configuration on OSD [1], we can safely remove the remaining wildcard resources that had been commented out as part of the 2.5 to 2.6 upgrade playbook. This also includes the 3Scale 3amp wildcard route

[1] https://issues.jboss.org/browse/INTLY-3317

**Validation**
- [ ] Install RHMI v1.4.1 on an OSD Cluster
- [ ] Follow [SOP](https://github.com/fheng/integreatly-help/blob/master/sops/3scale_wildcard_router_on_osd_routing_shard.md) for 3Scale wildcard router setup: 
- [ ] Check that the wildcard ImageStream and Route exist pre-upgrade
```
oc get ImageStream amp-wildcard-router -n openshift-3scale
oc get route apicast-wildcard-router-routing-shard -n openshift-3scale
```
- [ ] Upgrade to v1.5 via upgrade playbook
- [ ] Check that the wildcard ImageStream and Route no longer exist post-upgrade
```
oc get ImageStream amp-wildcard-router -n openshift-3scale
oc get route apicast-wildcard-router-routing-shard -n openshift-3scale
```
- [ ] Re-run upgrade playbook to validate repeatability

### DIFF
--- a/playbooks/upgrades/3scale_upgrade_2.5_to_2.6.yml
+++ b/playbooks/upgrades/3scale_upgrade_2.5_to_2.6.yml
@@ -686,6 +686,11 @@
       register: apicast_wildcard_router_dc_removed
       failed_when: apicast_wildcard_router_dc_removed.stderr and 'NotFound' not in apicast_wildcard_router_dc_removed.stderr
 
+    - name: "Remove 3amp wildcard route from {{ threescale_namespace }} namespace"
+      shell: oc delete route apicast-wildcard-router-routing-shard
+      register: apicast_wildcard_router_shard_removed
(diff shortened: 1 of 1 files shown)


maleck13:
Actually I just realised this is an upgrade. So only needs to go against 1.5 branch. I think we should revert this and keep the cherry pick

pmccarthy:
@maleck13 Not sure it makes much of a difference that this made it back to master, but I do understand your point that this is really only needed for v1.5 branch. Reverting this just adds another commit into the master log.

PR #971, opened 2019-09-09
use correct master vars for webapp

pb82:
Use the correct url vars in the webapp.

This also covers the upgrade because we re-apply the webapp CR.

### DIFF
--- a/playbooks/install_services.yml
+++ b/playbooks/install_services.yml
@@ -144,6 +144,7 @@
         name: webapp
       vars:
         openshift_master_url: "{{ hostvars['EVAL_VARS']['openshift_master_url'] }}"
+        openshift_asset_url: "{{ hostvars['EVAL_VARS']['openshift_asset_url'] }}"
       tags: ['webapp']
       when: eval_action == 'install' and webapp
 
--- a/playbooks/upgrades/upgrade.yaml
+++ b/playbooks/upgrades/upgrade.yaml
@@ -53,6 +53,7 @@
         rhsso_openshift_master_config_path: "{{ eval_openshift_master_config_path }}"
         rhsso_namespace: "{{ eval_rhsso_namespace }}"
         openshift_master_url: "{{ hostvars['EVAL_VARS']['openshift_master_url'] | replace('https://', '') }}"
+        openshift_asset_url: "{{ hostvars['EVAL_VARS']['openshift_asset_url'] | replace('https://', '') }}"
 
     - name: SSO upgrade
       include_role:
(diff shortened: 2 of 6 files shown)


wei-lee:
@pb82 how can I verify this?

pb82:
@wei-lee An installation or upgrade from this branch. I'm running this on my cluster now but we also need to test it on OSD.

pb82:
@wei-lee Installed now on https://master.pbraun-e728.open.redhat.com 

Login to MDC also seems to work. Upgrade testing still needs to be done.

maleck13:
Is someone doing the upgrade testing here?

pb82:
@maleck13 @odra is currently doing the upgrade verification

maleck13:
ok great thanks

psturc:
Hit an error during upgrade from 1.4.1 on OSD

```
TASK [webapp : Wait for the new web app operator to be ready] ******************
changed: [127.0.0.1]
TASK [webapp : Construct openshift host string] ********************************
fatal: [127.0.0.1]: FAILED! => {"msg": "The task includes an option with an undefined variable. The error was: 'ansible.vars.hostvars.HostVarsVars object' has no attribute 'openshift_asset_url'\n\nThe error appears to have been in '/var/lib/awx/projects/_1014__integreatly_update_v15_fix_webapp_login_and_console_links/roles/webapp/tasks/upgrade.yaml': line 21, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: Construct openshift host string\n  ^ here\n"}
```

pb82:
@psturc should be fixed with this commit: https://github.com/integr8ly/installation/pull/971/commits/a2b3aa6c6fa0cf13e23bfdac83706560ba5583eb

can you check if you ran it with this commit?

pb82:
@psturc It looks like this happens because we don't run the master tasks on OSD: https://github.com/integr8ly/installation/blob/master/inventories/managed.template#L16

But this is required to get those URLs from the master config.

maleck13:
any thing that needs to be sourced from master configs should be defined as a var so that it can be passed in during OSD install upgrade

pb82:
@maleck13 So it should be working as long as there is a value set for `openshift_asset_url` at the Tower level? I could add it to the managed template, but its hard to set a default value for it because it depends on the cluster URL.

maleck13:
I think it probably needs a SoP change to tell CS SRE how they can get the url and that the var must be set for new installs and upgrades. @psturc could then follow the same processs

pb82:
@maleck13 Ok, updating the SOP now
@psturc I'll ping you once i have it and then we can retry it following the SOP

PR #981, opened 2019-09-09
double delay value for mss when waiting for pods to be ready as on OS…

wei-lee:
…D it will take a long time to claim a PV


wei-lee:
@psturc FYI, if you can approve this.

wei-lee:
@odra need your help to approve this as well.

PR #984, opened 2019-09-10
use placeholders for the mdc namespace

wei-lee:
https://issues.jboss.org/browse/INTLY-3343


wei-lee:
@david-martin @odra quick review if that's ok?

PR #985, opened 2019-09-10
use placeholders for the mdc namespace

wei-lee:
cherry-pick of #984


wei-lee:
@david-martin @odra quick review if that's ok?

maleck13:
does this resolve the issue?

PR #991, opened 2019-09-11
INTLY-3387 Update the Github IDP in launchers SSO to add the user:ema…

david-martin:
## Additional Information
https://issues.jboss.org/browse/INTLY-3387

## Verification Steps
1. Prereqs before installing from release-1.4.1 (otherwise upgrade fails later)
```
oc create namespace openshift-integreatly-backups

cat <<EOF | oc apply -n openshift-integreatly-backups -f -
  apiVersion: v1
  kind: Secret
  metadata:
    name: s3-credentials
    namespace: openshift-integreatly-backups
  type: Opaque
  stringData:
    AWS_S3_BUCKET_NAME: my_aws_bucket_name
    AWS_ACCESS_KEY_ID: my_aws_key_id
    AWS_SECRET_ACCESS_KEY: my_aws_access_key
EOF
```
2. Install from release-1.4.1, including setting up a Github OAuth App, and configure the vars during install `github_client_id`, `github_client_secret`, and include backups `backup_restore_install=true` 
3. Verify you can login to launcher and get past the github authorization step in the wizard
4. Do an upgrade, ensuring you set extra vars for backups `aws_s3_backup_secret_name=s3-credentials`, `aws_s3_backup_secret_namespace=openshift-integreatly-backups` (otherwise, upgrade fails)
5. Verify again, with a different user, you can login to launcher and get past the github authorization step in the wizard. If using the same github account, you should remove the identity provider link first from that user in launchers sso. (ping for details if needed)


david-martin:
Although this is ready for review, I haven't verified 1 thing yet.
When updating the idp, the current config is retrieved first, and then sent back in the update request (with the modified scopes).
However, in that request, the client secret is all stars (e.g. **********). This is the same behaviour as the UI when updating a field for an idp. I haven't verified if the update actually keeps the real value of the client secret, or it sets to all stars. 

I'll take a look at this soon.

david-martin:
> I'll take a look at this soon.

I've taken a look at this and it seems to be OK. After upgrade, the I was able to auth against github in the launcher wizard. This suggests the secret was not modified

PR #992, opened 2019-09-12
Allow alertmanager config to be customised 

alanmoran:
## Additional Information
The goal of this PR is to remove the need for CS SRE to manually integrate pagerduty, sendgrid and deadmanssnitch in prom alertmanager. 

## Verification Steps
Set any of the following extra vars `sendgrid_api_key`, `deadmanssnitch_url` or `pagerduty_integration_key` and it will create an alertmanger config and apply it. 

N/A

- [x] Tested Installation
- [ ] Tested Uninstallation

### DIFF
--- a/roles/middleware_monitoring_config/tasks/create_alertmanager.yml
+++ b/roles/middleware_monitoring_config/tasks/create_alertmanager.yml
@@ -1,0 +1,27 @@
+---
+
+- name: Configure AlertManager
+  block:
+  - name: get alertmanager route for alertmanager config
+    shell: oc get route alertmanager-route -o template --template \{\{.spec.host\}\} -n {{ middleware_monitoring_namespace }}
+    register: alertmanager
+
+  - set_fact:
+      alertmanager_route: "{{alertmanager.stdout}}"
+    when: alertmanager
+
+  - name: Generate custom alertmanager config
+    template:
+      src: "alertmanager.yml.j2"
+      dest: /tmp/alertmanager.yaml
+
+  - name: Create and apply alertmanager-application-monitoring secret
+    shell: oc create secret generic alertmanager-application-monitoring --from-file=/tmp/alertmanager.yaml --dry-run -o yaml | oc apply -n {{ middleware_monitoring_namespace }} -f -
+
+  - name: Remove generated alertmanager template
+    file: path='/tmp/alertmanager.yaml' state=absent
+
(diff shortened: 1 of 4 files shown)


roles/middleware_monitoring_config/tasks/create_alertmanager.yml
```diff
@@ -0,0 +1,27 @@
...
+  - name: Create and apply alertmanager-application-monitoring secret
+    shell: oc create secret generic alertmanager-application-monitoring --from-file=/tmp/alertmanager.yaml --dry-run -o yaml | oc apply -n {{ middleware_monitoring_namespace }} -f -
+
+  - name: Remove generated alertmanager template
+    file: path='/tmp/alertmanager.yaml' state=absent
+
+  # Once any of these are present, update alertmanager secret
+  when: smtp_auth_password != '' or
+        dms_webhook_url != '' or
+        pd_service_key != '' | default('')
```
alanmoran: TODO - Remove the `default('')` from there

davidkirwan:
Installed from branch ``INTLY-3214``

```
ansible-playbook -i inventories/hosts.template playbooks/install.yml \
-e eval_self_signed_certs=false -e eval_seed_users_count=1 \
-e backup_restore_install=true 
-e sendgrid_api_key="sometestkey" 
-e deadmanssnitch_url="https://whatisthisevenfor.com" 
-e pagerduty_integration_key="somepagerdutykey"
```

Screenshot from 2019-09-16 13-28-44

The values are being populated correctly.

davidkirwan:
Question tho, should this change be made here?? Rather than after an installation? The installer prior to this change already creates this secret right? Should we not patch it with the updates we require?

alanmoran:
> The deadmansnitch values are being populated correctly, but I don't see the sendgrid_api_key or the pagerduty_integration_key ones being populated there actually.

Sendgrid API Key is listed at the top as smtp_auth_password. Pagerduty was also set correctly it is under the pagerduty_configs block

alanmoran:
> Question tho, should this change be made here?? Rather than after an installation? The installer prior to this change already creates this secret right? Should we not patch it with the updates we require?

So its the application-monitoring-operator that creates the secret - I suppose the new question is if we want the operator to be customisable or we expect customers to load their own config. 

The reason I went with the installer was that this felt like we were "RHMI"-ing the alertmanager for our RHMI needs.

PR #997, opened 2019-09-13
INTLY-3431 - fix user sso IDP setup on upgrade

matskiv:
## Additional Information
JIRA: https://issues.jboss.org/browse/INTLY-3431
In this PR:
- updating KeycloakRealm CRD - new property has been added, and is used in User SSO
- adding call for previously missing tasks file - `setup-master-realm.yml`
- fixing a mistake introduced in #995 

## Verification Steps
Get 1.4.1 cluster (or if you have partially updated cluster - remove user sso namespace)
Execute the upgrade to 1.5 using this PR
Verify that customer-admin can login into user sso administration web console

- [x] Tested Upgrade

### DIFF
--- a/playbooks/upgrades/install_user_rhsso.yml
+++ b/playbooks/upgrades/install_user_rhsso.yml
@@ -2,29 +2,46 @@
 - hosts: localhost
   gather_facts: true
   tasks:
-    - name: Include vars from rhsso
-      include_vars: "../../roles/rhsso/defaults/main.yml"
-    -
-      name: Install user rhsso
-      include_role:
-        name: rhsso
-        tasks_from: install_sso.yml
-      vars:
-        sso_namespace: "{{ eval_user_rhsso_namespace }}"
-        sso_namespace_display_name: "User Facing Red Hat Single Sign-On"
-        rhsso_provision_immediately: True
+    - block:
+      - name: Include vars from rhsso
+        include_vars: "../../roles/rhsso/defaults/main.yml"
+
+      - name: "Update KeycloakRealm CRD"
+        shell: "oc replace -f {{ rhsso_operator_resources }}/crds/KeycloakRealm_crd.yaml -n {{ eval_rhsso_namespace }}"
+
+      - name: Find out original customer-admin password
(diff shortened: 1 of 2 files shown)


trepel:
I had a chat with Oleg, we ran the upgrade and it looks good. However, somebody more knowledgeable should review the code itself.

playbooks/upgrades/install_user_rhsso.yml
```diff
@@ -2,29 +2,46 @@
...
+          sso_namespace: "{{ eval_user_rhsso_namespace }}"
+          sso_namespace_display_name: "User Facing Red Hat Single Sign-On"
+          rhsso_provision_immediately: true
+          
+      - name: Setup IDP and customer-admin permissions in master realm
+        include_role:
+          name: rhsso-user
+          tasks_from: setup-master-realm.yml
+        vars:
+          openshift_master_url: "{{ hostvars['EVAL_VARS']['openshift_master_url'] }}"
```
maleck13: does this work without ssh access? I presume it will default to the var passed in by the tower instance

maleck13:
@matskiv as the upgrade was validated (I assume with tower?) I think we are good to merge and cherry pick here

playbooks/upgrades/install_user_rhsso.yml
matskiv: Yes, it will be set based on vars passed by tower.

Implementation for that is here:
1) https://github.com/integr8ly/installation/blob/v1.5/playbooks/upgrades/upgrade.yaml#L16-L18
2) https://github.com/integr8ly/installation/blob/v1.5/roles/prerequisites/tasks/upgrade.yml#L2-L4
3) https://github.com/integr8ly/installation/blob/v1.5/roles/prerequisites/tasks/check_master_params.yml#L9-L13

matskiv:
This hasn't been tested with tower, but I explained how hostvars will be populated when it's run from tower, so IMHO it can be merged.
btw: this change is directly for v1.5 so no cherry pick :)

maleck13:
ok happy to merge based on comments here

PR #1000, opened 2019-09-13
Changing upgrade release version to 1.5.0

pmccarthy:
**Summary**
In order to support upgrade playbook re-runs for v1.5 we need to add it to the `from_versions` variable

**Validation**
- [ ] Upgrade from RHMI v1.4.1 to v1.5 on tower
- [ ] Re-run the same upgrade job on tower and ensure it completes successfully

### DIFF
--- a/playbooks/upgrades/upgrade.yaml
+++ b/playbooks/upgrades/upgrade.yaml
@@ -19,6 +19,7 @@
       vars:
         from_versions:
           - "release-1.4.1"
+          - "release-1.5.0"
     - set_fact:
         upgrade_webapp_image: quay.io/integreatly/tutorial-web-app:{{ webapp_version }}
         upgrade_webapp_operator_image: quay.io/integreatly/tutorial-web-app-operator:{{ webapp_operator_release_tag }}


pmccarthy:
@maleck13 We have an upgrade task in the prereq role that retrieves the current version of RHMI via the webapp, see [1]. After the first run of the upgrade playbook the version will now be v1.5.0 meaning and we are now unable to re-run the upgrade playbook if needed.

We already do this for v1.4, see [2]

[1] https://github.com/integr8ly/installation/blob/master/roles/prerequisites/tasks/upgrade.yml#L6-L8
[2] https://github.com/integr8ly/installation/blob/release-1.4.1/playbooks/upgrades/upgrade.yaml#L17-L20

davidffrench:
Copying my comment from JIRA.

I think the problem here is that we have no current way of knowing that an upgrade playbook has been run completely successfully. We use the webapp version, however, the webapp is actually one of the first things in the upgrade playbook to get upgraded so it is highly unlikely to fail before that image is updated which is then used to check the current version running on the cluster. I don't think this is something we will solve during this release so your PR makes sense until there is an alternative.

maleck13:
The version update in the manifest should be that last thing that is done in an upgrade. That way we can be sure the upgrade playbook executed correctly without any failures before the version was updated. If some fails after this point, we probably should be making the playbook wait until it can check that whatever was upgraded is now upgraded (IE pods running etc). 
I think it is a bug on our part if the upgrade playbook updates the version before it knows the upgrade is complete.
Given the short time frame we have here, I guess we can go ahead and merged this, but it shouldn't be our default behaviour.

PR #1011, opened 2019-09-17
Adding sop_url in prometheus alerts

davidkirwan:

- [x] Tested Installation
- [ ] Tested Uninstallation


david-martin:
👀

PR #1018, opened 2019-09-20
add operator cr patches for keycloak backup jobs

aidenkeating:
## Verification

- Run the 1.5.0 upgrade
- Ensure cronjobs now use `1.0.10`
- Ensure the Keycloak CRs also reference this image tag


aidenkeating:
@maleck13 @wei-lee This should update all backup CronJobs to use the correct backup container image.

@wei-lee I think I've covered all the mobile ones here (MSS and UPS), can you confirm?

PR #1031, opened 2019-09-25
Intly 3312 capacity alerting

StevenTobin:
## Additional Information
Adds three new alerts

ClusterSchedulableMemoryLow - Checks memory requests for pods running on compute nodes and alerts if total if above 85% of total of compute nodes
ClusterSchedulableCPULow - Checks CPU core requests for pods running on compute nodes and alerts if total if above 85% of total of compute nodes
PVCStorageAvailable - Checks disk usage of PVCs against total size requested of PVC and alerts if above 85%

## Verification Steps
1. Install from this branch
2. Ensure the three new alerts are in alertmanager
3. Edit some deployment configs to request large amounts of memory and CPU
4. You can use these prometheus queries to see current usage

*Memory*
`(sum(sum by(node) (sum by(pod, node) (kube_pod_container_resource_requests_memory_bytes * on(node) group_left() (sum by(node) (kube_node_labels{label_node_role_kubernetes_io_compute="true"} == 1))) * on(pod) group_left() (sum by(pod) (kube_pod_status_phase{phase="Running"}) == 1))))  / ((sum((kube_node_labels{label_node_role_kubernetes_io_compute="true"} == 1) * on(node) group_left() (sum by(node) (kube_node_status_allocatable_memory_bytes))))) * 100`

*CPU*
`(sum(sum by(node) (sum by(pod, node) (kube_pod_container_resource_requests_cpu_cores * on(node) group_left() (sum by(node) (kube_node_labels{label_node_role_kubernetes_io_compute="true"} == 1))) * on(pod) group_left() (sum by(pod) (kube_pod_status_phase{phase="Running"}) == 1))))  / ((sum((kube_node_labels{label_node_role_kubernetes_io_compute="true"} == 1) * on(node) group_left() (sum by(node) (kube_node_status_allocatable_cpu_cores))))) * 100`

## Is an upgrade task required and are there additional steps needed to test this?
- [x] Tested Installation
- [x] Tested Uninstallation
- [ ] Tested or Created follow on for Upgrade

PR #1037, opened 2019-09-26
Intly 3312 cherry pick

StevenTobin:
## Additional Information
Cherry-pick of https://github.com/integr8ly/installation/pull/1031

- [x] Tested Installation
- [x] Tested Uninstallation


david-martin:
Upgrade task for this already exists in upgrade playbook https://github.com/integr8ly/installation/blob/v1.5/playbooks/upgrades/upgrade.yaml#L132-L139, but will need to be tested after all other upgrade things removed. That will happen in a follow up PR

PR #1038, opened 2019-09-26
INTLY-3562 check customer admin

wojta:
## Additional Information
https://issues.jboss.org/browse/INTLY-3562

## Verification Steps
1. Install Integr8ly cluster 1.4.1, don't log in to the cluster
2. Run the upgrade playbook for 1.5.0 or newer
3. It should fail quickly with message that "customer-admin" doesn't exist
4. Login as new "customer-admin" user on the cluster
5. Run the upgrade playbook for 1.5.0 or newer
6. It should go through the upgrade process

It's upgrade only task. It shouldn't break installation.

- [ ] Tested Installation
- [ ] Tested Uninstallation

### DIFF
--- a/playbooks/upgrades/upgrade.yaml
+++ b/playbooks/upgrades/upgrade.yaml
@@ -4,8 +4,11 @@
   tasks:
     - include_role:
         name: openshift
-        tasks_from: set_master_vars
+        tasks_from: set_master_vars          
       when: run_master_tasks | default(true) | bool
+    - include_role:
+        name: openshift
+        tasks_from: check_customer_admin
 
 # Required for Ansible Tower installs that need to login via oc as a prerequisite
 - import_playbook: "../openshift.yml"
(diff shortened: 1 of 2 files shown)


david-martin:
@wojta There's a lot of changes showing up with this PR.
I'd suggest PR'ing against master rather than v1.5

wojta:
Please check also sensibility of Ansible playbook code. I'm really newbie in this.

david-martin:
closing this without merging for 2 reasons.

- Only relevant for 1.4.1 to 1.5.0 upgrade, which has already been signed off. (more info in https://issues.jboss.org/browse/INTLY-3562). Even then, it will only happen if the customer-admin user doesn't exist in openshift yet i.e. hasn't logged in
- This is not really the best way to fix the problem. A better more robust fix is to check the return code of commands [like this](https://github.com/integr8ly/installation/blob/v1.5/playbooks/upgrades/3scale_upgrade_2.5_to_2.6.yml#L705-L708) for 0 instead of failing if there's any stderr output. Warning output in stderr is fine when applying a role  to a user who doesn't exist yet.

PR #1051, opened 2019-10-04
fix: Updating code owners

davidffrench:
## Additional Information
Reducing the number of code owners


davidffrench:
@maleck13 @david-martin Looking for your input here on who you think should be code owners?

david-martin:
This will reduce the amount to people that get mails for every PR, but also put the onus on the right people to ensure PRs have all due consideration before merging. 👍

davidffrench:
Thanks @david-martin . Do you think it is the right people? Is there anyone you can think of that I am missing?

david-martin:
@davidffrench Sure. If we find someone else needs owner rights, we'll discuss then.

davidffrench:
Thanks @david-martin

PR #1052, opened 2019-10-04
make sure the manifest file will include mobile services

wei-lee:
https://issues.jboss.org/browse/INTLY-3791


psturc:
Testing on RHPDS...

PR #1057, opened 2019-10-10
[WIP] ups update

fh-automatron:


secondsun:
@mikenairn Can I get some help with this?

mikenairn:
@secondsun Yeah, i was talking to @grdryn earlier about some of the mobile jobs that are generating these PRs. Really we shouldn't need to PR the operator and ups itself separately. Will try and work that out tomorrow and then we can look at getting the test working for the operator update so we can try and get it merged back.

secondsun:
@mikenairn Since we only need the operator, in theory, should we close this PR?

secondsun:
@mikenairn Even if we don't we're about to tag a new UPS so this PR is soon obsolete

mikenairn:
@secondsun I never got round to looking at changing the jobs that creates these PRs. If we close this currently it will just re-open it again. We should ignore these one though(ups), and only try and get the one with the operator update merged back, this one https://github.com/integr8ly/installation/pull/1077

We can manually apply the ups version change that's in this PR to that one for the moment since both things are required in the RHMI installer and are supposed to be generated in a single PR.

Do you want to wait until this new release of UPS/UPS Operator are created before we try to fix this? Seems no point in doing anything with this if you know it's about to change.

secondsun:
So simplicity let's go ahead and get #1077 merged.  This version of UPS and that version of the operator are tested together.  We don't expect any problems with the operator and the coming soonversion of UPS, but we haven't tested that yet.

secondsun:
@mikenairn @grdryn https://github.com/integr8ly/installation/pull/1099 has the update to #1077 I can't make the update myself because of permissions

mikenairn:
/hold

This should not be it's own PR. Automation needs updated to stop this PR re-opening, applying a hold just now to show it should be ignored.

mikenairn:
Changes that are being added here have been cherry picked into https://github.com/integr8ly/installation/pull/1077

PR #1065, opened 2019-10-14
Modified CronJobsFailed PromQL query

obrienrobert:
## Additional Information
https://issues.jboss.org/browse/INTLY-3715

The original `CronJobsFailed` PromQL only gave a count of the number of failed backup jobs, and did not take failed jobs that then passed (i.e the query would still report on any prior failed attempts) into consideration.

This PR contains an updated version of the query that addresses the above issues by checking if the latest backup job that was executed for each job was successful or not.

## Verification Steps
- Login to the the following cluster, using the credentials supplied: https://master.robrien-1af4.open.redhat.com

- In the `openshift-integreatly-backups`, [manually](https://github.com/RHCloudServices/integreatly-help/tree/0ba42ae6d1d9287a8de1709a22024249175a5ca3/sops/backup) run a backup job for 2 different namespaces. As the `s3-credentials` secret contains incorrect credentials, the backups should fail (the alert will take 5 minutes to fire). You can check this by looking at the pods, which should report a error state. The `CronJobsFailed` alert in Prometheus should also trigger, and display the name of the job/s that failed

- Update the `s3-credentials` secret so that it contains correct credentials

- Manually run another backup for 1 of the failed jobs from above and ensure that it is successful.

- Verify that the `CronJobsFailed` alert now displays only 1 failed job, and does not display any failed jobs for which the last run was successful

### Installation Verification
- https://integreatly-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/view/PDS/job/pds-install/1403/console

- [ ] Yes
- [x] No

- [x] Tested Installation
- [ ] Tested Uninstallation

### DIFF
--- a/roles/backup/templates/backup-monitoring-alerts.yml.j2
+++ b/roles/backup/templates/backup-monitoring-alerts.yml.j2
@@ -39,8 +39,7 @@ spec:
           sop_url: https://github.com/RHCloudServices/integreatly-help/blob/master/sops/alerts_and_troubleshooting.md
           message: 'Job {{ '{{' }} $labels.namespace {{ '}}' }}/{{ '{{' }} $labels.job {{ '}}' }} has failed'
         expr: >
-          sum(kube_job_status_failed{namespace="{{ backup_namespace }}"})
-          > 0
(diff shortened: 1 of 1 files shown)


davidffrench:
/approve

based on Rachaels verification

PR #1071, opened 2019-10-15
[WIP] middleware-monitoring update

fh-automatron:


mikenairn:
/cc @david-martin @pb82

Might be worth looking into how we can select reviewers/approvers based on lines changed in that file

david-martin:
> Might be worth looking into how we can select reviewers/approvers based on lines changed in that file

Is that possible via openshift-ci-robot or github codeowners?

mikenairn:
@david-martin Was more a note to myself. It is possible to add OWNERS for different directories with both (prow OWNERS and GH CODEOWNERS), but not sure about actually checking what's changed in an individual file, probably not.

david-martin:
This has already been verified on the 1.5 stream in https://github.com/integr8ly/installation/pull/1073

@pb82  Any concerns with merging this to master?

pb82:
@david-martin No, it looks like the main change in 0.0.27 is this: https://github.com/integr8ly/application-monitoring-operator/commit/78284a4034ad0bc132bb722837c4320846c25ff4

There is also a later tag (0.0.28) but no release for it.

PR #1079, opened 2019-10-16
Update release script to reset upgrade playbook on final release

mikenairn:
## Additional Information
Update release script to reset upgrade playbook on final release (Anything that isn't a CR or ER)

## Verification Steps

* Checkout this branch and create a new test release branch from it and push to your fork (origin in my case):
`git checkout -b v9.9 && git push origin v9.9`
* Create a new rc1, and verify it creates a new rc1 tag from the HEAD of the v9.9 branch and upgrade playbook isn't modified:
`REMOTE=origin ./scripts/release.sh -b v9.9 -r release-9.9.0-rc1`
* Create a new rc2, and verify it creates a new rc2 tag from the HEAD of the v9.9 branch and upgrade playbook isn't modified:
`REMOTE=origin ./scripts/release.sh -b v9.9 -r release-9.9.0-rc2`
* Create a new final release, and verify it creates a new release tag from the HEAD of the v9.9 branch and upgrade playbook and variables are reset:
`REMOTE=origin ./scripts/release.sh -b v9.9 -r release-9.9.0`

### Upgrade Verification
N/A

/cc @maleck13


playbooks/upgrades/upgrade.template.yaml
```diff
@@ -0,0 +1,27 @@
...
+        tasks_from: upgrade
+      vars:
+        from_versions:
+          - "PREVIOUS_VERSION"
+
+# Individual product upgrade tasks get added here
+
+#Update product version (should always be last)
+- import_playbook: "../generate-customisation-inventory.yml"
+- import_playbook: "../mobile_generate_manifest.yml"
```
mikenairn: Not sure exactly what this should look like at the moment. Will likely need to iterate on this.

maleck13:
@mikenairn the code changes look good. Will test it out tomorrow

playbooks/upgrades/upgrade.template.yaml
```diff
@@ -0,0 +1,27 @@
...
+
+- hosts: localhost
+  gather_facts: yes
+  tasks:
+    - include_role:
+        name: prerequisites
+        tasks_from: upgrade
+      vars:
+        from_versions:
+          - "PREVIOUS_VERSION"
```
tremes: I was looking at current upgrade yaml file and there's also following:
```
    - set_fact:
        upgrade_webapp_image: quay.io/integreatly/tutorial-web-app:{{ webapp_version }}
        upgrade_webapp_operator_image: quay.io/integreatly/tutorial-web-app-operator:{{ webapp_operator_release_tag }}
        upgrade_fuse_image_tag: 1.7
```
I am just guessing but don't we need some placeholder for these too?

tremes:
Added one question but otherwise LGTM.

playbooks/upgrades/upgrade.template.yaml
mikenairn: Whatever this was doing previously will be dealt with elsewhere. We want to keep the upgrade playbook fairly generic, anything to do with specific products will be dealt with by a tasks included from the products role.

Note, this is just a start, we will be iterating on this over the next sprint. It's currently isolated form the current update playbook so will do no harm to merge as is.

playbooks/group_vars/all/upgrade.yml
```diff
@@ -0,0 +1,3 @@
+upgrade_from_version: release-1.5.1
```
pmccarthy: @mikenairn I'm assuming this file will be managed by Delorean? Just wondering how we'll ensure the release version is always up to date.

playbooks/upgrade.yml
```diff
@@ -0,0 +1,21 @@
...
+    set_fact: upgrade_{{ item }}=true
+    with_items: "{{ upgrade_product_roles }}"
+
+  - name: Upgrade products
+    include_role:
+      name: "{{ item }}"
+      tasks_from: "{{ upgrade_tasks_from }}"
+    with_items: "{{ upgrade_product_roles }}"
+
+  - name: Some Special webapp only upgrade thing
```
pmccarthy: Might want to change the task name to something more informative as it will show up on our upgrade runs (thinking of CS-SRE when they do rollouts to prod) :)

mikenairn: Yep, this should be removed. Was here as an example of how you could add tasks using the `when: upgrade_webapp|bool` conditions if needs be

playbooks/group_vars/all/upgrade.yml
mikenairn: The release script will reset all these variables when we do a final release (Anything that isn't a CR or ER), and sets this particular value "upgrade_from_version" to the release we have just created ready for the next one see https://github.com/integr8ly/installation/pull/1079/files#diff-0d28f82f2d1cc0deb12146b212778bc9R119-R124

scripts/release.sh
```diff
@@ -46,9 +50,32 @@ echo "Flags:
...
+    MAJOR_VERSION=${BASH_REMATCH[1]}
+    MINOR_VERSION=${BASH_REMATCH[2]}
+    PATCH_VERSION=${BASH_REMATCH[3]}
+    LABEL_VERSION=${BASH_REMATCH[4]}
+else
+    echo "Invalid release tag $releaseTag"
+    exit 1
+fi
+
+upgradeTasksFrom = "upgrade"
```
tremes: This should probably be `upgradeTasksFrom="upgrade"`. Because I was getting:
```
./scripts/release.sh: line 63: upgradeTasksFrom: command not found
./scripts/release.sh: line 67: upgradeTasksFrom: command not found
```

scripts/release.sh
```diff
@@ -46,9 +50,32 @@ echo "Flags:
...
 
+if [[ $PATCH_VERSION -gt 0 ]]; then
+    RELEASE_TYPE="patch"
+    upgradeTasksFrom = "upgrade_patch"
```
tremes: Same here.

tremes:
I have gone through the verification steps and the reset looks like this https://github.com/tremes/installation/commit/d637154b382c6c91db53296cc0b5c5c71bb79920. Is that OK - especially the second change?

scripts/release.sh
mikenairn: Yep, will update

mikenairn:
> I have gone through the verification steps and the reset looks like this tremes@d637154. Is that OK - especially the second change?
@tremes Not quite, "upgrade_tasks_from" should always have a value of "upgrade" or "upgrade_patch"

Could you try it again now that I've updated those variable assignments.

tremes:
@mikenairn OK. I've tried again and it looks better I guees https://github.com/tremes/installation/commit/1f0be52df18080bb89583f4acc269feaf1a6590c.

mikenairn:
This has now been updated to do the following:

* Reset upgrade playbook and variables on final release
* Upgrade playbook has default tasks added from template, including calling upgrade_images tasks for all products marked as being upgraded
* Upgrade variables file will be automatically updated by DeLorean jobs
* New task in each product/component role called upgrade_images (Not implemented)

Ran a test as follows:

```
 1271  git checkout -b v9.10 && git push origin v9.10
 1272  REMOTE=origin ./scripts/release.sh -b v9.10 -r release-9.10.0-rc1
 1273  REMOTE=origin ./scripts/release.sh -b v9.10 -r release-9.10.0-rc2
 1274  REMOTE=origin ./scripts/release.sh -b v9.10 -r release-9.10.0
 1275  REMOTE=origin ./scripts/release.sh -b v9.10 -r release-9.10.1-rc1
 1276  git push origin v9.10 < Modifed the upgrade playbook here
 1277  REMOTE=origin ./scripts/release.sh -b v9.10 -r release-9.10.1-rc2
 1279  REMOTE=origin ./scripts/release.sh -b v9.10 -r release-9.10.1
```

Result can be seen here https://github.com/mikenairn/installation/commits/v9.10

playbooks/group_vars/all/upgrade.yml
mikenairn: Hmmm, just thinking about this more, this will need to be kept up to date on the current release branch (v1.5), as well as master, which depending on if there is another release branch already in progress (v1.6) should be updated with those releases :thinking: 

Doesn't stop this being merged, but is something we might need to work on.

pmccarthy: We could always have `upgrade_from_version` set to an empty value on master and only set it on release specific branches?

playbooks/upgrade.yml
```diff
@@ -0,0 +1,45 @@
+# This file is re-generated after each release!
```
pmccarthy: Is there a need for this file to exist given that we already have a templated version? I would have thought we would auto generate `upgrade.yml` based on the template `upgrade.template.yaml`.

Also notice that we use both `yml` and `yaml` in the names, don't think it matters but would be good to try stay consistent

roles/3scale/tasks/upgrade_images.yml
```diff
@@ -0,0 +1,4 @@
+---
+#ToDo Implement patch/CVE update steps as described https://github.com/RHCloudServices/integreatly-help/blob/master/sops/cves/applying-cve-updates.md
```
pmccarthy: Should this comment only reference CVE's now? If so we could change the working to `#ToDo Implement CVE update steps as described ......`

playbooks/upgrade.yml
mikenairn: We do generate that file at release time yes, and i suppose you can argue that it doesn't need to exist on master if this method is only every used for patch release upgrades, but i would imagine going forward we would really want to deal with patch and minor releases in the same way, in which this case this file should exist to add upgrade tasks for the next minor release.

I think that's what you are asking anyway?

yml vs yaml is everywhere, will fix on these to be the same.

scripts/release.sh
```diff
@@ -81,12 +106,20 @@ a release with that name already exists
...
+#commit the change, tag
 git commit -am "release manifest version  update for ${releaseTag}"
-git push ${REMOTE} ${baseBranch}
-#tag and push
 git tag ${releaseTag}
+
+#reset upgrade playbook and variables if this is the final release
+if [[ -z $LABEL_VERSION ]]; then
+    echo "resetting upgrade playbook and variables after final release $releaseTag"
+    cp scripts/upgrade.template.yml playbooks/upgrade.yml
```
pmccarthy: Could we instead copy over `playbooks/upgrades/upgrade.template.yaml` ? We seem to have duplicate upgrade template files for scripts and playbooks. Would be good if we only had one.

roles/3scale/tasks/upgrade_images.yml
mikenairn: Sure, it's a comment that should be removed entirely by the end of the sprint in all files anyway, but i can update it for you.

scripts/release.sh
mikenairn: Hmm , yes, there aren't supposed to be 2. Will get rid of one

playbooks/upgrade.yml
pmccarthy: Sure, I suppose what I was really getting at is the fact that the `upgrade.yml` file and `upgrade.template.yml` file seem to be identical so wasn't sure if we only needed the templated version. Don't feel strongly about it either way.

playbooks/upgrade.yml
```diff
@@ -0,0 +1,45 @@
...
+# If you need to add tasks that should run on every upgrade, update the template here ../../scripts/upgrade.template.yml
```
pmccarthy: We seem to have `playbooks/upgrade.yml` and `playbooks/upgrades/upgrade.yaml`, both of these files are the same. The one tower expects is `playbooks/upgrades/upgrade.yaml ` so perhaps we can remove `playbooks/upgrade.yml`? Unless I'm missing something.

playbooks/upgrade.yml
mikenairn: They are the same because nothing has been changed in the upgrade.yml since it was generated. As development progresses and products/components are being updated with upgrade tasks you would expect that to change, and when a release is finalised, it resets it.

playbooks/upgrade.yml
mikenairn: This was my mistake, i checked in changes to playbooks/upgrades/upgrade.yaml unintentionally, reverted now. For the time being we will keep both of these(until after v1.6 i screated), but eventually we will only have playbooks/upgrade.yml

playbooks/group_vars/all/upgrade.yml
mikenairn: Possibly. So long as all releases are created with the release script, we should be able to make it update the current and previous release branches correctly. Not sure we should spend any more time on this for the moment though, we shoudl just keep an eye on the value to ensure it is actually what we expect it to be when creating a release.

PR #1091, opened 2019-10-30
DOCS for monitoring

R-Lawton:
## Additional Information
The following WIP PR is for the JIRA [INTLY-3670](https://issues.jboss.org/browse/INTLY-3670) Creating docs for monitoring

### DIFF
--- a/docs/monitoring_docs.adoc
+++ b/docs/monitoring_docs.adoc
@@ -1,0 +1,21 @@
+### Integreatly monitoring 
+The following document will cover different aspects of Integreatly monitoring including:
+
+
+:toc:
+
+== How to access the montoring stack?
+
+== Who can access the monitoring stack?
+
+== What metrics are available in the montioring stack?
+
+== How are metrics calculated?
+
+== What alerting is available?
+
+== How is alerting setup?
+
+
+
+
--- a/docs/monitoring_docs.adoc
+++ b/docs/monitoring_docs.adoc
@@ -1,11 +1,11 @@
-= Integreatly monitoring 
-The following document will cover different aspects of Integreatly monitoring.
+= Integreatly Monitoring 
+The following document will cover different aspects of monitoring an Integreatly cluster.
 
 :toc:
 == Intended audience and prerequisites
 The document is intended for the following users:
 
-* SRE managing integreatly cluster
+* SRE managing an Integreatly/RHMI cluster
 * Integreatly users:
 ** Workshop user
 ** RHMI customer
@@ -14,13 +14,12 @@ The document is intended for the following users:
 ** RH engineering
 ** Upstream contributor
 
-Its assumed that the reader knows about Integreatly, what it is and what Integreatly consists of.
+It is assumed that the reader knows what Integreatly is, and the features of Integreatly.
 
 NOTE: Certain dashboards and alerts may not be visible if users don't have appropriate permissions.
 
-
 == Middleware Monitoring Architecture
-In the middleware monitoring architecture there are two monitoring stacks available.
+In the middleware monitoring architecture there are two monitoring stacks running.
 
 image::arch.png[]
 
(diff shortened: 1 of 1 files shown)


docs/monitoring_docs.adoc
```diff
@@ -0,0 +1,66 @@
...
+
+== How to acces the middleware monitoring stack?
+To access the monitoring stack login into your clusters Openshift master console.
+
+HOW TO ACCESS VIA UI 
+
+
+== What metrics are available in the montioring stack?
+
+=== Kube stat metrics
```
david-martin: Kube state

docs/monitoring_docs.adoc
```diff
@@ -0,0 +1,66 @@
...
+
+Middleware Services need to satisfy some criteria for metrics to be scraped:
+
+* The namespace has a specific label with a specific value i.e. monitoring-key=middleware. This is to ensure we only monitor the namespaces we care about.
+* The namespace has a ServiceMonitor resource that defines the Services or Pods to scrape metrics from. This ServiceMonitor must also have a label of monitoring-key=middleware on it.
+
+== How to acces the middleware monitoring stack?
+To access the monitoring stack login into your clusters Openshift master console.
+
+HOW TO ACCESS VIA UI
```
david-martin: Providing a UI way & CLI way for finding the routes to relevant monitoring components would be good here.

Then info on how to login via UI and the permissions required.

docs/monitoring_docs.adoc
```diff
@@ -0,0 +1,66 @@
...
+
+kube-state-metrics is about generating metrics from Kubernetes API objects without modification. This ensures that features provided by kube-state-metrics have the same grade of stability as the Kubernetes API objects themselves. In turn, this means that kube-state-metrics in certain situations may not show the exact same values as kubectl, as kubectl applies certain heuristics to display comprehensible messages. kube-state-metrics exposes raw data unmodified from the Kubernetes API, this way users have all the data they require and perform heuristics as they see fit.
+
+The metrics are exported on the HTTP endpoint /metrics on the listening port (default 80). They are served as plaintext. They are designed to be consumed either by Prometheus itself or by a scraper that is compatible with scraping a Prometheus client endpoint. You can also open /metrics in a browser to see the raw metrics.
+
+Exposed metrics:
+Per group of metrics there is one file for each metrics. See each file for specific documentation about the exposed metrics:
+https://github.com/kubernetes/kube-state-metrics/tree/master/docs
+
+
```
david-martin: As discussed, node-exporter metrics can be mentioned here and linked to.

Also, it might be worth have a cmd to get the current config that's used for federation i.e. this file https://github.com/integr8ly/installation/blob/8c28f1865f28cc3fb86243b4cbeeed37c60ed934/roles/middleware_monitoring_config/templates/jobs/openshift_monitoring_federation.j2#L11-L18

docs/monitoring_docs.adoc
```diff
@@ -0,0 +1,112 @@
...
+==== AlertManager
+AlertManager receives any active alerts from Prometheus, and sends them to configured receivers based on the severity. Only currently active alerts will appear in AlertManger's web console.
+
+==== Grafana
+Grafana is configured via the Grafana Operator with dashboards from middleware namespaces. It leverages the GrafanaDashboard custom resource.
+
+Middleware Services need to satisfy some criteria for metrics to be scraped:
+
+* The namespace has a specific label with a specific value i.e. monitoring-key=middleware. This is to ensure we only monitor the namespaces we care about.
```
wojta: I would prefer if you use some formatting there to increase readability, to show that it's some kind of code,  configuration or filter...
like i.e. `monitoring-key=middleware`

docs/monitoring_docs.adoc
```diff
@@ -0,0 +1,112 @@
...
+* The namespace has a ServiceMonitor resource that defines the Services or Pods to scrape metrics from. This ServiceMonitor must also have a label of monitoring-key=middleware on it.
+
+== How to access the middleware monitoring stack?
+To access the monitoring stack via UI and who has permissions:
+
+1. To access the openshift master console follow the link provided in the provisioning email.
```
wojta: Shouldn't be this more generic? This is case of RHPDS, probably not other cases, although I'm also familiar only with RHPDS.

david-martin: Good point.
I'd say this subsection can be removed and skip onto `Navigating to the monitoring stack resources through the UI`. The assumption is the user has some account on openshift already and knows how to login. We don't need to be concerned with how they got that account

docs/monitoring_docs.adoc
```diff
@@ -0,0 +1,112 @@
+### Integreatly monitoring 
+The following document will cover different aspects of Integreatly monitoring including:
+
```
david-martin: @R-Lawton It might be good to have a section at the top about who the intended audience(s) is for this, and what assumptions are made.
For example, assume the reader knows what Integreatly is & whats in it. Also, they have an account if they are reading sections about viewing dashboards and alerts. (They might not have correct perms, so that's a good thing to mention below)

david-martin: Appropriate audiences I would argue are:

- Developer of Integreatly (e.g. RH engineering or upstream contributor)
- SRE managing an Integreatly cluster
- Any user of Integreatly (e.g. a workshop user, or an RHMI customer)
- Developer/User of any individual component that's in Integreatly (e.g. fuse/syndesis)

docs/monitoring_docs.adoc
R-Lawton: ui and cli documented currently doing permissions required

docs/monitoring_docs.adoc
R-Lawton: yes will do

docs/monitoring_docs.adoc
R-Lawton: okey dokes ill remove it

docs/monitoring_docs.adoc
R-Lawton: yes good idea

docs/monitoring_docs.adoc
R-Lawton: Where would i find this command i have had a look and i cant find info about it

david-martin: Try this
```oc get secret additional-scrape-configs -n middleware-monitoring  --template '{{index .data "integreatly.yaml"}}' | base64 --decode | grep -A 10 "params:"```

The namespace could also be `openshift-middleware-monitoring` depending on the cluster

R-Lawton: perfect thank you

david-martin:
/approve

@wojta will defer to you on an lgtm

PR #1094, opened 2019-10-30
[WIP] amq-online update

fh-automatron:


trepel:
AMQ Online 1.3.0 does not contain all the fixes currently in 1.2 stream. I think that it is fine since this is not going to be released sooner than in RHMI 1.6. By that time everything from 1.2 stream will be ported to 1.3 as well. Just saying, somebody with more insight might think this being an issue.

mikenairn:
/close

Will re-generate

PR #1096, opened 2019-10-30
Explicitly remove the mdc oauthclient

david-martin:
## Additional Information
During uninstall, all oauthclients are removed https://gist.github.com/david-martin/ca1fdd8311a67d99818eacd022e2297c

This is incorrect. Only the mdc oauthclient should be removed.

## Verification Steps

### Installation Verification
### Upgrade Verification

- [ ] Yes
- [x] No

- [x] Tested Installation
- [x] Tested Uninstallation

### DIFF
--- a/roles/mdc/tasks/uninstall.yml
+++ b/roles/mdc/tasks/uninstall.yml
@@ -4,7 +4,6 @@
   register: mdc_resource_list_cmd
   failed_when: false
   with_items:
-    - oauthclient
     - mobiledeveloperconsole
     - keycloakrealm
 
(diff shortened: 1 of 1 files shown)


david-martin:
full log https://gist.github.com/david-martin/bf07be1914c0aa841a692cbe707e44e7

david-martin:
@grdryn The process in the PR template was updated recently to put more upfront verification & proof on the PR creator that the PR does what it says it should do, hence the attached logs.
If you're happy with the code, and the log output, then job done.

It does need a second approver thought. ping @wei-lee @laurafitzgerald

laurafitzgerald:
@david-martin This looks good to me. One comment would be that if this variable https://github.com/integr8ly/installation/blob/master/roles/mdc/defaults/main.yml#L5 is changed between versions then the oauthclient wont be found for an uninstall. Should we leave some comment with the variable that it shouldn't be changed?

david-martin:
> Should we leave some comment with the variable that it shouldn't be changed?

I don't think that's necessary. 
Uninstall is only done during development & testing.
Typically the same version is used to uninstall as to install, or at most 1 version later if testing upgrades. 
Uninstall will not be run in a production cluster.

PR #1099, opened 2019-10-31
Updating UPS version

secondsun:
@mikenairn I don't have write access to the repo so I can't change the PR, but this should be the only change necessary.


mikenairn:
/close

@secondsun I rolled the change into this PR https://github.com/integr8ly/installation/pull/1077

PR #1104, opened 2019-11-01
INTLY-3980 - new codeready task for image upgrades

tremes:
## Verification Steps

### Installation Verification
### Upgrade Verification

- [ ] Yes
- [ ] No

- [ ] Tested Installation
- [ ] Tested Uninstallation


roles/codeready/tasks/upgrade.yml
```diff
@@ -0,0 +1,3 @@
+
```
pmccarthy: Not sure we need an `upgrade.yml` task in this role any longer. Should be ok to just have an `upgrade_images.yml` task that gets called by the main `upgrade.yml` playbook

tremes: OK sounds reasonable. I'll check the main playbook.

tremes: Updated. Can you please check once again @pmccarthy ?

tremes:
Fixed formatting which was broken similarly to https://github.com/integr8ly/installation/pull/1114/f

PR #1106, opened 2019-11-01
Adding image upgrade logic to 3Scale role

pmccarthy:
## Additional Information
Adding 3Scale image upgrade logic along with a placeholder upgrade_patch task in the 3Scale role

## Verification Steps
Run upgrade.yml playbook:

On RHPDS Environment
```
ansible-playbook -i inventories/pds.template playbooks/upgrade.yml
```

### DIFF
--- a/roles/3scale/tasks/upgrade_images.yml
+++ b/roles/3scale/tasks/upgrade_images.yml
@@ -1,4 +1,7 @@
 ---
-#ToDo Implement CVE image update steps as described https://github.com/RHCloudServices/integreatly-help/blob/master/sops/cves/applying-cve-updates.md
-- debug:
-    msg: "TODO Implement me!!"
+
+- name: Import Image Streams
+  shell: oc import-image {{ patch_image_stream_item }} -n {{ threescale_namespace }}
+  with_items: "{{ threescale_image_streams }}"
+  loop_control:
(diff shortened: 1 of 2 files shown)


roles/3scale/tasks/upgrade_images.yml
```diff
@@ -1,4 +1,7 @@
 ---
-#ToDo Implement CVE image update steps as described https://github.com/RHCloudServices/integreatly-help/blob/master/sops/cves/applying-cve-updates.md
-- debug:
-    msg: "TODO Implement me!!"
+
+- name: Import images streams
+  shell: oc import-image {{ patch_image_stream_item }} -n {{ threescale_namespace }}
+  with_items: "{{ threescale_image_streams }}"
+  loop_control:
```
pmccarthy: Need to do this `loop_control` block as we're looping within a loop. 

Without it the following warning is thrown:
```
[WARNING]: The loop variable 'item' is already in use. You should set
the `loop_var` value in the `loop_control` option for the task to 
something else to avoid variable collisions and unexpected behavior.
```

mikenairn: Good to know :+1:

tremes:
Verifying

PR #1108, opened 2019-11-01
amq-online update

fh-automatron:

### DIFF
--- a/inventories/group_vars/all/manifest.yaml
+++ b/inventories/group_vars/all/manifest.yaml
@@ -25,7 +25,7 @@ che_git_url: https://github.com/redhat-developer/codeready-workspaces-deprecated
 
 #controls whether enmasse is installed or not
 enmasse: True
-enmasse_version: '1.2.3.GA'
+enmasse_version: '1.3.1.GA'
 enmasse_git_url: https://github.com/jboss-container-images/amq-online-images.git
 
 #controls whether fuse is installed or not


playbooks/group_vars/all/upgrade.yml
```diff
@@ -1,3 +1,3 @@
-# DO NOT EDIT This file is auto generated!!
 upgrade_from_version: release-1.5.1
-upgrade_product_roles: []
+upgrade_product_roles:
+- amq-online
```
aidenkeating: @mikenairn just noticed this variable, from what i can see the upgrade automation will try to run a role with the name `amq-online`. the role in this repo is still called `enmasse`. are there plans to update this role name?

also, is/will this variable and upgrade logic be used as part of the 1.6.0 release?

tremes: Good point. One related note - role names are (or will be) used as variables which means a role name cannot contain a hyphen.

mikenairn: Bah, bloody naming. This is generated in DeLorean, so will have to update there, we had others with similar problems, names not matching the product names.

These aren't actually going to be used in 1.6 minor release, so probably not a major issue, but if you give me a minute i can update DeLorean to fix it.

/hold

mikenairn: /hold cancel

Resolved now, and updated here https://github.com/integr8ly/ci-cd/pull/340

mikenairn:
@aidenkeating @maleck13 I see this has been approved, but does it need any upgrade steps added to https://github.com/integr8ly/installation/blob/master/playbooks/upgrades/upgrade.yaml, or is this enough  https://github.com/integr8ly/installation/blob/master/playbooks/upgrades/upgrade.yaml#L68

aidenkeating:
@mikenairn good point, i ran the upgrade on pds (we'll be doing osd also) and the upgrade did happen. although i just noticed that in the existing upgrade playbook we call the main task and not the upgrade task. 

@davidffrench was this intentional?

https://github.com/integr8ly/installation/blob/master/playbooks/upgrades/upgrade.yaml#L57-L59

davidffrench:
> @mikenairn good point, i ran the upgrade on pds (we'll be doing osd also) and the upgrade did happen. although i just noticed that in the existing upgrade playbook we call the main task and not the upgrade task.
> 
> @davidffrench was this intentional?
> 
> https://github.com/integr8ly/installation/blob/master/playbooks/upgrades/upgrade.yaml#L57-L59

Yes, 100% intentional. After looking through the playbooks AMQ Online use to install and upgrade, it is the same deploy playbook so there is no reason for us to duplicate the work in another upgrade playbook

davidffrench:
I just looked into it and it looks to still be valid.

https://access.redhat.com/documentation/en-us/red_hat_amq/7.5/html-single/installing_and_managing_amq_online_on_openshift/index#upgrading-enmasse-using-ansible-messaging
https://access.redhat.com/documentation/en-us/red_hat_amq/7.5/html-single/installing_and_managing_amq_online_on_openshift/index#installing-using-ansible-messaging

https://github.com/jboss-container-images/amq-online-images/blob/1.3.1.GA/templates/ansible/playbooks/openshift/deploy_all.yml

aidenkeating:
tested installation on cluster `b0ec` with wt1, @ciaranRoche is testing upgrade

ciaranRoche:
tested upgrade on PDS with wt1. Small issue found, unrelated to this work.

playbooks/group_vars/all/upgrade.yml
```diff
@@ -1,3 +1,3 @@
-# DO NOT EDIT This file is auto generated!!
 upgrade_from_version: release-1.5.1
```
david-martin: @ciaranRoche  @aidenkeating 
The latest release is 1.5.2. Was testing based on upgrade from 1.5.1 or 1.5.2?

aidenkeating: @david-martin The upgrade testing was based on 1.5.2, specified in the `playbooks/upgrades/upgrade.yaml` file. I have removed bumps to the modified file to avoid confusion

aidenkeating:
@david-martin @jessesarn @mikenairn Any objections to this being merged? Looks like it just needs a lgtm label

mikenairn:
@aidenkeating None from me if we are covered upgrade wise and this is all working

PR #1114, opened 2019-11-05
INTLY-4011 - AMQ patch upgrade task

tremes:
## Verification Steps

### Installation Verification
### Upgrade Verification

- [ ] Yes
- [ ] No

- [ ] Tested Installation
- [ ] Tested Uninstallation


tremes:
@mikenairn @pmccarthy can you please check?

roles/enmasse/tasks/upgrade_images.yml
```diff
@@ -1,4 +1,25 @@
----
-#ToDo Implement CVE image update steps as described https://github.com/RHCloudServices/integreatly-help/blob/master/sops/cves/applying-cve-updates.md
-- debug:
-    msg: "TODO Implement me!!"
+-
```
pmccarthy: Seems to be some indentation issues with this file? 

For example:
```
-
  name: "get all deployments"
```

Should this not be:
```
- name: "get all deployments"
```
?

tremes: Oh yeah you're probably right. I'll check tomorrow.Thx

PR #1116, opened 2019-11-05
treat master url string as undefined if empty

aidenkeating:
## Additional Information
currently, empty string versions of the master public url are still being treated as defined. meaning the default filter will never be used. however, in some situations, we want it to be used.

this change ensures that empty strings are treated the same as an undefined variable, to ensure the default filter works as expected.

## Verification Steps
- run the installer on pds. ensure the webapp has it's OPENSHIFT_HOST variable set

### Installation Verification
### Upgrade Verification

- [ ] Yes
- [ ] No

- [ ] Tested Installation
- [ ] Tested Uninstallation

### DIFF
--- a/roles/webapp/tasks/provision-webapp.yml
+++ b/roles/webapp/tasks/provision-webapp.yml
@@ -5,7 +5,7 @@
   failed_when: false
 
 - name: Construct openshift host string
-  shell: echo {{ openshift_master_public_url | default(openshift_asset_url) }} | cut -d '/' -f3
+  shell: echo {{ openshift_master_public_url | default(openshift_asset_url, true) }} | cut -d '/' -f3
   register: openshift_host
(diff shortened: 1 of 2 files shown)


pmccarthy:
@aidenkeating Looks like we use the same logic for the webapp upgrade task: https://github.com/integr8ly/installation/blob/master/roles/webapp/tasks/upgrade.yaml#L3

No need to re-validate as we don't upgrade on PSD but I think it's worth fixing this too just in case

tonyxrmdavidson:
Successful installation :     https://integreatly-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/pds-install/1483/console

Services provisioned correctly :
Screenshot 2019-11-05 at 13 40 26

PR #1121, opened 2019-11-06
INTLY-4060 - don't install backups to PoC by default

tremes:
## Verification Steps
As the verifier of the PR the following process should be done:

### Installation Verification
- Ensure the author of the PR has attached a log of the installation run from his branch to the jira or pr and check that it exited as expected.
- Verify the fresh installation is correct on cluster provided by PR author 
### Upgrade Verification
- After installation verification, notify the PR author to begin an upgrade on their cluster
- Ensure the developer of the PR has attached a log of the upgrade run from his branch to the jira or pr and check that it exited as expected. 
- If possible, look at the tasks that ran and see they match the PR
- Verify the upgrade is correct on cluster provided by PR author 

## Is an upgrade task required and are there additional steps needed to test this?

- [ ] Yes
- [ ] No

- [x] Tested Installation
- [ ] Tested Uninstallation
- [ ] Tested or Created follow on for Upgrade

PR #1125, opened 2019-11-07
INTLY-4060 - don't install backups to PoC by default

tremes:
## Verification Steps

### Installation Verification
### Upgrade Verification

- [ ] Yes
- [ ] No

- [ ] Tested Installation
- [ ] Tested Uninstallation


tremes:
This is just a cherry-pick of already merged https://github.com/integr8ly/installation/pull/1121

PR #1130, opened 2019-11-08
Adding rhsso image upgrade logic

pmccarthy:
## Additional Information
Updating RH-SSO upgrade_images task with logic for importing image streams. 

Based on steps outlined here: https://github.com/RHCloudServices/integreatly-help/blob/master/sops/cves/applying-cve-updates.md#red-hat-sso

## Verification Steps

- [ ] Add rhsso to `upgrade_product_roles` list in upgrade group_vars
- [ ] Run upgrade task (assuming on PDS):
```
ansible-playbook -i inventories/pds.template playbooks/upgrade.yml 
```

### DIFF
--- a/roles/rhsso/defaults/main.yml
+++ b/roles/rhsso/defaults/main.yml
@@ -71,3 +71,7 @@ rhsso_resources:
     resources:
       requests:
         memory: 25Mi
+
+rhsso_image_streams:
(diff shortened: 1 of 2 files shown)


pmccarthy:
@tremes Correct, spoke with @mikenairn about this last week and we agreed that this task should only deal with updating the image streams and not manage the keycloak-operator

PR #1097, opened 2019-10-31
WIP : Adding two new PVC checks

alanmoran:
PVCStorageWillFillIn4Days and PVCStorageWillFillIn4Hours

WIP PR to test prediction based monitoring on PVC's

## Is an upgrade task required and are there additional steps needed to test this?

- [ ] Yes
- [x ] No

- [ ] Tested Installation
- [ ] Tested Uninstallation
- [ ] Tested or Created follow on for Upgrade

PR #1132, opened 2019-11-08
Intly 3965 modifcations

R-Lawton:
## Additional Information
The following is a modified version of another PR to keep a record of all the changes being made to fix the errors on the other PR [INTLY-3965](https://github.com/integr8ly/installation/pull/1097)

### DIFF
--- a/roles/middleware_monitoring_config/templates/kube_state_metrics_alerts.yml.j2
+++ b/roles/middleware_monitoring_config/templates/kube_state_metrics_alerts.yml.j2
@@ -208,7 +208,7 @@ spec:
         annotations:
           message: The {{ '{{' }} $labels.persistentvolumeclaim {{ '}}' }} PVC will run of disk space in the next 4 days.
         expr: |
-          (sum by(persistentvolumeclaim, namespace) (kubelet_volume_stats_available_bytes{job="kubelet"} and predict_linear(kubelet_volume_stats_available_bytes{job="kubelet"}[6h], 4 * 24 * 3600)) * on ( namespace) group_left(label_monitoring_key) kube_namespace_labels{label_monitoring_key="middleware"}) <= 0
+          (sum by(persistentvolumeclaim, namespace) (predict_linear(kubelet_volume_stats_available_bytes{job="kubelet"}[6h], 4 * 24 * 3600)) * on(namespace) group_left(label_monitoring_key) kube_namespace_labels{label_monitoring_key="middleware"}) <= 0
         for: 15m
         labels:
           severity: warning
@@ -216,7 +216,7 @@ spec:
         annotations:
           message: The {{ '{{' }} $labels.persistentvolumeclaim {{ '}}' }} PVC will run of disk space in the next 4 hours.
         expr: |
-          (sum by(persistentvolumeclaim, namespace) (kubelet_volume_stats_available_bytes{job="kubelet"} and predict_linear(kubelet_volume_stats_available_bytes{job="kubelet"}[1h], 4 * 3600)) * on ( namespace) group_left(label_monitoring_key) kube_namespace_labels{label_monitoring_key="middleware"}) <= 0
+          (sum by(persistentvolumeclaim, namespace) (predict_linear(kubelet_volume_stats_available_bytes{job="kubelet"}[1h], 4 * 3600)) * on(namespace) group_left(label_monitoring_key) kube_namespace_labels{label_monitoring_key="middleware"}) <= 0
         for: 15m
         labels:
           severity: critical
--- a/roles/middleware_monitoring_config/templates/kube_state_metrics_alerts.yml.j2
+++ b/roles/middleware_monitoring_config/templates/kube_state_metrics_alerts.yml.j2
@@ -204,3 +204,19 @@ spec:
         for: 15m
         labels:
           severity: warning
+      - alert: PVCStorageWillFillIn4Days
+        annotations:
+          message: The {{ '{{' }} $labels.persistentvolumeclaim {{ '}}' }} PVC will run of disk space in the next 4 days.
+        expr: |
+          (sum by(persistentvolumeclaim, namespace) (predict_linear(kubelet_volume_stats_available_bytes[6h], 4 * 24 * 3600)) * on ( namespace) group_left(label_monitoring_key) kube_namespace_labels{label_monitoring_key="middleware"}) < 0
+        for: 15m
+        labels:
+          severity: warning
+      - alert: PVCStorageWillFillIn4Hours
+        annotations:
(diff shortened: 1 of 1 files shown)


roles/middleware_monitoring_config/templates/kube_state_metrics_alerts.yml.j2
```diff
@@ -204,3 +204,27 @@ spec:
...
+        expr: |
+          (sum by(persistentvolumeclaim, namespace) (kubelet_volume_stats_available_bytes{job="kubelet"} and predict_linear(kubelet_volume_stats_available_bytes{job="kubelet"}[1h], 4 * 3600)) * on ( namespace) group_left(label_monitoring_key) kube_namespace_labels{label_monitoring_key="middleware"}) <= 0
+        for: 15m
+        labels:
+          severity: critical
+      - alert: PersistentVolumeErrors
+        annotations:
+           message: The PVC {{ '{{' }} $labels.persistentvolumeclaim {{ '}}' }} is in status {{ '{{' }} $labels.phase {{ '}}' }} in namespace {{ '{{' }} $labels.namespace {{ '}}' }}
+        expr: |
+          (sum by(persistentvolumeclaim, namespace, phase) (kube_persistentvolumeclaim_status_phase{phase=~"Failed|Pending"}) * on ( namespace) group_left(label_monitoring_key) kube_namespace_labels{label_monitoring_key="middleware"}) > 0
```
david-martin: @alanmoran  Can you suggest a way to test this alert for `PersistentVolumeErrors`?
Is there a way to get a PV into one of the non-good states

R-Lawton:
The alert for 4 days triggered 
4day

R-Lawton:
prometheus
@david-martin

david-martin:
The query as is doesn't seem quite right
```
          (sum by(persistentvolumeclaim, namespace) (kubelet_volume_stats_available_bytes{job="kubelet"} and predict_linear(kubelet_volume_stats_available_bytes{job="kubelet"}[1h], 4 * 3600)) * on ( namespace) group_left(label_monitoring_key) kube_namespace_labels{label_monitoring_key="middleware"}) <= 0
```

The result of `kubelet_volume_stats_available_bytes{job="kubelet"} and predict_linear(kubelet_volume_stats_available_bytes{job="kubelet"}[1h], 4 * 3600)` isn't returning the predicted value. It is returned the actual usage value.

I think using just `predict_linear(kubelet_volume_stats_available_bytes[1h], 4 * 3600)` would be sufficient.
And `predict_linear(kubelet_volume_stats_available_bytes[6h], 4 * 24 * 3600)` for the 4day version.
This would mean the value can go below 0, which is fine.

There is also another weird behaviour with this query. If for some reason a large amount of data is removed/cleaned up in the pv, the storage predication can jump to a value greater than whats available in the PV for some time (until the range window of [1h] or [6h]) has passed.
I don't think this would affect the alert, but it would be nice for graphing purposes to max the value at the claim amount.

e.g. `min(kubelet_volume_stats_capacity_bytes, predict_linear(kubelet_volume_stats_available_bytes[1h], 4 * 3600))`

So the full query I'd recommed is:

for 4 hours predication
```
(sum by(persistentvolumeclaim, namespace) (min(kubelet_volume_stats_capacity_bytes, predict_linear(kubelet_volume_stats_available_bytes[1h], 4 * 3600))) * on ( namespace) group_left(label_monitoring_key) kube_namespace_labels{label_monitoring_key="middleware"}) <= 0
```
and for 4 days prediction
```
(sum by(persistentvolumeclaim, namespace) (min(kubelet_volume_stats_capacity_bytes, predict_linear(kubelet_volume_stats_available_bytes[6h], 4 * 24 * 3600))) * on ( namespace) group_left(label_monitoring_key) kube_namespace_labels{label_monitoring_key="middleware"}) <= 0
```

@R-Lawton  Can you try these versions of the queries?

R-Lawton:
yeah sure @david-martin

roles/middleware_monitoring_config/templates/kube_state_metrics_alerts.yml.j2
alanmoran: @david-martin Maybe set a quota of 2 persistence volumes in an ns and then create 3. that would put one into a state of pending.

david-martin:
4Day alert triggering with 200K/minute (12M/hour, 288M/day, 1.152/4day)
image

3 graphs showing stats over time:
- actual volume usage (~945/1G)
- prediction for 4hrs from now (~898MB/1G) (4hrs = ~48M at current rate of writing)
- prediction for 4days from now (0/1G) (all data used up)

image

david-martin:
4hour (& 4Day) triggering with 6M/min (360M/hour, 1440M/4hrs)

image

3 graphs showing stats over time:

actual volume usage (~787/1G)
prediction for 4hrs from now (0/1G) (4hrs = ~1440M at current rate of writing)
prediction for 4days from now (0/1G) (4days = ~34.5G at current rate of writing)

image

roles/middleware_monitoring_config/templates/kube_state_metrics_alerts.yml.j2
david-martin: Hmm, doing that results in an error trying to create the additional claim

```oc create -f /tmp/storage.yaml
Error from server (Forbidden): error when creating "/tmp/storage.yaml": persistentvolumeclaims "test3" is forbidden: exceeded quota: storagequota, requested: persistentvolumeclaims=1,requests.storage=10Gi, used: persistentvolumeclaims=2,requests.storage=1000Mi, limited: persistentvolumeclaims=1,requests.storage=5Gi```

I'm wondering if there's an alternative way

**Comment by david-martin** [comment]
Verified PersistentVolumeErrors alert by creating a couple pvcs with gp2 storageclass & RWX. (gp2/ebs can only be RWO)

image
Also, the `Pending` status metric for that claim has a value of 1
image
The alert will go red and start triggering after 15m.
The event that gets generated in this case is:
```
1m          7m           26      test3.15d6b2f2a626cb8e                         PersistentVolumeClaim                                 Warning   ProvisioningFailed            persistentvolume-controller             Failed to provision volume with StorageClass "gp2": invalid AccessModes [ReadWriteMany]: only AccessModes [ReadWriteOnce] are supported
```

I've also added `Lost` to the statuses to look for.
To verify that, I removed the claimRef.UID from a bound PV.
This resulted in the `Lost` status metric having a value of 1.
image

The event that gets generated in this case looks like:
```
10s         10s          1       test2.15d6b356c1adc108                         PersistentVolumeClaim                                 Warning   ClaimLost                     persistentvolume-controller             Bound claim has lost its PersistentVolume. Data on the volume is lost!
```

PR #1142, opened 2019-11-12
Adding CVE rollout playbook

pmccarthy:
## Additional Information
Adding CVE rollouts playbook which calls on specific product role `upgrade_images` tasks

## Verification Steps

Run playbook with no tags specified. Expected behaviour is that upgrade_images will be run for all products.
```
ansible-playbook -i inventories/osd.template playbooks/cve_rollout.yml
```

Now re-run the above with specific product tags to only run the image upgrade task for these products.
```
ansible-playbook -i inventories/osd.template playbooks/cve_rollout.yml --tags 3scale

ansible-playbook -i inventories/osd.template playbooks/cve_rollout.yml --tags 3scale,rhsso
```

### DIFF
--- a/playbooks/cve_rollout.yml
+++ b/playbooks/cve_rollout.yml
@@ -1,0 +1,21 @@
+# Generic playbook for CVE rollouts
+---      
+- hosts: localhost
+  gather_facts: yes
+  # Ensure task is always run regardless of whether tags are specified
+  tags: ['always']
+  tasks:
+  # Required for Ansible Tower installs that need to login via oc as a prerequisite
+  - name: Openshift Login
+    include_role:
+      name: openshift
+      tasks_from: login.yml
+
+  - name: Update product images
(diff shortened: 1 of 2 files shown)


playbooks/openshift.yml
```diff
@@ -6,10 +6,12 @@
         name: openshift
         tasks_from: set_master_vars.yml
       when: run_master_tasks | default(true) | bool
+      tags: ['always']
```
pmccarthy: Unfortunately we have no choice but to apply this always tag, see ansible docs [1]. This ensures that these tasks always get run regardless of what tags get passed in.

For example, if the only tag specified in a run is `rhsso` then ansible will only execute tasks that have the `rhsso` tag applied or the `always` tag

[1] https://docs.ansible.com/ansible/latest/user_guide/playbooks_tags.html#special-tags

PR #1152, opened 2019-11-14
add current sheriff eng to owners

aidenkeating:
add current support sheriffs to owners to allow for upgrade work to progress

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -9,6 +9,8 @@ approvers:
 - pb82
 - jessesarn
 - davidffrench
+- aidenkeating
+- ciaranRoche
 
 reviewers:
 - pmccarthy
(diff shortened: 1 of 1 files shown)


aidenkeating:
@david-martin mind taking a look?

david-martin:
@aidenkeating I'm going to put this on hold, pending a discussion with @maleck13 .

If we added sheriff as owners, that would mean everyone gets added as rotation happens. The slippery slope argument here why not just add everyone now (and that goes against the idea of a small set of owners :) )

maleck13:
agree @david-martin the idea was to keep a small set of owners to try and help with quality on reviews. If stuff is not getting reviewed merged we need to address that though

aidenkeating:
@maleck13 @david-martin That makes sense to me, the pattern we've appeared to be following so far is:

- Someone from the sheriff team produces a PR
- Someone else from the sheriff team does a verification and review
- Someone from the owners file is asked for a final review

If we're happy with that, and it's not expected that the owners will need to perform the verification again, I'm happy to close this

PR #1154, opened 2019-11-15
INTLY-4012 Add Fuse managed upgrade task

tremes:
## Verification Steps

### Installation Verification
### Upgrade Verification

- [ ] Yes
- [ ] No

- [ ] Tested Installation
- [ ] Tested Uninstallation


roles/fuse_managed/defaults/main.yml
```diff
@@ -6,3 +6,50 @@ fuse_backup_postgres_secret_name: fuse-postgres-auth
 fuse_pull_secret_name: "syndesis-pull-secret"
 
 fuse_resources: []
+
+fuse_upgrade_images:
+  - name: "fuse-ignite-server"
+    image: "registry.redhat.io/fuse7/fuse-ignite-server:1.4-17"
+    tag: "1.7"
+    namespace: "openshift"
+    imageStreamName: "fuse-ignite-server"
```
mikenairn: What's the difference between name and imageStreamName, do we need both?

roles/fuse_managed/defaults/main.yml
```diff
@@ -6,3 +6,50 @@ fuse_backup_postgres_secret_name: fuse-postgres-auth
...
+    imageStreamName: "prometheus"
+  - name: "jboss-amq-6"
+    image: "registry.access.redhat.com/jboss-amq-6/amq63-openshift:1.3"
+    tag: "1.3"
+    namespace: "{{ fuse_namespace }}"
+    imageStreamName: "jboss-amq-63"  
+  - name: "fuse-online-operator"
+    image: "registry.redhat.io/fuse7/fuse-online-operator:1.4-16"
+    tag: "1.7"
+    namespace: "{{ fuse_namespace }}"
```
mikenairn: I had wondered about the best way to do this. I think this is fine, but we could also just default to the current namespace if this wasn't set in the object.

roles/fuse_managed/tasks/upgrade_images.yml
```diff
@@ -1,4 +1,16 @@
 ---
-#ToDo Implement CVE image update steps as described https://github.com/RHCloudServices/integreatly-help/blob/master/sops/cves/applying-cve-updates.md
-- debug:
-    msg: "TODO Implement me!!"
+- name: Add new tag to Fuse online images
+  shell: "oc tag --source docker {{ obj_item.image }} {{ obj_item.imageStreamName }}:{{ obj_item.tag }} -n {{ obj_item.namespace }}"
+  with_items: "{{ fuse_upgrade_images }}"
+  loop_control:
+    loop_var: obj_item
```
mikenairn: Might be good to keep these namings
 consistent with others https://github.com/integr8ly/installation/blob/master/roles/3scale/tasks/upgrade_images.yml#L7

roles/fuse_managed/defaults/main.yml
tremes: See e.g https://github.com/integr8ly/installation/pull/1154/files#diff-2181084ac6785df080ed6e18eafaad1cR40 and https://github.com/integr8ly/installation/pull/1154/files#diff-2181084ac6785df080ed6e18eafaad1cR35. For these two there's a difference

roles/fuse_managed/tasks/upgrade_images.yml
tremes: Yeah but I wanted to emphasize somehow that it's object....but I am ok with renaming

roles/fuse_managed/defaults/main.yml
```diff
@@ -6,3 +6,50 @@ fuse_backup_postgres_secret_name: fuse-postgres-auth
 fuse_pull_secret_name: "syndesis-pull-secret"
 
 fuse_resources: []
+
+fuse_upgrade_images:
```
mikenairn: We called the equivalent for 3scale "threescale_image_streams" https://github.com/integr8ly/installation/blob/master/roles/3scale/defaults/main.yml#L159, would probably makes sense to keep this naming consistent. Using just the "name" attribute also makes more sense then, since it's an array of image streams and the variable name suggests that.

roles/fuse_managed/tasks/upgrade_images.yml
```diff
@@ -1,4 +1,16 @@
...
+
+- name: Import new Fuse online images
+  shell: "oc import-image {{ obj_item.imageStreamName }}:{{ obj_item.tag }} --confirm='true' {{ obj_item.image }} -n {{ obj_item.namespace }}"
+  with_items: "{{ fuse_upgrade_images }}"
+  loop_control:
+    loop_var: obj_item 
+  register: import_image_result
+  until: import_image_result.rc == 0
+  retries: 5
+  delay: 5
```
mikenairn: Any reason we need to retry this? Is it likely to fail?

roles/fuse_managed/defaults/main.yml
tremes: But it's true that I don't use the `name` at all so I will update it.

roles/fuse_managed/tasks/upgrade_images.yml
tremes: Not sure really (copy/paste). I'll try to remove

mikenairn:
/lgtm

Can give it the final approval and merge when you are happy with the testing done on it.

mikenairn:
/lgtm

Can give it the final approval and merge when you are happy with the testing done on it.

mikenairn:
@tremes This PR will also need to add "fuse_managed" into this list https://github.com/integr8ly/installation/blob/master/playbooks/group_vars/all/cve.yml#L2

@pmccarthy That's the only change required when adding a new product to the CVE roll out list isn't it?

PR #1155, opened 2019-11-15
INTLY-4160 - extract postgresql image to a variable in Enmasse

tremes:
## Verification Steps

### Installation Verification
### Upgrade Verification

- [ ] Yes
- [ ] No

- [ ] Tested Installation
- [ ] Tested Uninstallation

### DIFF
--- a/roles/enmasse/defaults/main.yml
+++ b/roles/enmasse/defaults/main.yml
@@ -11,6 +11,7 @@ enmasse_clean_artifacts: true
 enmasse_backup_postgres_secret: 'enmasse-postgres-secret'
 enmasse_postgres_cronjob_name: 'enmasse-postgres-backup'
 enmasse_pv_cronjob_name: 'enmasse-pv-backup'
+enmasse_postgresql_image: 'postgresql:9.6'
 
 enmasse_resources:
   - name: api-server
--- a/roles/enmasse/tasks/upgrade_images.yml
+++ b/roles/enmasse/tasks/upgrade_images.yml
@@ -14,7 +14,7 @@
   with_items: "{{ deployments }}"
 
 - name: "Postgresql update the imagestreams"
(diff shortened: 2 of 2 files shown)


roles/enmasse/defaults/main.yml
```diff
@@ -11,6 +11,7 @@ enmasse_clean_artifacts: true
 enmasse_backup_postgres_secret: 'enmasse-postgres-secret'
 enmasse_postgres_cronjob_name: 'enmasse-postgres-backup'
 enmasse_pv_cronjob_name: 'enmasse-pv-backup'
+postgresql_image: 'postgresql:9.6'
```
mikenairn: Should we prepend enmasse to the variable name here? "enmasse_postgresql_image". If we start using generic names like this for different products, could cause all sorts of weirdness. @pmccarthy Any opinion?

tremes: yeah I was wondering exactly about the same. My reasoning was it's already in enmasse role...but on the other hand all other variable do have this prefix too...

mikenairn: Yes it's in the enmasse role, but as soon as someone sets it elsewhere for whatever reason, if more than 1 product use the same variable name it will affect both. We more than likely don't actually want that to happen.

roles/enmasse/tasks/upgrade_images.yml
```diff
@@ -14,7 +14,7 @@
   with_items: "{{ deployments }}"
 
 - name: "Postgresql update the imagestreams"
-  shell: oc import-image postgresql:9.6 -n openshift
+  shell: oc import-image {{ postgresql_image }} -n openshift
```
mikenairn: Needs to be the same here "enmasse_postgresql_image"

tremes: ahh sorry too quick commit

PR #1156, opened 2019-11-15
Adding keycloak operator patch task for CVE rollouts

pmccarthy:
## Additional Information

- Adding keycloak operator patch task for CVE rollouts
- Renaming `upgrade_sso_operator_image` variable to `rhsso_operator_image` to match existing naming convention

## Verification Steps

- [ ] Run `cve_rollout.yml` playbook against an RHMI cluster using this feature branch - nothing should change
- [ ] Force a change to the keycloak operator by setting an extra var `rhsso_operator_release_tag=v1.9.3`


mikenairn:
/lgtm

We happy enough with the level of testing done on this? If you are you can just approve this and have it merge.

pmccarthy:
@mikenairn Thanks. Yeah I tested each option on multiple occasions on my PDS tower environment so should be fine

PR #1157, opened 2019-11-15
🔀 Revert "Merge pull request #1077 from integr8ly/ups-operator-ga"

grdryn:
This reverts commit 5850c4b439530070bdc308986b6bb769f30aef40, reversing
changes made to a97525b6d2efa5b01e9248a82929198ffbbc6f7f.

Fixes: https://issues.jboss.org/browse/AEROGEAR-10070

## Additional Information

We're not upgrading UPS for RHMI 1.6.

## Verification Steps
I don't think any are needed, since it's just bringing it back to the previously tested state, right?

~As the verifier of the PR the following process should be done:~

~### Installation Verification~
~- Ensure the author of the PR has attached a log of the installation run from his branch to the jira or pr and check that it exited as expected.~
~- Verify the fresh installation is correct on cluster provided by PR author~
~### Upgrade Verification~
~- After installation verification, notify the PR author to begin an upgrade on their cluster~
~- Ensure the developer of the PR has attached a log of the upgrade run from his branch to the jira or pr and check that it exited as expected.~
~- If possible, look at the tasks that ran and see they match the PR~
~- Verify the upgrade is correct on cluster provided by PR author~

- [ ] Yes
- [x] No

- [ ] Tested Installation
- [ ] Tested Uninstallation


mikenairn:
If the decision is that we are now **not** actually updating UPS for 1.6, then i guess this is fine.

/shrug

PR #1159, opened 2019-11-18
INTLY-4159 update codeready upgrade_images task

tremes:
## Verification Steps

### Installation Verification
### Upgrade Verification

- [ ] Yes
- [ ] No

- [ ] Tested Installation
- [ ] Tested Uninstallation

### DIFF
--- a/roles/codeready/tasks/upgrade_images.yml
+++ b/roles/codeready/tasks/upgrade_images.yml
@@ -1,22 +1,30 @@
 ---
-- name: "patch codeready deployment"
-  shell: 'oc patch deployment codeready --patch=''{"spec":{"template":{"spec":{"containers":[{"name":"codeready","imagePullPolicy":"Always" }]}}}}'' -n {{ che_namespace }}'
+- name: "patch {{ che_deployment_name }} deployment"
+  shell: 'oc patch deployment {{ che_deployment_name }} --patch=''{"spec":{"template":{"spec":{"containers":[{"name":"{{ che_deployment_name }}","imagePullPolicy":"Always" }]}}}}'' -n {{ che_namespace }}'
   register: patch_codeready_result
   failed_when: patch_codeready_result.stderr != ""
 
-- name: "patch postgres deployment"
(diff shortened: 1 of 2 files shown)


roles/codeready/tasks/upgrade_images.yml
```diff
@@ -1,22 +1,33 @@
...
+- name: Patch {{ che_operator_deployment_name }} Deployment
+  shell: oc patch deployment {{ che_operator_deployment_name }} --patch='{"spec":{"template":{"spec":{"containers":[{"name":"{{ che_operator_deployment_name }}","image":"{{ che_operator_image_name }}"}]}}}}' --namespace {{ che_namespace }}
+  register: codeready_operator_image_patch_result
+  failed_when: codeready_operator_image_patch_result.stderr != ""
+
+- name: "redeploy {{ che_deployment_name }}"
+  shell: 'oc patch deployment {{ che_deployment_name }} --patch ''{"spec":{"template":{"metadata":{"annotations":{"last-restart":"`date +''%s''`"}}}}}'' -n {{ che_namespace }}'
   register: redeploy_codeready_result
   failed_when: redeploy_codeready_result.stderr != ""
   when: '"not patched" in patch_codeready_result.stdout'
```
mikenairn: @tremes @pmccarthy @maleck13 Based on what we said yesterday, i think for this to support both upgrades and CVE respins, these lines need to be removed fro the redeploy tasks. 

This will mean that all 3 things will get redeployed any time a CVE roll out or upgrade is carried out for codeready, i think this is fine, but what does anyone else think.

maleck13: I am ok with this as otherwise I imagine we will need to get very specific with our arguments etc

pmccarthy: @mikenairn I agree that this should be fine. As you said this task will only get run if a codeready change has been picked up by delorean

tremes: Removed. Thx

PR #1163, opened 2019-11-21
[WIP] Removed MDC from installation and configured uninstall on upgrade

ziccardi:
Upgrade: uninstall MDC (leaves the namespace and the resources)
Install: doesn't install MDC anymore


grdryn:
I've taken this, rebased it, and made some changes. I wasn't able to update this PR though, so I've opened #1171 instead.

grdryn:
Covered by #1171 which is merged now, so this can be closed.

PR #1164, opened 2019-11-21
Fuse 1.8 Install

ciaranRoche:
## Additional Information

This change, installs the latest version of fuse. 

Work carried out on the Upgrade will be solved in a follow up PR

JIRA - https://issues.jboss.org/browse/INTLY-4208


integreatly.fuse-online-rc.yml
```diff
@@ -0,0 +1,3 @@
+imageSync:
+  sourceRepo: brew-pulp-docker01.web.prod.ext.phx2.redhat.com:8888
```
aidenkeating: I don't think we need this file, right?

inventories/group_vars/all/manifest.yaml
```diff
@@ -31,15 +31,15 @@ enmasse_git_url: https://github.com/jboss-container-images/amq-online-images.git
...
 #not currently used as the operator decides what to install
 fuse_version: '7.4.1'
 
 #controls whether Fuse Online is installed or not
 fuse_online: True
 # Below Fuse vars are not currently used but will be used to source the resources needed to do the install.
-fuse_online_release_tag: '1.7.27'
-fuse_online_resources_base: 'https://raw.githubusercontent.com/syndesisio/fuse-online-install/{{fuse_online_release_tag}}/resources'
+fuse_online_release_tag: '1.8.9'
+fuse_online_resources_base: 'https://raw.githubusercontent.com/integr8ly/fuse-online-install/1.8.9/resources'
```
aidenkeating: We can probably continue to use `{{fuse_online_release_tag}}` here instead of an explicit tag.

Also, `1.8.13` is now available, we can probably swap to that.

roles/fuse/tasks/upgrade.yml
```diff
@@ -78,8 +69,6 @@
 - name: Create Fuse Console templates
   shell: oc replace --force -f {{ fuse_resource_base }}/{{ item }} -n openshift
   with_items:
-    - "fis-console-cluster-template.json"
-    - "fis-console-namespace-template.json"
```
aidenkeating: Why do we no longer install these? Are they set up via the CLI?

roles/fuse_managed/defaults/main.yml
```diff
@@ -5,6 +5,8 @@ fuse_backup_postgres_secret_name: fuse-postgres-auth
 
 fuse_pull_secret_name: "syndesis-pull-secret"
```
aidenkeating: This line will get overridden, can be remove it?

roles/fuse_managed/tasks/main.yml
```diff
@@ -35,20 +35,32 @@
...
+    dest: /tmp/fuse-binary
+
+- name: Run operator cluster install
+  shell: "/tmp/fuse-binary/syndesis-operator install cluster --namespace {{ fuse_namespace }}"
+
+- name: Run operator install
+  shell: "/tmp/fuse-binary/syndesis-operator install operator --namespace {{ fuse_namespace }}"
+
+#- name: TODO-REMOVE-THIS-AFTER-FUSE-ONLINE-INCLUDES-THESE-ARGS Patch args on operator to include monitoring resources addon
+#  shell: oc patch dc/syndesis-operator -p='{"spec":{"template":{"spec":{"containers":[{"name":"syndesis-operator","args":["-a","/conf/addons"]}]}}}}}' -n {{ fuse_namespace }}
```
aidenkeating: Should this block be removed?

roles/fuse_managed/tasks/upgrade.yml
```diff
@@ -1,104 +1,22 @@
...
-  shell: oc get serviceaccounts -n {{ fuse_namespace }} | grep syndesis | awk '{print $1}'
-  register: fuse_serviceaccounts
-
-- name: Link syndesis-pull-secret to fuse service accounts for image pull
-  shell: "oc secrets link {{ item }} {{ fuse_pull_secret_name }} --for=pull -n {{ fuse_namespace }}"
-  with_items: "{{ fuse_serviceaccounts.stdout_lines }}"
-
-- name: Link syndesis-pull-secret to builder service account
-  shell: "oc secrets link builder {{ fuse_pull_secret_name }} --for=pull,mount -n {{ fuse_namespace }}"
+- name: Download fuse binary
```
aidenkeating: Unless this has been confirmed to work, I think it probably makes more sense to have this print out a log stating that update will be completed in INTLY-4209

inventories/group_vars/all/manifest.yaml
```diff
@@ -31,15 +31,15 @@ enmasse_git_url: https://github.com/jboss-container-images/amq-online-images.git
 #controls whether fuse is installed or not
 fuse: True
 #This is the release tag for fuse on openshift in github currently used to pull in the templates and image streams
-fuse_release_tag: 'application-templates-2.1.fuse-740025-redhat-00004'
+fuse_release_tag: 'application-templates-2.1.fuse-750031-redhat-00001'
```
aidenkeating: Is there a reason we use this specific tag, rather than the later 75 tag of https://github.com/jboss-fuse/application-templates/tree/application-templates-2.1.fuse-750056-redhat-00006 ?

integreatly.fuse-online-rc.yml
aidenkeating: @ciaranRoche This file is causing the failing build

roles/fuse/tasks/upgrade.yml
ciaranRoche: These are in the upgrade task, for this work I will leave as is. We can address is in the follow up PR

roles/fuse/tasks/main.yml
```diff
@@ -1,47 +1,7 @@
...
-  shell: "oc replace --force -f {{ item }} -n openshift"
-  register: create_fuse_imagestream_cmd
-  failed_when: create_fuse_imagestream_cmd.stderr != '' and 'AlreadyExists' not in create_fuse_imagestream_cmd.stderr
-  changed_when: create_fuse_imagestream_cmd.rc == 0
-  with_items:
-    - "{{ fuse_online_imagestream_resources }}"
-    - "{{ fuse_on_openshift_imagestreams_url }}"
-
-# SETUP FUSE ON OPENSHIFT TEMPLATES
-- name: Create Fuse quickstart templates
```
aidenkeating: Why are we no longer creating these? The 7.5 templates are missing after an install

ciaranRoche: That was an oversight, once I run the install, ill ping you on the cluster to verify it.

aidenkeating:
@ciaranRoche Tested on your cluster, I'm not an owner, but this is working for installation.

Upgrade will be handled in a separate PR.

roles/fuse_managed/tasks/upgrade.yml
```diff
@@ -1,104 +1,23 @@
...
-    - syndesis-jvm-dashboard.yml
-    - syndesis-legacy-ui-serviceaccount.yml
-    - syndesis-legacy-ui.yml
-    - syndesis-meta-servicemonitor.yml
-    - syndesis-rest-api-dashboard.yml
-    - syndesis-rest-api-prometheus-rule.yml
-    - syndesis-server-servicemonitor.yml
\ No newline at end of file
+- debug: msg="Upgrade work to be carried out in INTLY-4092, the process has changed in latest fuse release"
+# - name: Download fuse binary
```
david-martin: I'd prefer to not leave commented code lying around, but I'm OK with this as it's a stub for upgrade changes to follow.

PR #1165, opened 2019-11-22
ups update

fh-automatron:
Updated manually by @grdryn 

- This update to the unifiedpush-operator removes a bunch of controllers that we don't use any longer, so this PR removes those CRDs (removing any CRs on the cluster first)
- It also changes the ClusterRole/ClusterRoleBinding to be a Role/RoleBinding.
- The operator now creates the monitoring resources also, so those are removed from the installation here

## Verification steps
- [x] Do a fresh install & make sure that it succeeds & there are no persistent errors in the UPS operator logs

**Log:** https://gist.github.com/grdryn/e9e22a187cf0dc8c24be6eaa77365500#file-install-log
**RHPDS cluster GUID:** `install-53e4`

- [x] Do an upgrade from 1.5.2 and make sure that it succeeds & there are no persistent errors in the UPS operator logs

**Log:** https://gist.github.com/grdryn/e9e22a187cf0dc8c24be6eaa77365500#file-upgrade-log
**RHPDS cluster GUID:** `upgrade-402e`


grdryn:
no i said _assign_

inventories/group_vars/all/manifest.yaml
```diff
@@ -98,7 +98,7 @@ ups: True
 ups_operator_release_tag: '0.2.0'
 ups_operator_resources: 'https://raw.githubusercontent.com/aerogear/unifiedpush-operator/{{ups_operator_release_tag}}/deploy'
 ups_operator_image: 'quay.io/aerogear/unifiedpush-operator:{{ ups_operator_release_tag }}'
-ups_release_tag: '2.3'
+ups_release_tag: '0.4.0'
```
grdryn: @mikenairn since I changed the delorean job to just be triggered on new versions of the _operator_, is there a way for that job to set `ups_operator_release_tag` rather than `ups_release_tag`, or should I just rename the variables here?

playbooks/group_vars/all/upgrade.yml
```diff
@@ -1,3 +1,3 @@
-# DO NOT EDIT This file is auto generated!!
 upgrade_from_version: release-1.5.1
```
grdryn: @mikenairn Is this file used, and should it say `release-1.5.2` here? Also, should I have added anything to `upgrade_product_roles` when I was removing mdc & mss as part of upgrade?

mikenairn: Don't worry about any of these pre 1.6. You can ignore anything here for the time being.

inventories/group_vars/all/manifest.yaml
mikenairn: Yes, this is wrong. The job should be updating ups_operator_release_tag and not ups_release_tag. What is ups_release_tag even needed for?

mikenairn: https://github.com/integr8ly/ci-cd/pull/367

PR #1161, opened 2019-11-19
INTLY 3781 Add heimdall

mikenairn:
## Additional Information

Adds the heimdall operator to the installation and enabled the image monitoring on all the products that current support CVE respins https://github.com/integr8ly/installation/blob/master/playbooks/group_vars/all/cve.yml#L2

## Jira

* https://issues.jboss.org/browse/INTLY-3781
* https://issues.jboss.org/browse/INTLY-3672

## Verification Steps

**Heimdall Disabled**

Heimdall is disabled by default, running the install from this branch without additional parameters shouldn't install the heimdall operator or adding any image monitoring alerts to monitoring stack.

**Heimdall Enabled**

Create or use a service account token from https://access.redhat.com/terms-based-registry

You need to get the username and token value from the openshift secret configuration i.e. https://access.redhat.com/terms-based-registry/#/token/-heimdall/openshift-secret
The token is the base64 encoded string value after ".dockerconfigjson"

Run the installation and verify that the heimdall operator is added and running:

```
$ ansible-playbook -i inventories/hosts.template playbooks/install_heimdall.yml -e heimdall_pull_secret_name=6340056-heimdall -e heimdall_pull_secret_token=eyJhd....9In19fQ==
```

```
$ oc get deployment heimdall -n heimdall
NAME       DESIRED   CURRENT   UP-TO-DATE   AVAILABLE   AGE
heimdall   1         2         1            1           1h
```

Verify that the image monitoring resources are created for each of the products listed in https://github.com/integr8ly/installation/blob/master/playbooks/group_vars/all/cve.yml#L2

```
$ oc get imagemonitor --all-namespaces
NAMESPACE   NAME       AGE
3scale      heimdall   1h
apicurito   heimdall   1h
codeready   heimdall   1h
enmasse     heimdall   1h
fuse        heimdall   1h
sso         heimdall   1h
```

### Installation Verification
- Ensure the author of the PR has attached a log of the installation run from his branch to the jira or pr and check that it exited as expected.
- Verify the fresh installation is correct on cluster provided by PR author 
### Upgrade Verification
- After installation verification, notify the PR author to begin an upgrade on their cluster
- Ensure the developer of the PR has attached a log of the upgrade run from his branch to the jira or pr and check that it exited as expected. 
- If possible, look at the tasks that ran and see they match the PR
- Verify the upgrade is correct on cluster provided by PR author 

## Is an upgrade task required and are there additional steps needed to test this?

- [ ] Yes
- [ ] No

[...]

PR #1169, opened 2019-11-25
INTLY-3672 - Grafana Dashboard and monitoring alert for image freshne…

tremes:
…ss labels

## Verification Steps

### Installation Verification
### Upgrade Verification

- [ ] Yes
- [ ] No

- [ ] Tested Installation
- [ ] Tested Uninstallation

### DIFF
--- a/roles/middleware_monitoring_config/templates/kube_state_metrics_alerts.yml.j2
+++ b/roles/middleware_monitoring_config/templates/kube_state_metrics_alerts.yml.j2
@@ -232,3 +232,11 @@ spec:
         for: 15m
         labels:
           severity: critical
+      - alert: ResolvableCriticalCVEs
+        annotations:
+           message: There is a following deployment which has some resolvable critical CVEs {{ '{{' }} $labels.label_deployment {{ '}}' }} in {{ '{{' }} $labels.namespace {{ '}}' }} namespace with {{ '{{' }} $labels.label_heimdall_resolvableCriticalCVEs {{ '}}' }} resolvable critical CVEs
+        expr: |
+          sum by (namespace,label_deployment,label_heimdall_resolvableCriticalCVEs) (kube_pod_labels{label_heimdall_resolvableCriticalCVEs!="0",label_heimdall_resolvableCriticalCVEs!="",label_heimdall_updatedImageAvailable="true"}) * on(pod) group_left() (count(kube_pod_labels{label_heimdall_resolvableCriticalCVEs!="0",label_heimdall_resolvableCriticalCVEs!="",label_heimdall_updatedImageAvailable="true"}) > 0)
+        for: 1m
+        labels:
+          severity: critical
--- a/roles/middleware_monitoring_config/defaults/main.yml
+++ b/roles/middleware_monitoring_config/defaults/main.yml
@@ -21,6 +21,7 @@ monitoring_grafanadashboards_resource_templates:
 - resources-by-namespace.yml
 - resources-by-pod.yml
 - cluster-resources.yml
+- images_cves.yml
 
 # PrometheusRules template variables
 monitoring_prometheusrules_resource_templates:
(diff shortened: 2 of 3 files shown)


roles/middleware_monitoring_config/templates/kube_state_metrics_alerts.yml.j2
```diff
@@ -232,3 +232,11 @@ spec:
         for: 15m
         labels:
           severity: critical
+      - alert: ResolvableCriticalCVEs
+        annotations:
+           message: There is a following deployment which has some resolvable critical CVEs {{ '{{' }} $labels.label_deployment {{ '}}' }} in {{ '{{' }} $labels.namespace {{ '}}' }} namespace with {{ '{{' }} $labels.label_heimdall_resolvableCriticalCVEs {{ '}}' }} resolvable critical CVEs
+        expr: |
+          sum by (namespace,label_deployment,label_heimdall_resolvableCriticalCVEs) (kube_pod_labels{label_heimdall_resolvableCriticalCVEs!="0",label_heimdall_resolvableCriticalCVEs!="",label_heimdall_updatedImageAvailable="true"}) * on(pod) group_left() (count(kube_pod_labels{label_heimdall_resolvableCriticalCVEs!="0",label_heimdall_resolvableCriticalCVEs!="",label_heimdall_updatedImageAvailable="true"}) > 0)
```
tremes: @maleck13 Does the condition `label_heimdall_updatedImageAvailable="true"` make sense to you here pls? I guess we don't want to fire this alert when there's no updated image available or am I wrong?

tremes:
I guess I will need to make this work also for upgrades right?

roles/middleware_monitoring_config/templates/kube_state_metrics_alerts.yml.j2
```diff
@@ -232,3 +232,11 @@ spec:
...
+        for: 15m
```
tremes: Not sure about this number....

tremes:
Grafana dashboard can be seen at https://grafana-route-middleware-monitoring.apps.tremes-3dac.open.redhat.com/d/rfF0oBbWz/images-cves?orgId=1

roles/middleware_monitoring_config/templates/kube_state_metrics_alerts.yml.j2
david-martin: @tremes I'm can't think of any obvious scenario where the alert might trigger for a while and then stop triggering (unlike most other alerts). If it triggers, it wont fix itself or revert back to no CVEs.
Alerting immediately sounds like the thing to do here.
That would make it quick to turn green after the image is patched as well.
So `1m` sounds right here.

Related question. Does heimdall remove or update labels if it's discovered new information about a pod on a subsequent run?

tremes: Yes makes sense. Thanks. I'll update it. Wrt removing and updating labels - I think it would make sense but haven't tried it yet. @maleck13  should know :)

roles/middleware_monitoring_config/templates/kube_state_metrics_alerts.yml.j2
maleck13: This is correct. No value firing the alert when no new image is available

roles/middleware_monitoring_config/templates/kube_state_metrics_alerts.yml.j2
maleck13: Once when the image in the dc or the deployment changes, heimdall will run the check again and update the labels with new info

roles/middleware_monitoring_config/templates/kube_state_metrics_alerts.yml.j2
tremes: OK thanks. 
Does the query looks good to you @david-martin pls? It seemed to work good for me, but I am not so experienced with PromQL so I would like to double check.

mikenairn:
/close

Rolled into https://github.com/integr8ly/installation/pull/1161

PR #1171, opened 2019-11-26
Remove MDC

grdryn:
## Additional Information

This is a continuation of #1163. I didn't have the ability to push to the head repo there, so I've just duplicated.

## Verification Steps

- [ ] Make sure MDC is not present (the upgrade case still leaves the namespace, to ensure no data is lost because of UPS or Keycloak CRs -- this may or may not be addressed for 1.6, but will be separate to this PR at least).
- [ ] Make sure there aren't any MDC dashboards in Grafana, & the endpoints dashboards don't have MDC listed either
- [ ] Make sure MDC is not listed in the `manifest` secret in the webapp namespace
- [ ] Make sure that the `CONSUMER_NAMESPACES` env var in the operator in the user sso project doesn't have the mdc namespace listed
- [ ] Make sure that the `APP_NAMESPACES` env var in the operator in the UPS project doesn't have the mdc namespace listed

### Installation Verification

Install: https://gist.github.com/grdryn/754e1842170698395d45bf22dd6dedfe/raw/dbe20c1a95aa7874affb27ad0fa35c79f00e4e06/install.log
The cluster is an RHPDS cluster with a GUID of `install-072b`.

Separately, a log of uninstall: https://gist.github.com/grdryn/754e1842170698395d45bf22dd6dedfe/raw/dbe20c1a95aa7874affb27ad0fa35c79f00e4e06/uninstall.log

### Upgrade Verification

Upgrade log: https://gist.github.com/grdryn/754e1842170698395d45bf22dd6dedfe/raw/dbe20c1a95aa7874affb27ad0fa35c79f00e4e06/upgrade.log
The upgraded cluster is an RHPDS with a GUID of `upgrade-1d9c`.

- [x] Yes

- [x] Tested Installation
- [x] Tested Uninstallation
- [x] Tested Upgrade


playbooks/upgrades/upgrade.yaml
```diff
@@ -7,6 +7,26 @@
         tasks_from: set_master_vars
       when: run_master_tasks | default(true) | bool
 
+    - name: Remove CORS entry from OpenShift master config
```
pmccarthy: This section should only be run if `run_master_tasks` is set to True. This would fail if run against an OSD cluster in its current state.

See block above for reference which uses:
```
when: run_master_tasks | default(true) | bool
```

roles/customisation/tasks/main.yaml
```diff
@@ -218,7 +217,7 @@
   when: enmasse
 
 - set_fact:
-    enmasse_rest_route: "https://{{enmsase_route_cmd.stdout}}"
+    enmasse_rest_route: "https://{{enmasse_route_cmd.stdout}}"
```
pmccarthy: Good spot! How did this ever work 😨

roles/mdc/defaults/main.yml
```diff
@@ -1,35 +1,9 @@
----
+mdc_operator_release_tag: '0.2.6'
+mdc_operator_image: 'quay.io/aerogear/mobile-developer-console-operator:{{ mdc_operator_release_tag }}'
 mdc_template_dir: /tmp
 mdc_namespace: "{{ eval_mdc_namespace | default('mobile-developer-console') }}"
```
pmccarthy: Are these `mdc` related vars are only being kept to support the upgrade playbook which removes `mdc`? Assuming these can be removed following the next release.

roles/mdc/tasks/uninstall.yml
```diff
@@ -22,6 +25,9 @@
   register: output
   failed_when: output.stderr != '' and 'not found' not in output.stderr
 
+- name: Delete mdc operator
```
pmccarthy: If for example we were to run an install from this feature branch (which wouldn't install mdc), is the uninstall intelligent enough to recognise mdc does not exist and not throw an error running these tasks?

playbooks/upgrades/upgrade.yaml
grdryn: Interesting, I'll add that condition, thanks. How do things work on OSD then, does someone have to manually make changes on master?

roles/customisation/tasks/main.yaml
grdryn: Good question! I have no idea :smile:

roles/mdc/defaults/main.yml
grdryn: Yeah, it should be fine to remove these from master after branching 1.6 :+1:

roles/mdc/tasks/uninstall.yml
grdryn: Yes, the uninstall playbook doesn't run the tasks from this uninstall task list from this role any longer. Also, I've tested uninstall and linked the run log in the PR description if you want to check it out!

playbooks/upgrades/upgrade.yaml
grdryn: Done.

roles/mdc/tasks/uninstall.yml
```diff
@@ -6,6 +6,9 @@
   with_items:
     - mobiledeveloperconsole
     - keycloakrealm
```
wei-lee: this will remove keycloak realm CRs, right. Will that cause the realm to be removed from the user-sso instance?

grdryn: Fixed, thanks!

pmccarthy:
Having review the changes, this looks good. Haven't tested it but looks like Wei has so happy for it to be merged 👍 

/lgtm
/approve

PR #1173, opened 2019-11-27
fix monitoring label on namespaces

aidenkeating:
fix boolean label on installation, labels must have string values

### DIFF
--- a/roles/namespace/defaults/main.yml
+++ b/roles/namespace/defaults/main.yml
@@ -1,4 +1,4 @@
 namespace_file: namespace.yml
 
 rhmi_label: rhmi
-rhmi_label_value: "true"
+rhmi_label_value: '"true"'


aidenkeating:
/assign david-martin 

@david-martin Mind taking a look? I'm happy to add quotes to the template either, this is just what I had in my workaround to unblock 3scale work

david-martin:
@aidenkeating Sure, I can approve. However, I'm wondering why this hasn't been a problem until now?
Also, will some patch cmd be needed on upgrade?

aidenkeating:
@david-martin I think we can close this off now, the change was introduced as part of a previous PR https://github.com/integr8ly/installation/pull/1174/files#diff-954ce3032a8c38363ddee682d59233d2R4

As for why it has only come up now, it was only merged 10 days ago. I'm not sure of the testing environment, but this has been tested on a fresh RHMI Workshop cluster. The upgrade task will also take this change into account.

PR #1174, opened 2019-11-27
Bump 3scale version & fuse version for solution pattern

aidenkeating:
bump 3scale to 2.7.0


aidenkeating:
upgrade will be handled in https://issues.jboss.org/browse/INTLY-4266

aidenkeating:
/assign david-martin

@david-martin Would you mind taking a look, it has been tested on two clusters and fixes an exiting installation bug introduced on master. Upgrade will be handled separately

PR #1175, opened 2019-11-27
Remove MDC namespace on upgrade

grdryn:
## Additional Information

In #1171 we removed MDC, but for existing clusters we left the namespace there to make sure that users didn't lose any config in UPS or RHSSO. The fallback plan for that was to have the namespace removed in 1.7, but this change allows us to remove it now.

## Verification Steps

- [ ] Have a cluster at version 1.5.2
- [ ] Create an app in MDC and bind everything
- [ ] Look at your config in UPS (app w/ relevant variant) and RHSSO (relevant realm)
- [ ] Run the upgrade playbook and make sure it's successful
- [ ] Make sure that your UPS and RHSSO config are still intact
- [ ] Make sure the MDC namespace is gone

~### Installation Verification~
~- Ensure the author of the PR has attached a log of the installation run from his branch to the jira or pr and check that it exited as expected.~
~- Verify the fresh installation is correct on cluster provided by PR author~

This affects upgrade only.

### Upgrade Verification

Successful upgrade log: https://gist.github.com/grdryn/52940316c2b788673d9216c47f8cbdc6
RHPDS cluster GUID: `upgrade-1d9c`

- [x] Yes

### DIFF
--- a/playbooks/upgrades/upgrade.yaml
+++ b/playbooks/upgrades/upgrade.yaml
@@ -50,16 +50,66 @@
     # Mobile Developer Console
     - name: Uninstall Mobile Developer Console
       block:
+        - name: Get User RHSSO deployment desired replicas
+          shell: "oc get deployment/keycloak-operator -o jsonpath='{.spec.replicas}' -n {{ eval_user_rhsso_namespace }}"
+          register: expected_sso_operator_replicas
+        - name: Scale down User RH-SSO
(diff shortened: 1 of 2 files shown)


wei-lee:
code changes looks good. I will get the cluster running and try it out.

wei-lee:
read through the logs and check the state of the cluster and I think the PR is good to merge!

davidffrench:
Code changes look good to me. Would be good to get the CRDs removed also for cleanup.

PR #1177, opened 2019-11-28
use fuse 1.8 3scale integration options instead of 1.7 options

aidenkeating:
the way 3scale is integrated with fuse online has changed in 1.8
and now requires the management url for the 3scale deploy to be
specified in the syndesis custom resource instead of a boolean.

upgrades are not handled in this commit, it will be handled in a
separate fuse 1.8 upgrade commit.

verification:
- run the installer
- ensure it finishes successfully
- ensure the syndesis cr in the shared fuse namespace has the new flag
- ensure deployed integrations from the shared fuse are visible in
  3scale

@ciaranRoche mind taking a look? can we also update the fuse 1.8 upgrade task to include the upgrade version of this change


pmccarthy:
Approving based on validation work completed by @aidenkeating 

/approve

PR #1179, opened 2019-12-02
INTLY-4100 standardise on pds cluster type

aidenkeating:
verification:
- run the installer using the pds inventory file
- ensure nexus is created

### DIFF
--- a/inventories/group_vars/all/manifest.yaml
+++ b/inventories/group_vars/all/manifest.yaml
@@ -1,9 +1,6 @@
 ---
 integreatly_version: master
 
-# Possible values are: rhpds, poc, osd, dev
-cluster_type: rhpds
-
 # threescale version is controlled by the operator. Changing the operator version will change the 3scale version
 
 # controls whether threescale is installed or not.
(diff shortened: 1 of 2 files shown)


inventories/group_vars/all/manifest.yaml
```diff
@@ -1,8 +1,8 @@
 ---
 integreatly_version: master
 
-# Possible values are: rhpds, poc, osd, dev
-cluster_type: rhpds
+# Possible values are: pds, poc, osd, dev
+cluster_type: pds
```
pmccarthy: Given that we set `cluster_type` in each of the environment specific group_vars we can probably just remove this altogether or at the very least, set as an empty string

david-martin:
/lgtm
/approve

@aidenkeating to follow up with the removal of hosts.template and some docs to remove any confusion on how to run the installer after recent inventory changes

aidenkeating:
@david-martin Thanks for taking a look, follow-up tasks captured in https://issues.jboss.org/browse/INTLY-4312

inventories/group_vars/all/manifest.yaml
jessesarn: Seems removing altogether may have had a side effect of breaking the integreatly installer on RHPDS (https://issues.redhat.com/browse/INTLY-4368)

Without a value being set by default, cluster_type is left undefined which none of the conditionals which are based on the value of the cluster_type handle gracefully.  Setting the value to rhpds may not have triggered any special logic, but it did ensure the variable was defined.

I haven't had enough time to look through all of the logic where something pds specific needs to happen to feel comfortable re-adding the property with a blank string and expecting it to use the pds commons/template.  For now, I'm going to try installing 1.6.0-rhpds with the value defaulted to pds to unblock the 1.6/pds testing.

PR #1181, opened 2019-12-02
update datasync templates

wtrocki:


david-martin:
Any steps required to update templates during upgrade?

wei-lee:
@david-martin yes, there are. Will add that now.

wei-lee:
This is replaced by #1183

PR #1185, opened 2019-12-03
3scale upgrade 2.6 to 2.7

pmccarthy:
## Additional Information
Adding upgrade task in 3Scale role for upgrading to v2.7

## Verification Steps

- [ ] Run upgrade playbook against an existing v1.5.2 RHMI environment

PDS based validation:
```
ansible-playbook -i inventories/pds.template playbooks/upgrades/upgrade.yaml
```

The playbook is automating this process: https://access.redhat.com/documentation/en-us/red_hat_3scale_api_management/2.7/html/migrating_3scale/upgrading-3scale


roles/3scale/tasks/upgrade.yml
```diff
@@ -0,0 +1,287 @@
+---
```
david-martin: @pmccarthy Is this process generic for 3scale upgrades, or specific for 2.6 to 2.7?

If specific, can we rename it to avoid confusion with future ugprades potentially trying to use it?

pmccarthy: Makes sense, I'll rename it now

david-martin:
/lgtm

Running upgrade now after following the solution pattern with 3scale in it.
Will verify the created API still works after upgrade.

Will also verify the solution pattern with another user after upgrade, and approve then.

PR #1198, opened 2019-12-11
fix: handle walkthrough upgrade/install and restore old work queue/spring templates

evanshortiss:
## Additional Information
Fixes walkthrough 1A (and subsequent walkthroughs that depend on it)

## Verification Steps

* Click "Get Started" for walkthrough 1a
* Verify that the project namespace "rhmi-lab-nodejs-order-ui" item deploys successfully

### Installation Verification

### Upgrade Verification

- [x] Yes
- [ ] No

- [ ] Tested Installation
- [ ] Tested Uninstallation


evanshortiss:
@tonyxrmdavidson @pb82 this is a fix for something I missed during dev/test

I had *already* applied this template to my cluster. So the fact it was missing from the installer was overlooked since it didn't cause a failure due to already existing on the cluster

tonyxrmdavidson:
\LGTM

trepel:
lgtm'd based on @tonyxrmdavidson comment

evanshortiss:
@david-martin 

> Should the existing template named nodejs-messaging-work-queue-frontend be removed from existing clusters during upgrade?

I'm not sure. If someone is using it and we remove it could it cause issues?

Has an 

> Has an upgrade task been added to create the new template in existing clusters?

Nope. Happy do this, though I'm not familiar with it.

evanshortiss:
FYI chatted to Dave in GChat to get answers. Adding upgrade.  Leaving in the old nodejs messaging app to be safe in case users are relying on it.

evanshortiss:
@david-martin so what I've done here is anytime a fresh install or upgrade runs the following happens:

1. All our walkthroughs are uninstalled. If they don't exist, e.g fresh install, this step will still pass
2. Install all walkthroughs

This would mean fresh and upgraded clusters always run the latest versions of walkthroughs

I'll test it shortly

roles/walkthroughs/tasks/walkthroughs.yml
```diff
@@ -1,8 +1,13 @@
 ---
+- name: Remove walkthroughs
+  include_tasks: 'uninstall.yml'
```
evanshortiss: To keep things clean, simply uninstall all our walkthroughs first, then reinstall. Should work for both upgrade and install. 

This will be helpful too for people testing/developing since it creates a clean slate each time, even if they messed with templates manually during dev.

roles/walkthroughs/templates/crud_spboot_example.yml
```diff
@@ -0,0 +1,269 @@
+apiVersion: v1
```
evanshortiss: Restoring this since some users might be reliant on it.

inventories/group_vars/all/manifest.yaml
```diff
@@ -1,5 +1,5 @@
 ---
-integreatly_version: master
+integreatly_version: release-1.6.0-er1
```
evanshortiss: When I changed target branch this was highlighted. Want me to change it back to `master`?

evanshortiss:
Tested this and it's now working. One issue though is that the new Solution Patterns aren't being loaded using the new [1.10.0 tag](https://github.com/integr8ly/tutorial-web-app-walkthroughs/releases/tag/1.10.0). 

Should I update [this line in the webapp defaults](https://github.com/integr8ly/installation/blob/master/roles/webapp/defaults/main.yml#L14-L15) to install the new patterns?

playbooks/upgrades/upgrade.yaml
```diff
@@ -24,7 +24,7 @@
       shell: oc get nodes
       register: task_result
       until: task_result.rc == 0
-      retries: 6
+      retries: 30
```
evanshortiss: Sometimes this took a while for me, so I bumped the retries

roles/walkthroughs/tasks/crud_spboot_example.yml
```diff
@@ -0,0 +1,14 @@
+---
+# ADDS CRUD SPRING BOOT EXAMPLE APP TO THE CATALOG
+- name: template crud spring boot example app
+  template:
+    src: crud_spboot_example.yml
+    dest: /tmp/crud_spboot_example.yml
+    
+- name: Add crud spring boot example app to the catalog
+  command: oc create -f /tmp/crud_spboot_example.yml -n openshift
```
wei-lee: probably `oc apply` will make things a bit simpler as you don't need to worry if it's already existed.

roles/webapp/defaults/main.yml
```diff
@@ -12,7 +12,7 @@ webapp_operator_resource_items:
   - "{{ webapp_operator_resources }}/crd.yaml"
   - "{{ webapp_operator_resources }}/operator.yaml"
 webapp_walkthrough_locations:
-  - "https://github.com/integr8ly/tutorial-web-app-walkthroughs#v1.9.0"
+  - "https://github.com/integr8ly/tutorial-web-app-walkthroughs#v1.10.0"
```
evanshortiss: New walkthroughs tag for 1.6

evanshortiss:
I've tested upgrading from 1.5.2 and it worked. Also, installing fresh is working too.

roles/walkthroughs/tasks/crud_spboot_example.yml
david-martin: I agree oc apply is usually simpler, but this task is only run at install time, or after the uninstall task. I won't hold up the PR without this change

david-martin:
All comments & questions look to be addressed.
/approve

david-martin:
Will cherrypick to v1.6

PR #1200, opened 2019-12-11
make sure syndesis-operator CLI is running with the right context file

wei-lee:
## Additional Information

https://issues.redhat.com/browse/INTLY-4508

### DIFF
--- a/inventories/group_vars/all/common.yml
+++ b/inventories/group_vars/all/common.yml
@@ -106,3 +106,7 @@ threescale_storage_s3_aws_access_key: "{{ eval_threescale_storage_s3_aws_access_
 threescale_storage_s3_aws_secret_key: "{{ eval_threescale_storage_s3_aws_secret_key | default('null') }}"
 threescale_storage_s3_aws_region: "{{ eval_threescale_storage_s3_aws_region | default('null') }}"
 threescale_storage_s3_aws_bucket: "{{ eval_threescale_storage_s3_aws_bucket | default('null') }}"
+
+# The default location where kubeconfig will be saved on when the job is running from Tower. For now this is used by the Fuse operator CLI.
+# You should ALWAYS check if the file exists on the cluster before using it as it doesn't exist on PDS.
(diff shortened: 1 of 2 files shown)


damienomurchu:
Changes look good. These changes are currently being used to provision a POC cluster (job [here](https://integreatly-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/openshift-cluster-integreatly-test/357/console)) which should verify fix works for POC clusters.

Re verifying fix against OSD, I believe that should be verified at some point today by @jhellar

inventories/group_vars/all/common.yml
```diff
@@ -106,3 +106,7 @@ threescale_storage_s3_aws_access_key: "{{ eval_threescale_storage_s3_aws_access_
 threescale_storage_s3_aws_secret_key: "{{ eval_threescale_storage_s3_aws_secret_key | default('null') }}"
 threescale_storage_s3_aws_region: "{{ eval_threescale_storage_s3_aws_region | default('null') }}"
 threescale_storage_s3_aws_bucket: "{{ eval_threescale_storage_s3_aws_bucket | default('null') }}"
+
+# The default location where kubeconfig will be saved on when the job is running from Tower. For now this is used by the Fuse operator CLI.
+# You should ALWAYS check if the file exists on the cluster before using it as it doesn't exist on PDS.
+kube_config_file: "/var/lib/awx/.kube/config"
```
pmccarthy: This could break RHPDS deployments. Ideally this should be empty by default in common.yml and set in the poc and osd group_vars instead.

pmccarthy: After a second look this would work as the logic is handling whether or not the file exists. My concern is with referencing the path in common.yml as it's led to problems previously with group_vars. IMO it's safer to instead set these types of vars in the relevant environment group_vars as really this change is specific to OSD and POC environments.

damienomurchu: yes, had similar thoughts, logic should work but better to confine them to environment group_vars I think too

PR #1204, opened 2019-12-13
Cherrypick 1198

david-martin:
## Additional Information
Cherrypick of commits from https://github.com/integr8ly/installation/pull/1198/commits


inventories/group_vars/all/manifest.yaml
```diff
@@ -1,5 +1,5 @@
 ---
-integreatly_version: release-1.6.0-er1
+integreatly_version: master
```
wei-lee: should this be changed?

playbooks/upgrades/upgrade.yaml
```diff
@@ -265,6 +266,17 @@
...
+    - name: Update walkthroughs
+      include_role:
+        name: walkthroughs
+        tasks_from: upgrade
+    
+    # User SSO (evals realms for new walkthroughs)
+    - name: Create evalsNN realms in user sso for new walkthroughs
+      include_role:
+        name: rhsso-user
+        tasks_from: create-eval-user-realms
```
wei-lee: will this run in all environments? I thought POC and OSD environments don't have evals users setup?

david-martin: The number of eval users would be 0 in those environments.
@pmccarthy can you confirm, and as such this block wouldn't end up creating any realms in the user sso

pmccarthy: @david-martin @wei-lee Correct, on OSD clusters eval user count is set to 0, see https://github.com/integr8ly/installation/blob/master/inventories/group_vars/osd/osd.yml#L34-L36

Hoping that the playbooks should handle that and move on

inventories/group_vars/all/manifest.yaml
david-martin: reverted

PR #1210, opened 2019-12-17
Update solution explorer to 2.20.8 and operator to v0.0.40

mfrances17:
## Additional Information
Update webapp to 2.20.8 and the webapp operator to v0.0.40. Install log attached, see that the install completed successfully and that the manifest now shows 2.20.8 as the webapp version.

[121619_uxddev-b62c_install.log](https://github.com/integr8ly/installation/files/3971437/121619_uxddev-b62c_install.log)


davidffrench:
Peter Braun is on PTO @mfrances17 .

PR #1234, opened 2020-01-08
create one shared realm for the walkthroughs

wei-lee:
## Additional Information
https://issues.redhat.com/browse/INTLY-4660

## Verification Steps
Run the upgrade playbook and when it finishes login to the user-sso. There should be only one realm created for the walkthroughs (it's called "walkthroughs").

### DIFF
--- a/roles/rhsso-user/tasks/create-eval-user-realms.yml
+++ b/roles/rhsso-user/tasks/create-eval-user-realms.yml
@@ -1,5 +1,5 @@
 ---
-- name: Get user SSO route for evalsNN realm creation
+- name: Get user SSO route for walkthroughs realm creation
   shell: oc get route/sso -o template --template \{\{.spec.host\}\} -n {{ eval_user_rhsso_namespace | default('user-sso') }}
   register: user_sso_route_cmd
 
@@ -58,9 +58,10 @@
 
 - debug: var=user_sso_token
 
-- name: Create evals user realms
+- name: Create the walkthroughs realm
   include_tasks: ./create-realm.yml
-  with_sequence: count={{ eval_seed_users_count }} format=evals%02d
+  with_items:
+    - walkthroughs
 
 - name: Create customer-admin Realm JSON from template
   template: src=evals-user-realm.j2 dest=/tmp/customer-admin-realm.json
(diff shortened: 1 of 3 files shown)


wei-lee:
@evanshortiss mind reviewing this PR?

roles/rhsso-user/tasks/create-eval-user-realms.yml
```diff
@@ -58,9 +58,10 @@
 
 - debug: var=user_sso_token
 
-- name: Create evals user realms
+- name: Create the walkthroughs realm
   include_tasks: ./create-realm.yml
-  with_sequence: count={{ eval_seed_users_count }} format=evals%02d
+  with_items:
+    - walkthroughs
```
evanshortiss: @wei-lee so the usage we're going for here is that any user using this will login as `walkthroughs@example.com`/`password`?

👍 from me, in the future it'll be good if we can find a cleaner UX but this is good to get things working now

wei-lee: @evanshortiss yes, that's right. Then each user will create their own clients in the realm.
 
We will make it clear in the WT as well the username/password that user should use to login and continue with the WTs, and warn them that the realm is shared with other people. It is just for the demonstration of the WT only, and should not be used for production environments.

wei-lee:
Deployed this this cluster: Openshift Master Console: https://master.release-bef6.open.redhat.com:443

roles/rhsso-user/tasks/create-eval-user-realms.yml
evanshortiss: @wei-lee could a user change the `walkthroughs@example.com` password? If they can, maybe we want to prevent them doing that somehow? It's an edge case of course, so not a major concern.

wei-lee: @evanshortiss good question. I don't think we can prevent them from changing the password, but I am not sure. I think we can do is call out in the WT that user shouldn't change the password for this user as it's a shared one. @pwright wdyt?

davidffrench:
/approve

approving based on previous comments

PR #1236, opened 2020-01-10
Updating predictive PVC checks to check that PVC still exists

alanmoran:
## Problem
Our new predictive PVC checks will alert for PVC's that have been deleted. This occurs as we use a range selector when looking at past data. ie [6h] is data from 0 -> 6 hours ago. 

## Verification
PR has been verified by CS SRE in rhmi-cssre cluster. 
1) Create a new pod that is PV backed (mongod from image catalogue is a good candidate) - Must be in an RHMI namespace (label middleware) 
2) `oc rsh <mongo-pod>` and dump some files into the volume path (df -h).(`dd if=/dev/zero of=file.out bs=1MB count=500` will create a 500mb file for you)
3) `oc rsh <any rhmi pod that is PV backed>` and do the same as above
4) Verify that Prometheus is showing both PVC's in the alert/query screen
Screen Shot 2020-01-10 at 12 57 04 pm
5) Delete Mongo pod/service (this should delete the PVC)
6) Verify that Prometheus is still reporting the deleted PVC
7) Test new queries. Results should now only return one PVC 
Screen Shot 2020-01-10 at 1 00 27 pm

### DIFF
--- a/roles/middleware_monitoring_config/templates/kube_state_metrics_alerts.yml.j2
+++ b/roles/middleware_monitoring_config/templates/kube_state_metrics_alerts.yml.j2
@@ -212,7 +212,7 @@ spec:
         annotations:
           message: The {{ '{{' }} $labels.persistentvolumeclaim {{ '}}' }} PVC will run of disk space in the next 4 days.
         expr: |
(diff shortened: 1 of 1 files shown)


david-martin:
/lgtm
/approve

Tested on OSD4 (by copying the expression query).
Note that this cannot be tested on RHPDS as volume stats are unavailable with the storage type used there.

PR #1242, opened 2020-01-16
Add support for specifying any number of additional seed users to be created/configured during install

jessesarn:
[INTLY-4127](https://issues.redhat.com/browse/INTLY-4127)

## Additional Information
Can be used to create additional cluster admins, customer admins, and/or non-admin RHMI users by overriding the rhsso_seed_user_list via inventory, vars file, cmd line.

For customer admins, they get the same special roles/objects as the primary/standard customer-admin (role binder role for managed-fuse, their own realm in user-sso, admin user created in 3scale, and, for osd only, are added to dedicated-admins group automatically.

Non-admin users get the same access/objects created as evals users (access to 3scale and own realm created in user-sso).

Cluster admins are given the cluster-admin cluster role; however, they don't get anything special within rhmi.

### Structure is really simple, there are 4 fields:
 - username: ...       # only required field is username
   password: ...   # if missing/empty, one is randomly generated (initial value can be looked up from the corresponding secret only if it hasn't been changed)
   email: ...      # if missing/empty; appends @example.com to username; otherwise, 3scale setup would fail for user
   admin: ...      # supported values: "cluster", "customer", "no" (default if missing/empty)

#### Sample yaml:
```
rhsso_seed_user_list:
  -
    username: cssre-admin
    password: 'C@/\/-u-3|/e|\/-R3@c|-|\/|3'
    admin: cluster
  -
    username: extra-customer-admin
    admin: customer
  -
    username: basic-user
```

#### Sample for cmd line override using json:
```
-e '{"rhsso_seed_user_list": [{"admin":"cluster","password":"C@/\/-u-3|/e|\/-R3@c|-|\/|3","username":"cssre-admin"},{"admin":"customer","email":"extra-customer-admin@example.com","username":"extra-customer-admin"},{"username":"basic-user"},{"username":"tricky-password-user1","password":"AHnb7,ANjzK4wE:"}]}'
```

## Verification Steps

1. Run the install playbook while specifying a value for the rhsso_seed_user_list variable containing a custom list of additional users to be created during the install.  (see above for example in json format).
2. Verify that the additional users have been created properly with the proper access based on their admin value.

### Installation Verification
### Upgrade Verification

- [ ] Yes
- [x] No

- [x] Tested Installation
- [x] Tested Uninstallation


jessesarn:
@pb82 Looks like it still needs a /lgtm from somebody before it will merge.  Can you please add that as well.  thanks.

PR #1246, opened 2020-01-24
Upgrade playbook updates v1.6

mikenairn:
## Additional Information
* Replace playbooks/upgrades/upgrade.yml with symlink to playbooks/upgrade.yml
* Reset upgrade variables after final release release-1.6.0

The upgrade vars should have been updated automatically when using the release script, but weren't https://github.com/integr8ly/installation/blob/master/scripts/release.sh https://github.com/integr8ly/installation/blob/master/docs/release.md#installation-repo

The removal of the upgrades/upgrade.yml playbook was something we needed to do after 1.6.0 was released. This also needs to be applied to master.


pb82:
@mikenairn so how will upgrades work in the Future? It looks like `playbooks/upgrades/upgrade.yml` only calls into [this](https://github.com/integr8ly/installation/blob/master/playbooks/upgrade.yml) which will only runs the `upgrade_images` role of the products in `upgrade_product_roles`. Where should we implement additional upgrade logic?

mikenairn:
@pb82 Additional upgrade logic should just be added to playbooks/upgrade.yml like you would have done in playbooks/upgrades/upgrade.yml previously.

If you are upgrading a product version, it should be listed in here https://github.com/integr8ly/installation/blob/master/playbooks/group_vars/all/upgrade.yml#L2 which will call the upgrade images role for that product, which in most cases should be enough for a patch release.

The updating of that file is automated for anything DeLorean is monitoring.

pb82:
@mikenairn there's a couple of automatically generated product upgrade PRs, e.g. https://github.com/integr8ly/installation/pull/1224

They are currently targetting master but since those are patch version updates, we should probably cherry pick them into the v1.6 branch and test them there?

PR #1247, opened 2020-01-24
Replace playbooks/upgrades/upgrade.yml

mikenairn:
## Additional Information
Replace playbooks/upgrades/upgrade.yml with symlink to playbooks/upgrade.yml

Related:

https://github.com/integr8ly/installation/pull/1246


playbooks/upgrades/upgrade.yml
```diff
@@ -0,0 +1 @@
+../upgrade.yml
```
pmccarthy: If this is just a symlink, why are we keeping this file? Tower does reference it but we can update the playbook path here: https://github.com/integr8ly/ansible-tower-configuration/blob/master/playbooks/roles/integreatly/tasks/bootstrap_osd_update.yml#L84

Also didn't realise that playbook path is hardcoded in the tower-config so will need to make this a variable, not relevant here though.

mikenairn:
/hold cancel

@pb82 @pmccarthy Can we get an lgtm on this please.

PR #1256, opened 2020-01-28
Replace any hyphens in role names when setting upgrade facts.

mikenairn:
Replace any hyphens in role names when setting upgrade facts.

e.g. role = rhsso-user, fact = upgrade_rhssouser


mikenairn:
/close

@dimitraz Will roll this change into her own PR where it is required. That PR will also need to be cherrypicked back into v1.6

PR #1268, opened 2020-02-11
[WIP] codeready update

fh-automatron:


wei-lee:
@ciaran-byrne do you think if this is still required? It's a patch release for Codeready on the 1.6 release.

ciaran-byrne:
Chatted with @jasonmadigan on this one, and we are agreed this is not needed for now (even for a RHMI 1.6 patch release) as the newer version of CRW 2.0 is already done

PR #1112, opened 2019-11-02
codeready update

fh-automatron:
Created automatically by jenkins

PR #1273, opened 2020-02-14
install codeready using the cli tool

wei-lee:
## Additional Information
https://issues.redhat.com/browse/INTLY-5336

## Verification Steps

### Installation Verification
- Provision an OpenShift 3.x cluster
- Run the installation playbook. Wait until Codeready is installed.
- Login to Codeready and create a workspace. It should work.


wei-lee:
@damienomurchu can you try this branch for the pipeline? it should fix the issue you are having.

damienomurchu:
@wei-lee just seeing this now - will try pds install from this branch and see if it resolves issue

damienomurchu:
@wei-lee - verifying changes resolved issue when run against a pds cluster.

Do you we need upgrade logic in this PR?

wei-lee:
@damienomurchu upgrade is done in a different PR: https://github.com/integr8ly/installation/pull/1112 which is already merged.

pb82:
@damienomurchu upgrade logic was already verified, i'm also currently running an installation from this branch.

PR #1275, opened 2020-03-03
Disable gitea by default

grdryn:
Gitea was only used for RHPDS workshops, but it's not actually enabled
there any longer, which means that to test upgrade from 1.6.0 clusters
on RHPDS, we have to override it in the ansible-playbook command to
get the upgrade to work.

Maybe all of the Gitea stuff should just be removed from this repo?


pmccarthy:
@grdryn Not sure there's a need to validate the installation on all 3 environments here (OSD,POC,RHPDS). Change is very straightforward so happy to approve 👍 

/lgtm

PR #1282, opened 2020-03-18
Use kube_job_failed metric to check for job completion spec

pepedocs:
## Additional Information

**Jira Issue**: CSSRE-806
**Expected Behavior**: `CronJobsFailed` alert should only fire if a job completion spec is not met
**Observed Behavior**: `CronJobsFailed` alert fired even though job completion was met.

The following are job info for a job that caused `CronJobsFailed` alert to fire. Notice that the completion spec was met. Note that I can't put all the information here as it contains confidential information. Please comment on the Jira issue for more information.

```
Parallelism:    1
Completions:    1
Start Time:     
Pods Statuses:  0 Running / 1 Succeeded / 3 Failed
```

**Impact for CSSRE**

The `CronJobsFailed` alert fires false-positives at least twice in a month since February 2020 and CSSRE team needs to investigate this to ensure no jobs failed. Although this seems like a minor issue, this makes the CronJobsFailed alert unreliable which is counter-productive for the CSSRE team. 

It would be great to fix this issue whichevery way so that we don't develop a habit to ignore this alert.

**Explanation of Changes**
`kube_job_status_failed` seems to only check for the number of failed attempts the job has tried. It does not check for the job completion spec. There is a relevant [issue](https://github.com/kubernetes/kube-state-metrics/issues/367) which already somewhat figured out that `kube_job_failed` is checking for the job completion spec. Looking at the [Go code](https://github.com/kubernetes/kube-state-metrics/blob/507a2f8ce67c35d01a13504b4cc60a171d0cf8ba/collectors/job.go#L208) in the said issue, it looks like `kube_job_status_failed` is really just counting the number of failures. This will cause a false-positive alert since some jobs can have failed attempts but eventually will succeed. 

Note: I'm not quite sure which kube-state-metrics we are using so the code I've referenced might have an advanced version.

**Test**
To test this without needing to update this installation template, I have simply run the new query in prometheus and that cleared all missing cron job failed status.

## Verification Steps
`CronJobsFailed` alert should only fire when job completion spec is not met. Eventually this will stop false-positives coming form this alert.

### Installation Verification
Not sure
### Upgrade Verification
Not sure

### DIFF
--- a/roles/backup/templates/backup-monitoring-alerts.yml.j2
+++ b/roles/backup/templates/backup-monitoring-alerts.yml.j2
@@ -39,7 +39,7 @@ spec:
           sop_url: https://github.com/RHCloudServices/integreatly-help/blob/master/sops/alerts_and_troubleshooting.md
           message: 'Job {{ '{{' }} $labels.namespace {{ '}}' }}/{{ '{{' }} $labels.job {{ '}}' }} has failed'
         expr: >
-          clamp_max(max(kube_job_status_start_time * ON(job) GROUP_RIGHT() kube_job_labels{label_cronjob_name!=""} ) BY (job, label_cronjob_name, namespace) == ON(label_cronjob_name) GROUP_LEFT() max(kube_job_status_start_time * ON(job) GROUP_RIGHT() kube_job_labels{label_cronjob_name!=""}) BY (label_cronjob_name), 1) * ON(job) GROUP_LEFT() kube_job_status_failed > 0
+          clamp_max(max(kube_job_status_start_time * ON(job) GROUP_RIGHT() kube_job_labels{label_cronjob_name!=""} ) BY (job, label_cronjob_name, namespace) == ON(label_cronjob_name) GROUP_LEFT() max(kube_job_status_start_time * ON(job) GROUP_RIGHT() kube_job_labels{label_cronjob_name!=""}) BY (label_cronjob_name), 1) * ON(job) GROUP_LEFT() kube_job_failed > 0
         for: 5m
         labels:
           severity: critical


david-martin:
@pepedocs This is very interesting.
We saw similar behaviour to what you describe before, and changed the query to what it is now to cater for the scenario of previously failed jobs with the latest being a success. The changes were made in https://github.com/integr8ly/installation/pull/1065.

Based on those changes and the testing done at the time, I'm surprised there's still a problem.
However, if you believe changing to `kube_job_failed` is the right move, and the alert behaves as expected with that change, let's make the change.

pepedocs:
Hi @david-martin,

> We saw similar behaviour to what you describe before, and changed the query to what it is now to cater for the scenario of previously failed jobs with the latest being a success

I've looked into the changes. It seems like they were for aggregating the jobs and also to query only the latest job that was run, however, it was still using the query `kube_job_status_failed` which I suspect is not checking the job spec.

> However, if you believe changing to `kube_job_failed` is the right move, and the alert behaves as expected with that change, let's make the change.

The code I've referenced above and also the relevant issue, confirmed my suspicion that `kube_job_status_failed` is simply counting the number of job failures. However, to test my suspicion I made a little experiment using minikube/prometheus/kube-state-metrics setup. Ideally I should've tested this in a cssre test cluster, however, I think simulating this job failures is a bit complicated in there. In any case, I'm only testing the prom query which has nothing to do with any cluster;

Job: **succfail-1584692400**
This job is designed to have failures but will also have at least 1 success.
```
[jcueto@localhost minikube-prometheus-demo]$ kb describe job -n sandbox succfail-1584692400
Name:           succfail-1584692400
Namespace:      sandbox
Selector:       controller-uid=9dbac0f8-5e02-488b-9283-909b49f7b995
Labels:         controller-uid=9dbac0f8-5e02-488b-9283-909b49f7b995
                job-name=succfail-1584692400
Annotations:    <none>
Controlled By:  CronJob/succfail
Parallelism:    1
Completions:    1
Start Time:     Fri, 20 Mar 2020 18:20:14 +1000
Completed At:   Fri, 20 Mar 2020 18:23:26 +1000
Duration:       3m12s
Pods Statuses:  0 Running / 1 Succeeded / 3 Failed
Pod Template:
  Labels:  controller-uid=9dbac0f8-5e02-488b-9283-909b49f7b995
           job-name=succfail-1584692400
  Containers:
   succfail:
    Image:      python:latest
    Port:       <none>
    Host Port:  <none>
    Args:
      python
      -c
      from random import randint as rint;import sys;sys.exit(0) if rint(1,10)%2==0 else sys.exit(-1)

```
Query: **kube_job_status_failed**
image

Query: **kube_job_failed**
image

Also to make sure **kube_job_failed** returns something for a real failed job I simulated an always failed job.

Job: **fail-1584692880**

```
[jcueto@localhost minikube-prometheus-demo]$ kb describe job -n sandbox fail-1584692880
Name:           fail-1584692880
Namespace:      sandbox
Selector:       controller-uid=1b2fc33e-cefa-4b91-9e80-ada47ec10a07
Labels:         controller-uid=1b2fc33e-cefa-4b91-9e80-ada47ec10a07
                job-name=fail-1584692880
Annotations:    <none>
Controlled By:  CronJob/fail
Parallelism:    1
Completions:    1
Start Time:     Fri, 20 Mar 2020 18:28:04 +1000
Pods Statuses:  0 Running / 0 Succeeded / 5 Failed
Pod Template:
  Labels:  controller-uid=1b2fc33e-cefa-4b91-9e80-ada47ec10a07
           job-name=fail-1584692880
  Containers:
   fail:
    Image:      python:latest
    Port:       <none>
    Host Port:  <none>
    Args:
      python
      -c
      from random import randint as rint;import sys;sys.exit(-1)

```
Queries: **kube_job_failed** counts 1, while **kube_job_status_failed** counts every failed attempt.

image


I hope this helps.

david-martin:
@pepedocs Thanks for the explanation.
So what is the desired behaviour with regards to ordering of success/failures?
The query was written in a way that should fire if the last job failed, and stop firing if the last job succeeded. i.e. It's OK if there were many failures, as long as the most recent one (which may have been the scheduled one or one SRE ran manually after fixing some issue) succeeded.

pepedocs:
@david-martin 

> So what is the desired behaviour with regards to ordering of success/failures?

As far as I've observed, `kube_job_failed` checks for the overall job completion status. If the job has not completed, it will return nothing, otherwise it will return something (truthy value) if and only if the job spec was not met. For example, `Completion=1, Success=0, Failed=6` (no success). If the job spec was met it will return nothing otherwise.

On the other hand, `kube_job_status_failed` counts the number of failure attempts. So regardless of whether the job has completed successfuly or not, it will return something (truthy value) if `Failed>=1` otherwise it will return nothing. 


> It's OK if there were many failures, as long as the most recent one (which may have been the scheduled one or one SRE ran manually after fixing some issue) succeeded.

`kube_job_failed` checks for the overall job completion status, therefore it will satisfy this condition because a job completes immediately when its job spec was met.

`kube_job_status_failed` won't be able to satisfy this condiction because "many failures" will return a truthy value, because it counts the failure attempts of the job run. 

Thanks

pepedocs:
@david-martin  Thanks for reviewing. May I know which release I should expect for this change to take effect?

david-martin:
@pepedocs  Once merged to master, it will be included in the next minor release (1.7.0 at this time).

It needs an lgtm before merging, @alanmoran or @obrienrobert should be able help there

obrienrobert:
Reviewed the logic, and reasoning behind the change.

PR #1285, opened 2020-03-18
fuse-online update

fh-automatron:


jessesarn:
The issue with this updated fuse release version that was causing this to be held from merging has been resolved.  I double checked that 1.9.12 was still used for the release version.

PR #1287, opened 2020-03-19
Filtering which https-metrics with federate from cluster prom

alanmoran:
## Additional Information
This fix resolves the issue we are seeing in https://issues.redhat.com/browse/CSSRE-830.
In an example cluster, we pull ~150000 metrics every 30 seconds. We only need access to ~100 of those. The metric pull is so large that we hit the 10-second scrape timeout. 

Below is a graph of scrape duration in seconds. As you can see we've had a noticeable change in timings going from ~10/12 seconds - ~2/4 seconds
image

cc @david-martin as this will need to be applied to 2.x

### DIFF
--- a/roles/middleware_monitoring_config/templates/jobs/openshift_monitoring_federation.j2
+++ b/roles/middleware_monitoring_config/templates/jobs/openshift_monitoring_federation.j2
@@ -9,7 +9,8 @@
   metrics_path: /federate
   params:
     match[]:
-    - '{{ "{" }} endpoint="https-metrics" {{ "}" }}'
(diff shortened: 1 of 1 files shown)


david-martin:
@StevenTobin Would you mind casting an 👁  over this as the original matching filter was added by yourself. I think it was just to get kubelet_volume metrics for the new pv alerts added at the time. #1031

alanmoran:
@StevenTobin and @david-martin This slipped off the radar. Are we still good?

StevenTobin:
Also I didn't seem to get notifications for this so just seeing now, apologies for that.

PR #1283, opened 2020-03-18
[WIP] amq-online update

fh-automatron:
Created automatically by jenkins

PR #1292, opened 2020-03-26
disable enmasse own monitoring as it breaks application monitoring stack install

austincunningham:
## Verification Steps

### Installation Verification
### Upgrade Verification

## Checklist

- [ ] Tested Installation
- [ ] Tested Uninstallation

- [ ] Yes
- [x] No

### DIFF
--- a/roles/enmasse/templates/enmasse_hosts.j2
+++ b/roles/enmasse/templates/enmasse_hosts.j2
@@ -18,5 +18,5 @@ monitoring_namespace=enmasse-monitoring
 # Don't need to install a monitoring stack. Will use existing middleware monitoring one.
 monitoring_operator=False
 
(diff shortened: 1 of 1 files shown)


matskiv:
This needs to be tested with #1283 right?
We could change this PR to target `amq-online-ga` branch instead of master.

jessesarn:
Does somebody have install/upgrade logs already?

matskiv:
@jessesarn I don't. I haven't ran this. I was expecting we would just merge this to `amq-online-ga` branch, and then verify PR from `amq-online-ga` to `master`

jessesarn:
Good point.  Since its going to the amq-online-ga branch first, for further testing, I see no problem with merging.
/approve

PR #1297, opened 2020-04-03
CSSRE-518: Added variables and config for alertmanager slack integration

pepedocs:
## Additional Information
1. Added slack api and default alerts channel variables for alertmanager config 
2. These variables will be populated either at install or post install

`alertmanager_slack_api_url` - A slack webhook/api url (incoming) for sending alerts from alertmanager to slack
`alertmanager_slack_default_channel` - The default slack channel that is mapped to the default alert manager receiver. This adds an additional optional receiver aside from email.

## Verification Steps
The alertmanager config was tested using the team 's test cluster. This was done in an ad-hoc way, specifically modifying the alertmanager config with test values for the said variables.

### Installation Verification
NA

## Checklist
NA

- [x] Yes
- [ ] No

To test this, the said variables must be set during or after installtion/upgrade

### DIFF
--- a/roles/middleware_monitoring_config/defaults/main.yml
+++ b/roles/middleware_monitoring_config/defaults/main.yml
@@ -48,3 +48,8 @@ dms_webhook_url: "{{ deadmanssnitch_url | default('') }}"
 
 # pagerduty
 pd_service_key: "{{ pagerduty_integration_key | default('') }}"
+
+# slack
+alertmanager_slack_api_url: "{{ alertmanager_slack_api_url | default('') }}"
+alertmanager_slack_default_channel: "{{ alertmanager_slack_default_channel | default('') }}"
+
--- a/roles/middleware_monitoring_config/templates/alertmanager.yml.j2
+++ b/roles/middleware_monitoring_config/templates/alertmanager.yml.j2
@@ -6,6 +6,9 @@ global:
   smtp_auth_username: {{ smtp_auth_username }}
   smtp_auth_password: {{ smtp_auth_password }}
 {% endif %}
+{% if alertmanager_slack_api_url %}
+  slack_api_url: {{ alertmanager_slack_api_url }}
+{% endif %}
 route:
   group_wait: 30s
   group_interval: 5m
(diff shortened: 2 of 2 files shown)


roles/middleware_monitoring_config/templates/alertmanager.yml.j2
```diff
@@ -27,6 +28,11 @@ receivers:
   - send_resolved: true
     to: {{ alertmanager_to_email }}
     html: '{{ '{{' }} template "email.enhanced.html" . {{ '}}' }}'
+  slack_configs:
```
alanmoran, line 31: Can we wrap this in an if so it is only added if alertmanager_slack_api_url  is present? I can't see Eng/QE using this functionality.

roles/middleware_monitoring_config/defaults/main.yml
```diff
@@ -48,3 +48,8 @@ dms_webhook_url: "{{ deadmanssnitch_url | default('') }}"
 
 # pagerduty
 pd_service_key: "{{ pagerduty_integration_key | default('') }}"
+
+# slack
+alertmanager_slack_api_url: "{{ alertmanager_slack_api_url | default('') }}"
+alertmanager_slack_default_channel: "{{ alertmanager_slack_default_channel | default('') }}"
```
alanmoran, line 54: What happens if the url is populated but this is left as ''?

pepedocs: If you mean the channel is left '', slack allows a default channel when a webhook is set up. However, I'm not quite sure about alertmanager if it allows no channel is set up.

roles/middleware_monitoring_config/templates/alertmanager.yml.j2
pepedocs: Yes

roles/middleware_monitoring_config/defaults/main.yml
pepedocs: I tried sending a message without a channel and the http response was a 404, but this is nothing different to a user misspelling a channel/url name. It's cmoplicated to validate the variables.

pepedocs:
@alanmoran  I've added an `if` statement in the alertmanager config. Please review.

pepedocs:
/assign @philbrookes

Can you please approve? Thanks.

PR #1301, opened 2020-04-10
Intly 6500 update 3scale

jessesarn:
## Additional Information
https://issues.redhat.com/browse/INTLY-6500

Upgrades 3scale from 2.7 to 2.8

## Verification Steps

### Installation Verification
### Upgrade Verification

Log of upgrade: https://gist.github.com/04ec9e140d222c6dd4e7e1b483f3e070
Diff of all pod/container images before vs. after: https://gist.github.com/jessesarn/04ec9e140d222c6dd4e7e1b483f3e070#file-diff-of-i8-get-image-report-before-vs-after

## Checklist

- [ ] Tested Installation
- [ ] Tested Uninstallation

- [ ] Yes
- [ ] No

### DIFF
--- a/inventories/group_vars/all/manifest.yaml
+++ b/inventories/group_vars/all/manifest.yaml
@@ -7,7 +7,7 @@ integreatly_version: master
 threescale: True
 # below threescale variables below are not currently used. They define where and what to pull in from github. These will be important as we look to break out and decouple the resources from the installer ###############
 # not currently used but records the version installed by the operator, this should come from the status block
-threescale_version: '2.7.1.GA'
+threescale_version: '2.8.0.GA'
 threescale_namespace: "{{ns_prefix | default('')}}3scale"
 threescale_display_name: "Red Hat 3scale API Management Platform"
 threescale_resources_base: 'https://raw.githubusercontent.com/3scale/3scale-amp-openshift-templates/{{threescale_version}}/amp'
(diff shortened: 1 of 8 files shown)


JameelB:
code changes looks good to me.

@jessesarn should the customisation not have picked up the 3scale version to be `2.8.0` instead of `2.7.0` [here](https://gist.github.com/jessesarn/04ec9e140d222c6dd4e7e1b483f3e070#file-integreatly-upgrade-log-L2112)?

jessesarn:
@JameelB Good catch.  Yeah, I would expect it should report the new version.  I'll run it again on my new cluster to see if I can find out what is missing.

jessesarn:
ah, we made a mistake.  The attached  integreatly-upgrade.log includes output from the upgrade from 1.6.0 -> 1.6.1 and then 1.6.1 -> master.

The 2.7.0.GA you saw in line 2112 which is still for the 1.6.1 upgrade.  You can tell its for the 1.6.1, because the paths for the 1.6.1 upgrade are all within /root/installation-1.6.1/...

The upgrade from 1.6.1 -> master begins on line [3822](https://gist.github.com/jessesarn/04ec9e140d222c6dd4e7e1b483f3e070#file-integreatly-upgrade-log-L3822).  You can't actually see the string used for the 3scale version in the later output because the logic for upgrades changed signficantly between v1.6 and master.  The logic in master does a straight delete and recreate of the manifest from the manifest.yml included in the installer inventories group_vars.

JameelB:
/lgtm

@jessesarn may need to ping an approver as i'm not on that list

jessesarn:
got it covered. I'm an approver ;)
/approve

PR #1302, opened 2020-04-10
codeready update

fh-automatron:


trdoyle81:
Verification [WIP]
1.7 install logs [here](https://gist.github.com/trdoyle81/a0d5c30816b9046a2977a0f6970fec49#file-intly-7027-v1-7-install-log-L89)

laurafitzgerald:
@trdoyle81 

Looks like in the past with the code ready version bumps we've been adding version specific upgrade files. see [1.0-1.2](https://github.com/integr8ly/installation/blob/master/roles/codeready/tasks/upgrade_1.0_to_1.2.yml) and [1.2-2.0](https://github.com/integr8ly/installation/blob/master/roles/codeready/tasks/upgrade_1.2_to_2.0.yml)

The latter looks to be a simple version bump path. We can add this to upgrade_images.yml.
As this file is currently empty I would expect that this automated change here does nothing.

By adding the change to upgrade_images.yml it means that when these automated prs come in in future there should be no manual changes required. Unfortunately the manual changes are required in this case.

I suggest leaving a note in the code so that future sheriffs know they don't need to add version specific upgrade files unless it is required. e.g. to manipulate openshift resources. 

cc @jessesarn

laurafitzgerald:
Also a note on the installation versioning here.

We're in a wierd state where we are currently doing a release so the upgrade from version on master is a little bit funky.

To verify these changes the path is.

Install `release-1.7.0-rc3` - most up to date version of 1.7.0
Upgrade using this branch - you will need to manually manipulate the upgrade_from version. Change it from `release-1.7.0` to `release-1.7.0-rc3` (most recent rc) .

StevenTobin:
Correct version after upgrade
version

Correct version images pulled by workspace creation
workspace version

Walkthrough 3 was also successfully completed

obrienrobert:
Verified on an RHPDS cluster with valid certificates using the following steps:

- Installed `release-1.7.0` (https://gist.github.com/obrienrobert/9fa2295ee6d5ad4afdb8e43150197f2b)
2 0 0 GA
- Completed an upgrade to this branch (https://gist.github.com/obrienrobert/d34313e7d3c3559f724005986d8972ab)
2 1 1 GA
- Verified the correct image is being pulled
2 1 1-image

PR #1313, opened 2020-04-22
Resetting the upgrade playbook on master

laurafitzgerald:
## Additional Information
https://issues.redhat.com/browse/INTLY-6827

The docs here mention that this is a manual task, however we found a [template](https://github.com/integr8ly/installation/blob/master/scripts/upgrade.template.yml) for resetting the upgrade playbook here. The change so far is to manually copy that template (it's scripted [here](https://github.com/integr8ly/installation/blob/master/scripts/release.sh#L119)) - the diff below is shown. Does it make sense to use the template going forward. It may require updates if there are things being removed here that we don't want removed. Haven't worked on this stuff in a while so any info on this would be great. 👍 

cc @mikenairn @jessesarn @JameelB 

### Upgrade Verification
This is resetting the upgrade so not sure what verification is required.

- [ ] Tested Installation
- [ ] Tested Uninstallation

- [ ] Yes
- [x] No

### DIFF
--- a/docs/release.md
+++ b/docs/release.md
@@ -51,7 +51,7 @@ To cut subsequent RCs for a patch release:
 ### Minor Release
 
 There may be logic in the upgrade playbook that is targetted at a specific release only.
-After the minor release branch is created, the upgrade playbook in `playbooks/upgrades/upgrade.yml` should be reviewed and reset on `master` to remove any version specific blocks, tasks or roles being included.
+After the minor release branch is created, the upgrade playbook in `playbooks/upgrade.yml` should be reviewed and reset on `master` to remove any version specific blocks, tasks or roles being included.
 
 As this is a manual task, here are some guidelines for doing the review.
 
--- a/playbooks/upgrade.yml
+++ b/playbooks/upgrade.yml
@@ -26,8 +26,6 @@
     set_fact: upgrade_{{ item | replace('-', '') }}=true
     with_items: "{{ upgrade_product_roles }}"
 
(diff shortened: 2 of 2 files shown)


mikenairn:
> The docs here mention that this is a manual task, however we found a template for resetting the upgrade playbook here. The change so far is to manually copy that template (it's scripted here) - the diff below is shown. Does it make sense to use the template going forward. It may require updates if there are things being removed here that we don't want removed.

@laurafitzgerald That script resets it on the release branch once the final release tag is set so that branch is ready to be updated for whatever potential patch updates might come along after on that branch.

The release script doesn't however deal with resetting it on the base branch(master) when the release is initially started since it doesn't *currently* create the branch, hence the manual step to reset the playbook.

However the steps required to rest it on the base branch (master) should be exactly the same and the template should be used as you are suggesting https://github.com/integr8ly/installation/blob/master/scripts/release.sh#L117-L120

laurafitzgerald:
Note on this one. Chatted to @JameelB and @pb82 this morning. There is no verification done here as we're resetting the upgrade. there are some following upgrade steps being added by myself and @trdoyle81 which will very the upgrade on master. I will create a follow on JIRA to update the docs to use the template so we can this one merged to unblock other changes.

JameelB:
/lgtm 

Code changes looks good to me and happy to merge based on the comments above.

@laurafitzgerald you'll need to assign an approver for approval so this can be merged as i'm not on that list.

PR #1317, opened 2020-04-24
Intly-6419 - Update fail message for 2nd run of upgrade

trdoyle81:
Updated out-message for 2nd run of upgrade to advise that target-version is already installed

## Additional Information
https://chat.google.com/room/AAAAS5TXmPc/_bLqNU-j4sE
@jcoleman0redhat0com advised that it's sufficient to just update the output of the playbook to say that a failure is expected in the case where the upgrade has already been run.

### Installation Verification
N/A
### Upgrade Verification
N/A

## Checklist

- [ ] Tested Installation
- [ ] Tested Uninstallation

- [ ] Yes
- [x ] No

### DIFF
--- a/roles/prerequisites/tasks/upgrade.yml
+++ b/roles/prerequisites/tasks/upgrade.yml
@@ -15,5 +15,5 @@
   when: target_version.stderr != ""
 
 - fail:
-    msg: "The target version must be in {{ from_versions }} to run this upgrade playbook, {{ target_version.stdout }} is currently installed"
+    msg: "The target version {{ target_version.stdout }} is already installed"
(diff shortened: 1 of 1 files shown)


trdoyle81:
@jessesarn merging to master to be cherrypicked to 1.7 ? Is that the correct process ?

jessesarn:
Yes, but noting the the real branch that you would cherry pick into would be named v1.7, not 1.7.

There is a branch named 1.7 that seems to contain stuff from back in 1.3 timeframe which nobody seems to know how it came to be.

JameelB:
/assign @pb82 

Can we get an approval here please? :smiley:

PR #1321, opened 2020-04-24
remove deploy/api-server resource override

jessesarn:
[INTLY-7154 ](https://issues.redhat.com/browse/INTLY-7154) involved several fixes to unblock testing of 1.7.0 rc testing.  Most of the fixes were specific to upgrading 1.6.1 to 1.7.0 and were merged to the v1.7 branch in [1314](https://github.com/integr8ly/installation/pull/1314).   However, this PR covers the one fix which also applies to master.

The api-server component was removed in AMQ 1.4 so its resource override is no longer necessary.


JameelB:
/lgtm

:cowboy_hat_face:

PR #1334, opened 2020-04-30
add jameelb to approvers

JameelB:
Requesting to get approver permissions

@jessesarn @ciaran-byrne can I get added as an approver just until rc2 is cut? This should help with getting PRs over the line without having to ping @jessesarn all the time :) 

I'll follow up with another PR to remove myself from the list once rc2 is cut.


jessesarn:
Totally agree

/lgtm
/approve

PR #1361, opened 2020-05-12
Adding automation for resetting the upgrade files on master after final release is cut

laurafitzgerald:
## Additional Information
https://issues.redhat.com/browse/INTLY-7207

Testing this is a little tricky.
The script automatically points to `origin`
You can update your local config to have `origin` point to your own fork.

**NOTE: Do not run the below commands unless you have change the fork as described above. It will create a release against origin. We don't want to do this during testing.**

### Case: Final Release
Make sure your local repo is clean
Run `scripts/release.sh -b v1.7 -r release-9.30.0` bump the version if you are testing a second time. 

Note, depending on when you forked you may need to use an older vX.Y. 

This creates a new release on your fork with the version provided

Screenshot 2020-05-12 at 11 14 29

This will also update the release branch _and_ create a new branch with the changes required for master

Screenshot 2020-05-12 at 11 14 14

On the new branch verify: 
1) that the upgrade playbook is reset to match the [template file ](https://github.com/integr8ly/installation/blob/master/scripts/upgrade.template.yml)
2) the vars file matches the [template file](https://github.com/integr8ly/installation/blob/master/scripts/upgrade_vars.template.yml) with the version that's passed. e.g. release-9.30.0

NOTE: We've already reset the upgrade playbook for master so you will only see one change in here for now removing a block. This block shouldn't be removed and it's added back in here but the script is using master to update so it will be seen as removed during verification.

Feel free to delete the generated branch after verifying the above.

### Case: rc release
Run 
Make sure your local repo is clean
Comment out lines 73-125
`scripts/release.sh -b v1.7 -r release-9.30.0-rc1`

A release will be created but no branch to reset master upgrade is pushed.

### Cleanup

On your fork, feel free to delete the generated tags and branches that were created by this verification.
Reset you remote configuration as you want it/had it before.

### DIFF
--- a/docs/release.md
+++ b/docs/release.md
@@ -44,30 +44,14 @@ To cut subsequent RCs for a patch release:
 
 `./scripts/release.sh -b v1.5 -r release-1.5.1-rc2`
 
+### Final release
 
-## Resetting the upgrade playbook
+When the script above is run for the final release e.g. (non rc release - release-x.y.z) the upgrade files will be reset. This was done manually but is now automated by the release script to reset these files on both the release branch and on master.
 
+Release branch - A commit is pushed to the release branch to reset the files
+Master branch - A new branch titled ${releaseTag}-master-upgrade-reset is pushed
 
-### Minor Release
-
-There may be logic in the upgrade playbook that is targetted at a specific release only.
-After the minor release branch is created, the upgrade playbook in `playbooks/upgrade.yml` should be reviewed and reset on `master` to remove any version specific blocks, tasks or roles being included.
-
-As this is a manual task, here are some guidelines for doing the review.
-
-* Any blocks that include a version specific upgrade task such as `upgrade_sso_72_to_73` can be removed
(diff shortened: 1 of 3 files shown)


JameelB:
@laurafitzgerald can you add this to the template as well? https://github.com/integr8ly/installation/blob/master/playbooks/upgrade.yml#L35-L40

laurafitzgerald:
> @laurafitzgerald can you add this to the template as well? https://github.com/integr8ly/installation/blob/master/playbooks/upgrade.yml#L35-L40

Yes sure. I also noted this was missing when I tested.

laurafitzgerald:
@JameelB I've also updated the verification steps to use a fork instead of commenting out the lines.

laurafitzgerald:
ping @trdoyle81 @pawelpaszki @jessesarn made some changes here to the code and verification steps. It's ready for review again. If we can get it merged we can use it during the final cut of the 1.7.0 release.

PR #1368, opened 2020-06-11
INTLY-8400: Removed JobRunningTimeExceeded (>300) and include in upgrades

pepedocs:
Jira: https://issues.redhat.com/browse/INTLY-8400
https://issues.redhat.com/browse/CSSRE-806
https://issues.redhat.com/browse/CSSRE-1004

Summary:
1. Removed JobRunningTimeExceeded with the 300s threshold
2. Attempt to update "upgrade.template.yml" to include these changes for the next patch relase. I will need guidance on this as I'm not quite sure how this should be done properly. Basically, we need backup monitoring will be included in the next release. CC @david-martin . 
3. The fix for CronJobsFailed was already merged 2 months ago but was not in the 1.7 release, we need to include this in the next patch release.

Please read the Jira for more accurate and updated information. 

P.S. I base off my PR from the 1.7 branch however just found out I can only make a PR to the master branch. TBH I'm not quite sure about the correct process for this particular repo.

### DIFF
--- a/docs/monitoring_docs.asciidoc
+++ b/docs/monitoring_docs.asciidoc
@@ -179,14 +179,6 @@ receivers:
   - send_resolved: true
     to: cssre-alerts@redhat.com
 - name: deadmansswitch
-inhibit_rules:
-- source_match:
-    alertname: 'JobRunningTimeExceeded'
-    severity: 'critical'
-  target_match:
-    alertname: 'JobRunningTimeExceeded'
-    severity: 'warning'
-  equal: ['alertname', 'job', 'label_cronjob_name']
 ```
 
 
--- a/roles/middleware_monitoring_config/templates/alertmanager.yml.j2
+++ b/roles/middleware_monitoring_config/templates/alertmanager.yml.j2
@@ -51,13 +51,5 @@ receivers:
   webhook_configs:
     - url: "{{ dms_webhook_url }}"
 {% endif %}
-inhibit_rules:
-- source_match:
-    alertname: 'JobRunningTimeExceeded'
-    severity: 'critical'
-  target_match:
-    alertname: 'JobRunningTimeExceeded'
-    severity: 'warning'
-  equal: ['alertname', 'job', 'label_cronjob_name']
 templates:
 - '*.tmpl'
--- a/playbooks/upgrade.yml
+++ b/playbooks/upgrade.yml
@@ -48,6 +48,11 @@
 #        tasks_from: upgrade_patch
 #      when: upgrade_webapp|bool
 
+  - name: Upgrade backup
+    include_role:
+      name: backup
+      tasks_from: upgrade
+
   - name: Upgrade fuse online
     include_role:
       name: fuse_managed
--- a/roles/backup/templates/backup-monitoring-alerts.yml.j2
+++ b/roles/backup/templates/backup-monitoring-alerts.yml.j2
@@ -9,16 +9,14 @@ spec:
   groups:
     - name: general.rules
       rules:
-      {% for check in [{ 'time': '300', 'severity': 'warning' }, { 'time': '600', 'severity': 'critical' }] %}
       - alert: JobRunningTimeExceeded
         annotations:
           sop_url: https://github.com/RHCloudServices/integreatly-help/blob/master/sops/alerts_and_troubleshooting.md
-          message: Job {{ '{{' }} $labels.namespace {{ '}}' }}/{{ '{{' }} $labels.job {{ '}}' }} has been running for longer than {{ check['time'] }} seconds
+          message: Job {{ '{{' }} $labels.namespace {{ '}}' }}/{{ '{{' }} $labels.job {{ '}}' }} has been running for longer than 600 seconds
         expr: |
-          time() - (max(kube_job_status_active * ON(job) GROUP_RIGHT() kube_job_labels{label_monitoring_key="middleware"}) BY (job) * ON(job) GROUP_RIGHT() max(kube_job_status_start_time * ON(job) GROUP_RIGHT() kube_job_labels{label_monitoring_key="middleware"}) BY (job, namespace, label_cronjob_name) > 0) > {{ check['time'] }}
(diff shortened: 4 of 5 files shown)


pepedocs:
CC @alanmoran @david-martin

austincunningham:
Thanks for your contribution @pepedocs 
Can you provide verification step to confirm your pr does what expected.

The pr seems to be removing the warning level alert for JobRunningTimeExceeded so this should be removed from elsewhere in the installer see https://github.com/integr8ly/installation/search?q=JobRunningTimeExceeded&unscoped_q=JobRunningTimeExceeded

pepedocs:
> Thanks for your contribution @pepedocs
> Can you provide verification steps to confirm your pr does what expected.
> 
> The pr seems to be removing the warning level alert for JobRunningTimeExceeded so this should be removed from elsewhere in the installer see https://github.com/integr8ly/installation/search?q=JobRunningTimeExceeded&unscoped_q=JobRunningTimeExceeded

@austincunningham I've just removed the references to the JobRunningTimeExceeded warning alert.

In regards to the verification, this is really difficult to test. It requires a particular scenario where some metric data are delayed due to possibly prometheus app delays or federation of metric data. Please see the ticket here which is best explained - https://issues.redhat.com/browse/CSSRE-1004

Can the absence of this warning alert in the alertmanager/prometheus dashboard considered verification already? We just need to remove this warning alert as besides it's not accurate the 300s threshold is also not reasonable.

jhellar:
For the verification, IMO it is enough to check that the alert is gone. This is how it will be also tested for 1.7.1 - https://gitlab.cee.redhat.com/integreatly-qe/integreatly-test-cases/-/blob/1.7.1/tests/alerts/c06-verify-that-there-is-only-one-alert-for-job-timeout.md

docs/monitoring_docs.asciidoc
```diff
@@ -179,14 +179,6 @@ receivers:
   - send_resolved: true
     to: cssre-alerts@redhat.com
 - name: deadmansswitch
-inhibit_rules:
```
austincunningham, line 182: only need to remove the target_match for warning
``` yaml
inhibit_rules:
- source_match:
    alertname: 'JobRunningTimeExceeded'
    severity: 'critical'
  equal: ['alertname', 'job', 'label_cronjob_name']
```

roles/middleware_monitoring_config/templates/alertmanager.yml.j2
```diff
@@ -51,13 +51,5 @@ receivers:
   webhook_configs:
     - url: "{{ dms_webhook_url }}"
 {% endif %}
-inhibit_rules:
```
austincunningham, line 54: same here only need to remove the target_match for warning
``` yaml
inhibit_rules:
- source_match:
    alertname: 'JobRunningTimeExceeded'
    severity: 'critical'
  equal: ['alertname', 'job', 'label_cronjob_name']
```

docs/monitoring_docs.asciidoc
pepedocs: It looks like the source match will be useless without the target match, that's why I removed it. Please let me know if Im missing something here.

pepedocs:
@austincunningham Could yoou also please verify if I made the right changes so that these alerts will be applied in the next upgrade?

austincunningham:
@pepedocs deployed this to a cluster this morning and no `JobRunningTimeExceeded` alerts were showing up in Prometheus, going to deploy your latest changes to see if that resolves the issue

pepedocs:
> @pepedocs deployed this to a cluster this morning and no `JobRunningTimeExceeded` alerts were showing up in Prometheus, going to deploy your latest changes to see if that resolves the issue

Thanks for testing it. I'm quite sure why the changes would cause all the Job* alerts to disappear. Please let me know when you have updates. If it fails your test, I will investigate on this tomorrow and test it on a live cluster.

austincunningham:
@pepedocs second time was the same, decided to see if it was something I was doing wrong so tried master and it was the same , I was missing an install flag `-e backup_restore_install=true` , will test again with the updated flag.

pepedocs:
> @pepedocs second time was the same, decided to see if it was something I was doing wrong so tried master and it was the same , I was missing an install flag `-e backup_restore_install=true` , will test again with the updated flag.

By the way could you also please check/test if the alert `CronJobsFailed` is using the metric `kube_job_failed` instead of `kube_job_status_failed`? This is part of the fix.

pepedocs:
@austincunningham Thanks for testing this out again. Could someone in your team please provide guidance on this? The changes are pretty straightforward, I'm almost out of ideas why it removes the cron job alert. The behavior expected for these changes must be;

1. JobRunningTimeExceeded must exist but only for the 600s threshold.
2. CronJobsFailed alert must exist but should use the `kube_job_failed` prom metric.

CC @david-martin

pepedocs:
@austincunningham Looking at the install logs you've provided, it seems like it didn't perform the upgrades for the backup monitoring. Particularly, it doesn't have the task `Delete existing alerts` (https://github.com/integr8ly/installation/blob/master/roles/backup/tasks/upgrade.yaml)

Therefore, can someone from your team please provide the steps on how/what files to update so that changes such as these backup monitoring alerts will be applied to upgrades and not just installs? This is not documented AFAICT in this repository.

pepedocs:
@austincunningham I'm still not sure how to include these changes to integreatly upgrades (see my previous comments). However, I've just made another attempt (changes) and I've simply moved the backup monitoring upgrade role to the upgrade.yml playbook instead of the template.

**Could you please perform a test if you think this will fix the issue?** Reason why I attemtpted this is for us to save time waiting for the next day and most importantly because I also saw a PR that was modified the said playbook for including some changes in the next upgrades, so I'm really just inferring here.

austincunningham:
/approve 
/lgtm
### Logs
https://issues.redhat.com/secure/attachment/12475971/12475971_integreatly-upgrade.log
https://issues.redhat.com/secure/attachment/12475972/12475972_integreatly-install.log

/assign @jessesarn

PR #1369, opened 2020-06-21
Add custom SOP URLs to enmasse alert mails; Fix alertmanager slack vars

grdryn:
JIRA links:
https://issues.redhat.com/browse/INTLY-8385
https://issues.redhat.com/browse/INTLY-8386

This change adds custom sop_url info to 5 alerts. It also ensures that
2 of those alerts get routed to the critical receiver in AlertManager,
even though the severity label is set to warning.

See the first alert in the following example (the second alert is unrelated, but useful to see that the sop_url annotation is the same format):

Screenshot from 2020-06-21 17-40-47

There's also a fix for an issue with how the slack properties are
configured, which meant that sometimes install/upgrade would fail when
you didn't specify those vars, but did specify other unrelated vars
for building the main alertmanager config.

## Verification Steps

### Installation Verification

Install log: https://gist.github.com/grdryn/e56350f9ad74884f1f28d14252c9bdce/raw/8a1f8a0be58eee00fb6e3ddf8b87aacf8f855de2/integreatly-install.log

### Upgrade Verification

Upgrade log: https://gist.github.com/grdryn/e56350f9ad74884f1f28d14252c9bdce/raw/8a1f8a0be58eee00fb6e3ddf8b87aacf8f855de2/integreatly-upgrade.log

## Checklist

- [x] Tested Installation
- [x] Tested Uninstallation

- [ ] Yes
- [ ] No
- [x] i added an upgrade task to the upgrade playbook, and I added an upgrade log from a successful run

## NOTE
* I don't actually know what needs to go into the SOPs for these alerts, but so I've just created a skeleton of each of the files referenced in this PR: https://github.com/RHCloudServices/integreatly-help/pull/564
* I don't know how to trigger all of these alerts, so I haven't been able to verify that the two warning severity alerts are correctly being routed to the critical receiver now
* These changes, once accepted, will need to be cherry-picked to the 1.7 branch

### DIFF
--- a/roles/middleware_monitoring_config/defaults/main.yml
+++ b/roles/middleware_monitoring_config/defaults/main.yml
@@ -55,8 +55,8 @@ alertmanager_slack_default_channel: ''
 
 # Custom SOP URLs for alerts
 sops:
-  BrokerMemory: https://github.com/integr8ly/integreatly-help/blob/master/sops/AMQOnline_BrokerMemory.asciidoc
-  ComponentHealth: https://github.com/integr8ly/integreatly-help/blob/master/sops/AMQOnline_ComponentHealth.asciidoc
-  AuthenticationService: https://github.com/integr8ly/integreatly-help/blob/master/sops/AMQOnline_AuthenticationService.asciidoc
-  RouterMeshUndeliveredHealth: https://github.com/integr8ly/integreatly-help/blob/master/sops/AMQOnline_RouterMeshUndeliveredHealth.asciidoc
-  RouterMeshConnectivityHealth: https://github.com/integr8ly/integreatly-help/blob/master/sops/AMQOnline_RouterMeshConnectivityHealth.asciidoc
+  BrokerMemory: https://github.com/integr8ly/integreatly-help/blob/master/sops/alerts/AMQOnline_BrokerMemory.asciidoc
+  ComponentHealth: https://github.com/integr8ly/integreatly-help/blob/master/sops/alerts/AMQOnline_ComponentHealth.asciidoc
+  AuthenticationService: https://github.com/integr8ly/integreatly-help/blob/master/sops/alerts/AMQOnline_AuthenticationService.asciidoc
+  RouterMeshUndeliveredHealth: https://github.com/integr8ly/integreatly-help/blob/master/sops/alerts/AMQOnline_RouterMeshUndeliveredHealth.asciidoc
+  RouterMeshConnectivityHealth: https://github.com/integr8ly/integreatly-help/blob/master/sops/alerts/AMQOnline_RouterMeshConnectivityHealth.asciidoc
--- a/roles/middleware_monitoring_config/defaults/main.yml
+++ b/roles/middleware_monitoring_config/defaults/main.yml
@@ -53,3 +53,10 @@ pd_service_key: "{{ pagerduty_integration_key | default('') }}"
 alertmanager_slack_api_url: "{{ alertmanager_slack_api_url | default('') }}"
 alertmanager_slack_default_channel: "{{ alertmanager_slack_default_channel | default('') }}"
 
+# Custom SOP URLs for alerts
+sops:
+  BrokerMemory: https://github.com/integr8ly/integreatly-help/blob/master/sops/AMQOnline_BrokerMemory.asciidoc
+  ComponentHealth: https://github.com/integr8ly/integreatly-help/blob/master/sops/AMQOnline_ComponentHealth.asciidoc
+  AuthenticationService: https://github.com/integr8ly/integreatly-help/blob/master/sops/AMQOnline_AuthenticationService.asciidoc
+  RouterMeshUndeliveredHealth: https://github.com/integr8ly/integreatly-help/blob/master/sops/AMQOnline_RouterMeshUndeliveredHealth.asciidoc
+  RouterMeshConnectivityHealth: https://github.com/integr8ly/integreatly-help/blob/master/sops/AMQOnline_RouterMeshConnectivityHealth.asciidoc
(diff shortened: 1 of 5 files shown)


grdryn:
/cc @jessesarn @david-martin @alanmoran @ciaran-byrne 

Please add others you feel should review.

alanmoran:
@grdryn sorry for the delay in reviewing. Shouldn't these SOPs live in /alerts (https://github.com/RHCloudServices/integreatly-help/tree/master/sops/alerts)

roles/middleware_monitoring_config/defaults/main.yml
```diff
@@ -50,6 +50,13 @@ dms_webhook_url: "{{ deadmanssnitch_url | default('') }}"
 pd_service_key: "{{ pagerduty_integration_key | default('') }}"
 
 # slack
-alertmanager_slack_api_url: "{{ alertmanager_slack_api_url | default('') }}"
-alertmanager_slack_default_channel: "{{ alertmanager_slack_default_channel | default('') }}"
-
+alertmanager_slack_api_url: ''
```
alanmoran, line 53: @pepedocs Can you have a look at this? I'm fairly sure CS SRE inject an `alertmanager_slack_api_url ` variable during these runs?

roles/middleware_monitoring_config/templates/alertmanager.yml.j2
```diff
@@ -35,7 +38,7 @@ receivers:
     - send_resolved: true
     - api_url: {{ alertmanager_slack_api_url }}
       channel: {{ alertmanager_slack_default_channel }}
-      text: "message: {{ .CommonAnnotations.message }}"
+      text: "message: {{ '{{' }} .CommonAnnotations.message {{ '}}' }}"
```
david-martin, line 41: @grdryn Is this escaping correct?
Is it the first pass of templating that has the `.CommonAnnotations.message` var,
or at runtime?

grdryn:
> @grdryn sorry for the delay in reviewing. Shouldn't these SOPs live in /alerts (https://github.com/RHCloudServices/integreatly-help/tree/master/sops/alerts)

Yes, fixed now (and the files updated in the corresponding PR to integreatly-help

roles/middleware_monitoring_config/templates/alertmanager.yml.j2
grdryn: It's correct now, it wasn't before. It's processed by the AlertManager go-templating at runtime rather than Ansible's jinja2 templating. See [this thread](https://chat.google.com/room/AAAApI2jS20/H-tfsJ7ArT0) for more info.

jessesarn: Yes, this is correct.. it is supposed to be evaluated at runtime.  We worked on this in the integreatly-support channel last week.

roles/middleware_monitoring_config/defaults/main.yml
grdryn: See [this thread](https://chat.google.com/room/AAAApI2jS20/H-tfsJ7ArT0) for more info. Whether CSSRE do or don't pass it shouldn't matter here. This change fixes the fact that when I _don't_ pass it, but _do_ pass an unrelated `sendgrid_api_key` variable, then upgrade fails. If CSSRE pass a value for this `alertmanager_slack_api_url`, then it will override this empty default.

pepedocs: @alanmoran Yes that variable is suppose to allow CSSRE (or any users) to inject an api for alertmanager slack integration.

@grdryn If ansible merge behavior is set to override vars this is ok, otherwise this will not work. I'm talking about this - https://docs.ansible.com/ansible/2.3/intro_configuration.html#hash-behaviour

grdryn: @pepedocs my understanding of that section of the docs is that it's only referring to hashes (dictionaries), and how ansible should merge them (or not); where these `alertmanager_slack_*` variables are mere strings, rather than hashes.

I believe this is a more relevant section of the docs: https://docs.ansible.com/ansible/2.3/playbooks_variables.html#variable-precedence-where-should-i-put-a-variable

As you can see there, variables defined at this level (role defaults) have the lowest possible precedence, so if you set values for them anywhere else, that will override the empty values here.

grdryn: As additional context, you'll notice that other variables here follow the other pattern that I'm changing for these vars, and I'm not changing it for those other ones. The difference, as I see it, is that those other examples are setting a variable default of "some _other_ variable, if it exists, or an empty string" -- the important part being that they're referencing a different variable, rather than this one.

E.g.:

```yaml
pd_service_key: "{{ pagerduty_integration_key | default('') }}"
```

rather than:
```yaml
pd_service_key: "{{ pd_service_key | default('') }}"
```

grdryn:
/hold

I've got one more thing to add here

roles/middleware_monitoring_config/defaults/main.yml
pepedocs: @grdryn That's right it's just a mere string. Sorry wrong docu. Generally speaking I was saying we just ensure we are expecting the right precedence, thus -e will overwrite things. So there should be no problem here if this is the case.

grdryn:
@jessesarn @david-martin this is ready for review again now. There are two changes since last review:

* Ensure routing to critical receiver for all 5 alerts (the one that I'm able to actually trigger, ComponentHealth, didn't seem to be getting routed there by default for some reason
* Ensure that SOP URL is available in pagerduty. It's in a slightly different place than before:

12475915_image-2020-06-23-19-08-40-823

jessesarn:
Looks good from my end, @david-martin said he was going to review as well.   Deferring to him for the official LGTM.

david-martin:
/lgtm
Thanks @grdryn

PR #1372, opened 2020-06-26
intly-8601 add dummy/null receiver for UnifiedPushJavaHeapThresholdExceeded alert

ciaranRoche:
## Additional Information

JIRA -> https://issues.redhat.com/browse/INTLY-8601

Based on a discussion in INTLY-8601 it was decided to remove the alert to reduce SRE toil. Until the alert has been removed from the Unified Operator https://github.com/aerogear/unifiedpush-operator/pull/100, this PR adds a null receiver for the UnifiedPushJavaHeapThresholdExceeded to stop it from firing.

## Verification Steps

### Installation Verification
### Upgrade Verification

## Checklist

- [ ] Tested Installation
- [ ] Tested Uninstallation

- [ ] Yes
- [ ] No

### DIFF
--- a/roles/middleware_monitoring_config/templates/alertmanager.yml.j2
+++ b/roles/middleware_monitoring_config/templates/alertmanager.yml.j2
@@ -15,6 +15,9 @@ route:
   repeat_interval: 12h
   receiver: default
   routes:
+  - receiver: "null"
+    match:
+      alertname: UnifiedPushJavaHeapThresholdExceeded
   - match_re:
       alertname: ^RouterMeshConnectivityHealth$|^RouterMeshUndeliveredHealth$|^ComponentHealth$|^BrokerMemory$|^AuthenticationService$
     receiver: critical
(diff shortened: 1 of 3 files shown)


roles/middleware_monitoring_config/templates/alertmanager.yml.j2
```diff
@@ -24,6 +24,9 @@ route:
 {% if dms_webhook_url %}
     receiver: deadmansswitch
 {% endif %}
+  - receiver: "null"
+    match:
+      alertname: UnifiedPushJavaHeapThresholdExceeded
```
david-martin, line 29: I think you also need to define an empty receiver in the receivers block with the same name.

e.g.

```
receivers:
- name: null
```

ciaranRoche:
### Install
**alertmanager-application-monitioring secret**
image
**alertmanager config**
image
image
image
**reduced alert firing**
image
**null receiver catching alert**
image

PR #1370, opened 2020-06-25
INTLY-8459 Allow for metric returning a 0 value

david-martin:
## Additional Information
https://issues.redhat.com/browse/INTLY-8459

Note, the upgrade playbook already has logic to update alerts.
This change is intended for 1.7.1, but will get this landed on master first

## Verification Steps
As the verifier of the PR the following process should be done:

- After installation, scale down the 3scale system-app pod(s)
- Verify the ThreeScaleDeveloperUIBBT, ThreeScaleSystemAdminUIBBT &  ThreeScaleAdminUIBBT alerts goes to pending (and eventually fire)

### Installation Verification
- Ensure the author of the PR has attached a log of the installation run from his branch to the jira or pr and check that it exited as expected.
- Verify the fresh installation is correct on cluster provided by PR author 
### Upgrade Verification
- After installation verification, notify the PR author to begin an upgrade on their cluster
- Ensure the developer of the PR has attached a log of the upgrade run from his branch to the jira or pr and check that it exited as expected. 
- If possible, look at the tasks that ran and see they match the PR
- Verify the upgrade is correct on cluster provided by PR author 

## Checklist

- [ ] Updated [Changelog](https://github.com/integr8ly/installation/blob/master/CHANGELOG.md)
- [ ] Tested Installation
- [ ] Tested Uninstallation
- [ ] Tested or Created follow on for Upgrade

Is an upgrade task required and are there additional steps needed to test this?
- [x] Yes
- [ ] No

PR #1374, opened 2020-06-26
Revert "Reset upgrade variables after final release release-1.7.0"

damienomurchu:
This reverts commit a7d6bc7b741a5e5ba19e463b754eed029f80a4cb.

## Additional Information
https://issues.redhat.com/browse/INTLY-8978
Upgrade logic removed after 1.7.0 now needs to be added back to upgrade to 1.7.1. 

## Verification Steps
As the verifier of the PR the following process should be done:

### Installation Verification
N/A, only relevant to upgrade.

### Upgrade Verification

- Provision RHMI on cluster
- Perform upgrade using a branch that includes the upgrade logic from this PR, and product updates to any of the products specified [here](https://github.com/integr8ly/installation/pull/1374/files#diff-a26bad16e0c47fd2a91b905d91c7f58bR2-R13)
- verify that upgrade completes successfully and intended product(s) get updated.

** a branch that can be used to verify such an upgrade is included in https://issues.redhat.com/browse/INTLY-8978

## Checklist

- [ ] Updated [Changelog](https://github.com/integr8ly/installation/blob/master/CHANGELOG.md)
- [ ] Tested Installation
- [ ] Tested Uninstallation
- [x] Tested or Created follow on for Upgrade

Is an upgrade task required and are there additional steps needed to test this?
- [x] Yes
- [ ] No

PR #1375, opened 2020-06-26
Intly 8459 probe success alerts fix cherry pick

david-martin:
## Additional Information
Cherrypick of #1370 
https://issues.redhat.com/browse/INTLY-8459

## Verification Steps

* After installation, scale down the 3scale system-app pod(s)
* Verify the ThreeScaleDeveloperUIBBT, ThreeScaleSystemAdminUIBBT & ThreeScaleAdminUIBBT alerts goes to pending (and eventually fire)

## Checklist

- [X] Updated [Changelog](https://github.com/integr8ly/installation/blob/master/CHANGELOG.md)
- [ ] Tested Installation
- [ ] Tested Uninstallation

- [X] Yes
- [ ] No

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,8 @@ Some of these changes may include:
 * [INTLY-3847] - Update Alert Manager emails to include cluster URL and timestamps
 * [INTLY-5856] - Improve resiliency of sso/user-sso w/ 2nd replica and qos of postgres pods up from BestEffort to Burstable
 * [INTLY-2544] - allow customer-admins view 3Scale logs in Kibana
-
+* [INTLY-8459] - Fix 3Scale probe alerts
+ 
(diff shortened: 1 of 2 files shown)


playbooks/upgrade.yml
```diff
@@ -21,6 +21,7 @@
     vars:
       from_versions:
       - "{{ upgrade_from_version }}"
+      - "release-1.6.1"
```
grdryn, line 24: Do you think maybe this should be added as part of #1374 rather than as part of this change? I'm just thinking from a "remember to remove this after release" point of view.

david-martin: yes, lets do that
cc @damienomurchu

ciaranRoche:
/lgtm
verified from fresh install
image
image

PR #1377, opened 2020-06-30
cherry pick to v1.7 adding dummy/null reciever for UnifiedPushJavaHeapThresholdExceeded alert

ciaranRoche:
## Additional Information

cherrypick of https://github.com/integr8ly/installation/pull/1372

This adds a null receiver for alert `UnifiedPushJavaHeapThresholdExceeded` as discussed in -> https://issues.redhat.com/browse/INTLY-8601

## Verification Steps
During installation and upgrades this change requires the following params to be added : 
```
  -e smtp_auth_password=dummy \
  -e dms_webhook_url=https://example.com \
  -e pd_service_key=dummy \
```
This is to ensure the alertmanger config role gets executed. With out these params added the custom alertmanager config is not created and a default config is used.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,8 +20,10 @@ Some of these changes may include:
 * [INTLY-5856] - Improve resiliency of sso/user-sso w/ 2nd replica and qos of postgres pods up from BestEffort to Burstable
 * [INTLY-2544] - allow customer-admins view 3Scale logs in Kibana
 * [INTLY-8459] - Fix 3Scale probe alerts
- 
+* [INTLY-8601] - Add dummy/null reciever for UnifiedPushJavaHeapThresholdExceeded alert
+
 ### Removed
+
 
 ### Bug Fixes
 * [INTLY-8385] - Added SOPs to 5 new alerts in 1.7.0
--- a/roles/middleware_monitoring_config/tasks/create_alertmanager.yml
+++ b/roles/middleware_monitoring_config/tasks/create_alertmanager.yml
@@ -1,5 +1,4 @@
 ---
-
 - name: Configure AlertManager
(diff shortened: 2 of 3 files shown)


pepedocs:
@ciaranRoche The silence will work. However why not just completely remove the alert from here;

https://github.com/integr8ly/installation/blob/a8b28fa86b35e5d5463af0f8a5e64a403940fb23/roles/ups/templates/prometheus_rule.yml.j2#L40

, and add something here for UPS for upgrades

https://github.com/integr8ly/installation/blob/master/playbooks/upgrade.yml

david-martin:
> However why not just completely remove the alert from here

@pepedocs My understanding is that file is no longer used. The rules are being created by the unifiedpush operator.

@ciaranRoche it might be best to remove that file to avoid confusion & verify it's definitely not used anywhere. (Happy for that to be a separate PR to master)

pepedocs:
@david-martin In that case this looks good to me. 

Out of curiosity also, are there plans for the operator to dynamically grab new config/code so that it can be more modular with respect to upgrades? For ex. we just need to update the operator code/config, operator grabs them from a production branch and applies them.

/lgtm

david-martin:
> Out of curiosity also, are there plans for the operator to dynamically grab new config/code so that it can be more modular with respect to upgrades? For ex. we just need to update the operator code/config, operator grabs them from a production branch and applies them.

@pepedocs The short answer is no, at least not for 1.x.
I'd like to understand this ask in more detail.
Also, keep in mind any sizeable change is more than likely only going to be done for 2.x.
I think it's important to balance determinism of a release, ability to react to actual production situations in a specific cluster, and management of an entire fleet of clusters.

PR #1384, opened 2020-07-13
fix: Update DataSync walkthough

wtrocki:
## Additional Information
 
Introducing changes from https://github.com/aerogear/mobile-walkthrough/pull/38 to remove keycloak instructions from the walkthrough that disables instructions for keycloak.

Keycloak setup in Walkthrough do not make much value as we do not provide any authz rules (as opposed to the new walkthrough that gives developers full integration and out of the box auth rules)

> NOTE: This is change is completely independent of the keycloak update and it can be merged at any point.

## Verification Steps

On any integrate cluster simply change tag of the mobile-wt to 0.8.0 and see if walkthrough is rendering. You should no longer see keycloak/auth sections


wtrocki:
CC @maleck13

wtrocki:
This bot is a /joke

PR #1389, opened 2020-07-24
ensure /tmp/codeready-1.2.0.GA is gone

tonyxrmdavidson:
small fix to ensure the ensure /tmp/codeready-1.2.0.GA is removed prior to downloading the codeready installer


damienomurchu:
Closing as cp'ed into 1.6.1-rc2-fuse-fix instead

PR #1351, opened 2020-05-05
INTLY-6512

laurafitzgerald:
## Additional Information
https://issues.redhat.com/browse/INTLY-6512

### Installation Verification
Case: Install with all products true. v1.8 (aka this branch)
Install command
```
ansible-playbook -i inventories/pds.template \
>   playbooks/install.yml \
>   -e eval_self_signed_certs=false \
>   -e rhsso_identity_provider_ca_cert_path= \
>   -e amq_streams=true \
>   -e rhsso_seed_users_password=<redacted> \
>   -e rhsso_evals_admin_password=<redacted> \
>   -e rhsso_cluster_admin_password=<redacted> \
>   -e gitea=false \
>   -e heimdall_pull_secret_token=<redacted>\
>  | tee -a /tmp/integreatly-install-`date +%h%m%s%d%b%Y`.log 
```
Install [Logs](https://gist.github.com/laurafitzgerald/ac90b019e426de37e221874d50cdae0a#file-heimdall-fresh-install-L278-L304)
Heimdall installed
Screenshot 2020-05-12 at 15 39 08

And crs created.
Screenshot 2020-05-12 at 16 38 37

#### Case 1: All products true
Install release-1.7.0-rc2
Run 1.8.0 (aka this branch) upgrade - [logs](https://gist.github.com/laurafitzgerald/4563055655deaa1f0109e7370e450def#file-intly-6512-upgrade-L229-L308)
Verify that heimdall is installed and all crs are created
Screenshot 2020-05-12 at 12 56 24
Verify that the related crs are created.
Screenshot 2020-05-12 at 12 57 19

#### Case 2: Some products false
Verify that heimdall is installed and cases where product is false there's no attempt to create the crs.
Ordered a 1.6.0 pds. - ran an upgrade from this branch.
Upgrade [logs](https://gist.github.com/laurafitzgerald/89e116fc5f30a1b34c2462078be34db9#file-intly-6512-upgrade-apicurito-threescale-false-L171-L237)
`-e threescale=false` and `-e apicurito=false` were passed. Verify there were two skipped creation of Image Monitor items in the linked logs above. It failed but that is because the installation included them. Just wanting to verify here that if a cluster doesn't have a specific product installed then the upgrade wont try to create an image monitor for it. It's very much an edge case.

## Checklist
- [x] Updated [Changelog](https://github.com/integr8ly/installation/blob/master/CHANGELOG.md)
- [x] Tested Installation
- [x] Tested Uninstallation
[...]

PR #1397, opened 2020-08-06
INLTY-9733 - Revert Heimdall installation

KevFan:
This reverts commit bc63b87342a4533e2caf8e05cbe9394edbb18a65, reversing changes made to 621e32833769da66c88abb1cab7ccdfa070086d2.

Jira:
* https://issues.redhat.com/browse/INTLY-9733

## Verification Steps

### Installation Verification
### Upgrade Verification

## Checklist

- [x] Tested Installation
- [ ] Tested Uninstallation

- [ ] Yes
- [ ] No

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -119,7 +119,7 @@ Before running the installer, please consider the following variables:
 | github_client_id | GitHub OAuth client ID to enable GitHub authorization for Launcher. If not defined, GitHub authorization for Launcher will be disabled
 | github_client_secret | GitHub OAuth client secret to enable GitHub authorization for Launcher. If not defined, GitHub authorization for Launcher will be disabled
 | prerequisites_install | Boolean var that skips the installation of system wide tools/packages that are required by the installer if set to false (needs to be set to false when running the installer in a linux container). Defaults to `true`. 
-| heimdall_pull_secret_token | Heimdall pull secret token - See section below for where to access the token - alternatively disable heimdall by using heimdall=false
+| heimdall_pull_secret_token | Heimdall pull secret token
 |===
 
 Some products can be excluded from the install by setting a var. For example, setting `gitea=false` will not install gitea. The flags for installing specific products can be found in the link:./inventories/group_vars/all/manifest.yaml[manifest.yaml] file
@@ -155,11 +155,9 @@ To enable Heimdall you will need a valid pull secret token to access the registr
 
 Click on `view its contents`. Use the entry at data.dockerconfigjson to replace `<your_secret_token>` below.
 
-Alternatively you can disable the installation of heimdall by passing `-e heimdall=false`
-
 [source,shell]
 ----
--e heimdall_pull_secret_token=<your_secret_token>
+-e heimdall_pull_secret_token=<your_secret-token>
 ----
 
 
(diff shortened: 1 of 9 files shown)


KevFan:
Upgrade was reverted already by a reset after 1.7.1 release - https://github.com/integr8ly/installation/commit/59bc69472808c67c926729d106ae9930d6e69caa#diff-87051ab24211e0e970f9fb3f11a1b1fb

roles/3scale/tasks/heimdall.yml
```diff
@@ -7,6 +7,4 @@
   vars:
     imagemonitor_cr_namespace: "{{ threescale_namespace }}"
     excludePattern: "{{ threescale_heimdall_exclude_pattern }}"
-  when:
-    - heimdall | default(true) | bool
-    - threescale | default(true) | bool
```
philbrookes, line 12: This is confusing me, why is this only when heimdall is true? The original seems more sensible? Same for the similar changes below.

KevFan: I think it's because we only want to create the `ImageMonitor` resource to the product namespaces only when heimdall is installed. Heimdall would watch for namespaces with `ImageMonitor` CR and check the deployments etc for CVE updates :thinking:

README.adoc
```diff
@@ -153,13 +154,9 @@ The following flag can be used if self signed certs are used.
...
 To enable Heimdall you will need a valid pull secret token to access the registry, you can retrieve these from https://access.redhat.com/terms-based-registry/#/token/-heimdall/openshift-secret
 
-Click on `view its contents`. Use the entry at data.dockerconfigjson to replace `<your_secret_token>` below.
-
-Alternatively you can disable the installation of heimdall by passing `-e heimdall=false`
-
 [source,shell]
 ----
--e heimdall_pull_secret_token=<your_secret_token>
+-e heimdall_pull_secret_name=<your_secret-name> -e heimdall_pull_secret_token=<your_secret-token>
```
laurafitzgerald, line 159: Can we keep this change. It is a valid change whether heimdall is installed by default or not.

roles/heimdall/templates/heimdall-dockercfg.yml.j2
```diff
@@ -2,6 +2,7 @@ apiVersion: v1
 kind: Secret
 metadata:
   name: heimdall-dockercfg
+  username: {{ heimdall_pull_secret_name }}
```
laurafitzgerald, line 5: Can we keep this removed. Same as above.

README.adoc
laurafitzgerald: The reasoning is explained in the commit message here: https://github.com/integr8ly/installation/pull/1351/commits/31be9ff5dc917c3c36519d08098c7fde573d6918

laurafitzgerald:
@KevFan will there be an upgrade required here? i.e. 1.7.x -> 1.8 would have installed heimdall. Should 1.8 -> 1.x.x have an upgrade step to delete heimdall and it's related Image Monitors?

README.adoc
```diff
@@ -153,13 +153,9 @@ The following flag can be used if self signed certs are used.
 
 To enable Heimdall you will need a valid pull secret token to access the registry, you can retrieve these from https://access.redhat.com/terms-based-registry/#/token/-heimdall/openshift-secret
 
-Click on `view its contents`. Use the entry at data.dockerconfigjson to replace `<your_secret_token>` below.
```
laurafitzgerald, line 156: Can we keep this line in as well. Sorry missed it last time.

laurafitzgerald:
@KevFan As I'm not on sheriff I'll leave it with yee folks to verify/approve. But it looks fine to me if it works. Thanks for making the changes I suggested. 👍

PR #1419, opened 2020-09-08
Release 1.9.0 rc1 master upgrade reset

mikenairn:
## Additional Information
Reset upgrade playbook and CHANGELOG after 1.9.0 release cut.

## Checklist

- [ ] Tested Installation
- [ ] Tested Uninstallation

- [ ] Yes
- [ ] No


mikenairn:
@jessesarn @R-Lawton Can someone give this an lgtm please.

Note, the release script is wrong and would normally create this branch after the final release, but it should actually be getting created now when the release branch is created from master.

CHANGELOG is also not taking into account in the release script.

Might be worth a follow on task to look into fixing that release script, but that really depends i guess how many more of these we are likely to be doing.

jessesarn (approved):
/lgtm

PR #1423, opened 2020-09-10
[wip] lowering UnifiedPushJavaNonHeapThresholdExceeded receiver to null

R-Lawton:
## Verification Steps

### Installation Verification
### Upgrade Verification

## Checklist

- [ ] Tested Installation
- [ ] Tested Uninstallation

- [ ] Yes
- [ ] No

### DIFF
--- a/roles/middleware_monitoring_config/templates/alertmanager.yml.j2
+++ b/roles/middleware_monitoring_config/templates/alertmanager.yml.j2
@@ -18,6 +18,9 @@ route:
   - receiver: "null"
     match:
       alertname: UnifiedPushJavaHeapThresholdExceeded
+  - receiver: "null"
+    match:
+      alertname: UnifiedPushJavaNonHeapThresholdExceeded
   - match_re:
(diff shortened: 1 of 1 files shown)


roles/middleware_monitoring_config/templates/alertmanager.yml.j2
```diff
@@ -18,6 +18,9 @@ route:
   - receiver: "null"
     match:
       alertname: UnifiedPushJavaHeapThresholdExceeded
+  - receiver: "null"
+    match:
+      alertname: UnifiedPushJavaNonHeapThresholdExceeded
```
mikenairn, line 23: @david-martin Is adding an additional entry for the "null" receiver rather than a single one using a regexp a valid configuration option? Just checking, we will be testing it anyway.

mikenairn:
@R-Lawton Can you apply and PR this against v1.9

R-Lawton:
closing this to make a new Pr to push these changes to 1.9 to make it easier for testing

roles/middleware_monitoring_config/templates/alertmanager.yml.j2
david-martin: @mikenairn Yes, this should be valid.
Each entry is a `route` with a separate matcher & receiver combo

PR #1435, opened 2020-12-01
Add new alert target for the Openshift Prometheus Federation.

Boomatang:
## Additional Information

A new warning TargetDown alert specifically for the openshift-monitoring-federation scrape target needs to be created so we are notified if federation is failing.

**JIRA** https://issues.redhat.com/browse/INTLY-10161

## Verification Steps
- install RHMI as normal
- in the openshift-monitorig edit the stateful-set for prometheus-k8s.
 - reduce the number of replicas to zero.
- go to the middleware-monitoring prometheus instance https://prometheus-route-middleware-monitoring.apps.<CLUSTER URL>/alerts
- find the alert `OpenshiftPrometheusFederationTargetDown`.
  - The alert should be in a pending state
  - After 10 minutes the alert should be firing
- Check the `TargetDown` alert
  - This should **not** be alerting
- During the uninstall this alert should not be firing. 

### Installation Verification
### Upgrade Verification

## Checklist

- [ ] Tested Installation
- [ ] Tested Uninstallation

- [ ] Yes
- [x] No

### DIFF
--- a/roles/middleware_monitoring_config/templates/target_down_alerts.yml.j2
+++ b/roles/middleware_monitoring_config/templates/target_down_alerts.yml.j2
@@ -12,7 +12,7 @@ spec:
         annotations:
           message: >-
             {{ '{{' }} printf "%.4g" $value {{ '}}' }}% of the {{ '{{' }} $labels.job {{ '}}' }}/{{ '{{' }} $labels.service {{ '}}' }} targets in the {{ '{{' }} $labels.namespace {{ '}}' }} namespace are down.
-        expr: 100 * (count(up{job!="blackbox"} == 0) BY (job, namespace, service) / count(up{job!="blackbox"}) BY (job, namespace, service)) > 10
+        expr: 100 * (count(up{job!="blackbox",  job!="openshift-monitoring-federation"} == 0) BY (job, namespace, service) / count(up{job!="blackbox",  job!="openshift-monitoring-federation"}) BY (job, namespace, service)) > 10
         for: 10m
         labels:
           severity: warning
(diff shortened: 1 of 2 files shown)


obrienrobert:
@Boomatang using the verification steps, the alert looks to be working as expected. I've noticed the alert message might need to be formatted/modified to be easier to understand:

OpenshiftMonitoringTargetDown

Boomatang:
@obrienrobert Thanks for that, I have updated the alert message and if the same information from your image is used it should read. `The endpoint, prometheus-k8s.openshift-monitoring.svc:9091, is down in openshift-monitoring namespace.`

roles/middleware_monitoring_config/templates/target_down_alerts.yml.j2
```diff
@@ -23,4 +23,12 @@ spec:
         expr: (up{job="blackbox"}==0)
         for: 10m
         labels:
-          severity: warning
\ No newline at end of file
+          severity: warning
+      - alert: OpenshiftPrometheusFederationTargetDown
+        annotations:
+            message: >-
+                The endpoint, {{ '{{' }} $labels.instance {{ '}}' }}, is down in {{ '{{' }} $labels.namespace {{ '}}' }} namespace.
```
obrienrobert, line 30: Great, thank for making that change! One small thing. How about changing `is down in` to `is down in the`?

Boomatang: Do you know what I was reading it back after I force push and thought the very same thing. Couldn't decide if I should push again.

obrienrobert:
I've had a quick look at the Prometheus docs around scrape targets:
```
In Prometheus terms, an endpoint you can scrape is called an instance, usually corresponding to a 
single process. A collection of instances with the same purpose, a process replicated for scalability 
or reliability for example, is called a job.
```
It looks like once endpoint is an instance. Maybe it would make sense to reflect this in the message, something like:

instanceDown

WDYT?

Boomatang:
@obrienrobert It does read better as the instance and I have changed it around. I do feel also we maybe getting into the weeds about now and if there is more changes maybe it would be better to have a call about it.

obrienrobert:
Verified this using the steps provided.

/lgtm

obrienrobert (approved):

PR #1436, opened 2020-12-15
INTLY-10338 - delete zync ServiceMonitor

matskiv:
## Additional Information

JIRA: https://issues.redhat.com/browse/INTLY-10338
We decided that it would be best to removing this ServiceMonitor, since it's hasn't been working since 2.x GA and zync metrics are not available in RHMI 2.x and RHOAM.

## Verification Steps
Run upgrade playbook and verify that "3scale-zync-servicemonitor" ServiceMontior was removed from 3scale ns
OR carefully review following log:
https://gist.github.com/matskiv/b41ae842f33023b5e302c3210d6094c6#file-upgrade-log-intly-10338

### Installation Verification
### Upgrade Verification

## Checklist

- [ ] Tested Installation
- [ ] Tested Uninstallation

- [ ] Yes
- [ ] No


grdryn:
/approve

@matskiv I presume there are no alerts or dashboards that rely on metrics that prometheus would have collected as a result of this servicemonitor? Also, do you think it's worth adding a changelog entry for this?

matskiv:
@grdryn I didn't find any usage of the zync metrics, and AFAIK, this ServiceMonitor was broken since LA release, so the metrics were never collected.

jessesarn:
Automation is being stupid.  Just gonna merge it.

PR #1440, opened 2021-04-20
Fix install

CathalOConnorRH:
## Verification Steps

### Installation Verification
### Upgrade Verification

## Checklist

- [ ] Tested Installation
- [ ] Tested Uninstallation

- [ ] Yes
- [ ] No


roles/rhsso/tasks/install_sso.yml
```diff
@@ -26,6 +24,22 @@
   register: rhsso_keycloak
   failed_when: rhsso_keycloak.stderr != '' and 'AlreadyExists' not in rhsso_keycloak.stderr
 
+- name: "remove unused service-ca volume mount"
+  shell: oc set volume dc/sso --remove --name=service-ca -n {{ rhsso_namespace }}
```
jessesarn, line 28:
```suggestion
  shell: oc set volume dc/sso --remove --name=service-ca -n {{ sso_namespace }}
```

jessesarn (changes requested):
Failed to remove the service-ca volume during rhsso_user install because it was using rhsso_namespace instead of sso_namespace var.

jessesarn:
Verified successfully.

Full install log: https://gist.github.com/jessesarn/6cae6744c7fea8e4a7e61a3a5626b806

jessesarn (approved):
Looks good

